PR #4, opened 2019-11-28
Update Readme.md

Nielsbishere:
Updated links. Please move travis to Immediate-Mode-UI/nuklear.

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -1,6 +1,6 @@
 # Nuklear
 
-[![Build Status](https://travis-ci.org/vurtun/nuklear.svg)](https://travis-ci.org/vurtun/nuklear)
+[![Build Status](https://travis-ci.org/Immediate-Mode-UI/nuklear.svg)](https://travis-ci.org/Immediate-Mode-UI/nuklear)
 
 This is a minimal state immediate mode graphical user interface toolkit
 written in ANSI C and licensed under public domain. It was designed as a simple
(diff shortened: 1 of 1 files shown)


dumblob:
Speaking about Travis, I've heard, that the GitHub CI is much faster and adopts packages etc. quicker. We could than use it instead of Travis. What do you think? Didn't try that and have no idea how difficult that would be, but if you looked into that (at best as part of this PR), I'd be really glad.

Nielsbishere:
Adding C/C++ CMake works on my own repo.
I've heard that build times can differ a lot for bigger builds. It can build for both linux and windows.
```yaml
name: C/C++ CI

on: [push, pull_request]

jobs:
  build:
    name: Windows build
    runs-on: windows-latest
    steps:
    - uses: actions/checkout@v1
    - name: cmake build
      run: cmake . -G "Visual Studio 16 2019"
    - name: build
      run: cmake --build . -j 8
```
Though you aren't using cmake, so you'd run python instead.

dumblob:
Thanks for the highlights. Would you want to add a plain `make` (not `cmake`) github CI support to this repo? If rather not, I'll merge this PR as is.

Nielsbishere:
Never used make before. Best case scenario, you'd just have to generate a cpp file with the NK_IMPLEMENTATION define and include and then throw that to gcc or something. However, if you want to run the python script, you could do that too.

dumblob:
Merged as is. Using plain `make` is important for ouf-of-box cross platform support.

PR #9, opened 2019-12-03
Scrollfix1

lundril:
fix nk_do_edit autoscroll.
    
    How to test:
    Create an edit widget.
    Enter a long text into the edit widget.
    Press "home" so that widget autoscrolls to start of text.
    Press "end" so that widget now should jump to end of text.
    Previously you had to press "end" several times to jump to the end of text.
    This patch fixes that (pressing "end" once will jump to the end of text).

### DIFF
--- a/doc/nuklear.html
+++ b/doc/nuklear.html
@@ -702,6 +702,7 @@ as well as some other configuration values have to be configured by filling out 
 `nk_convert_config` struct.
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~c
 nk_flags nk_convert(struct nk_context *ctx, struct nk_buffer *cmds,
+    struct nk_buffer *vertices, struct nk_buffer *elements, const struct nk_convert_config*);
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 Parameter   | Description
 ------------|-----------------------------------------------------------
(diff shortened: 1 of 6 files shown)


dumblob:
Thanks @lundril. May I ask one question - would you recommend some measures to make contributions to Nuklear easier? While reading your PR I thought e.g. about some (more) automation around `doc/nuklear.html` and `package.json`.

Btw. would you be interested in reviewing PRs of others? If so, I'll invite you as an admin member of this org :wink:.

lundril:
Hello dumblob,

> May I ask one question - would you recommend some measures to make contributions to Nuklear easier?

The information is already somewhere there.
So basically it says somewhere to first modify the ".c" files and then generate the ".h" file
and the same is somewhere there for the ".html" I think.
I will look more closely: Maybe it is a good idea to just write it down more prominently somewhere.


> Btw. would you be interested in reviewing PRs of others? If so, I'll invite you as an admin member of this
> org :wink:.

I feel honored!
But right now I am not sure if I am able to fulfill this responsibility.
I think my response time would be too bad.
Let me think about it, before I commit to this.

so long
  lundril

dumblob:
>Maybe it is a good idea to just write it down more prominently somewhere.

Thanks for the suggestion, will try once I'll find some more time.

>But right now I am not sure if I am able to fulfill this responsibility.

Take it easy as it won't feed you :wink:. In open source it's all good to publicly say *"I'm stepping back for an undefined period"* any time.

Issue #11, opened 2019-12-04
Compile issues with strict C++17


Nielsbishere:
Strict C++17 (all warnings as errors), doesn't allow if with a constexpr expression in there. That's what `if constexpr` is for. This is an easy fix; only define the nk_memcpy & nk_memset implementations if they are used. I did it by adding a `NK_DEF_MEMCPY` and `NK_DEF_MEMSET` if the user didn't define theirs. And the check for if(sizeof(nk_draw_index)==2) should only happen ifndef NK_UINT_DRAW_INDEX. 
https://github.com/Nielsbishere/Nuklear/commit/d0c22ebf09e5b72c502b4e88da2d88c474f189c4
The branch above fixes it, but I don't know where those are located in the actual source (only the nuklear.h is modified).

dumblob:
`grep -Er 'define NK_MEMSET' src/` is my friend :wink::
```
src/nuklear_internal.h:#define NK_MEMSET nk_memset
```

```
master 0$ grep -Er 'NK_UINT_DRAW_INDEX' src/
src/nuklear.h:#ifdef NK_UINT_DRAW_INDEX
src/HEADER:/// NK_UINT_DRAW_INDEX              | Defining this will set the size of vertex index elements when using NK_VERTEX_BUFFER_OUTPUT to 32bit instead of the default of 16bit
src/HEADER:///     - NK_UINT_DRAW_INDEX
```

```
master 0$ grep -Er 'nk_memcopy' src/
src/nuklear_internal.h:#define NK_MEMCPY nk_memcopy
src/nuklear_internal.h:NK_LIB void* nk_memcopy(void *dst, const void *src, nk_size n);
src/nuklear_util.c:nk_memcopy(void *dst0, const void *src0, nk_size length)
```

dumblob:
Btw. why not use directly `#ifdef NK_MEMSET` directly and introduce another "chained" define `#ifdef NK_DEF_MEMCPY`?

Nielsbishere:
The point is that there's a few checks like this:
`if(sizeof(nk_draw_index) > 2)` which in c++17 with /Wall /Werr doesn't compile since it should be constexpr. This expression can be done by a simple ifdef, because it is only larger if it's defined as an int. I myself replace memcpy and memset by std::memcpy and std::memset, so it shouldn't even try to compile those IMO.

PR #13, opened 2019-12-04
Strict c++17 now compiles

Nielsbishere:
Fixes issue https://github.com/Immediate-Mode-UI/Nuklear/issues/11

Now detecting if memcpy and memset are actually needed. And same goes for the vertex check (they check correctly now).

Paq.sh now automatically outputs to nuklear.h.


dumblob:
Continuing discussion from #11:

Why is it needed to introduce yet another define? Couldn't we use `#ifdef NK_MEMSET` and `#ifdef NK_MEMCPY` directly as mentioned in https://github.com/Immediate-Mode-UI/Nuklear/issues/11#issuecomment-561688043 ?

Nielsbishere:
Because the moment that ifdef is used, it is immediately defined.

dumblob:
I mean, c89 supports `#ifdef NK_MEMSET == nk_memset`...

Nielsbishere:
Does it? C++17 gave me an error on that.

dumblob:
Right, `nk_memset` would get expanded to `0`, so that won't work. In that case, could you please rename `NK_DEFAULT_MEMCPY` to `NK_MEMCPY_BUILTIN` and `NK_DEFAULT_MEMSET` to `NK_MEMSET_BUILTIN`? It's usually better to use the same prefix if chaining anything.

dumblob:
And because it changes `nuklear.h`, we should also bump version.

Nielsbishere:
Okay, I'll get on it when I got time again

Nielsbishere:
Fixed.

dumblob:
One last little piece - the version info has to be updated also in https://github.com/Immediate-Mode-UI/Nuklear/blob/master/src/CHANGELOG :wink: (and `nuklear.h` regenerated).

Nielsbishere:
Done.

Nielsbishere:
Merged with the latest PR and udpated so it works with strict clang c++17 too.

dumblob:
I'm sorry, didn't notice it before, but could you make `paq.bat` behave in a way, that it outputs no `CRLF`, but just `LF` line endings? `nuklear.h` was always with `LF` and I'd like it to stay so.

Thanks again for your effort!

Nielsbishere:
Fixed @dumblob, python used 'print' which prints newline characters based on the current architecture. I set it to force linux line endings.

Hejsil:
LGTM. Maybe just squash the commits into one, to keep the git history tidy :)

PR #14, opened 2019-12-04
Fix click

Nielsbishere:
https://github.com/vurtun/nuklear/pull/803 
keharriso's Prevent clicking more than one (overlapping) button per frame.

 Problem with existing PR was as follows:
- Sets click to 0 every start of frame
- Input was still down for that frame, so it triggered the next event

Now it waits for your button to go back up before resetting the mouse's clicked variable, so as long as your mouse is down, only one item is clicked.


src/nuklear_button.c
```diff
@@ -78,6 +78,8 @@ nk_button_behavior(nk_flags *state, struct nk_rect r,
 #else
                 nk_input_is_mouse_pressed(i, NK_BUTTON_LEFT);
 #endif
+			if (ret)
```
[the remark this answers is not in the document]
Nielsbishere: What about it?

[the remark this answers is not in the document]
dumblob: There are 3 tabs on that line, but `nuklear.h` uses spaces for indentation. At best, we should stick with that instead of introducing inconsistencies.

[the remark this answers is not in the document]
Nielsbishere: Fixed, I think.

src/nuklear_button.c
```diff
@@ -108,7 +110,7 @@ nk_draw_button(struct nk_command_buffer *out,
 }
 NK_LIB int
 nk_do_button(nk_flags *state, struct nk_command_buffer *out, struct nk_rect r,
-    const struct nk_style_button *style, const struct nk_input *in,
+    const struct nk_style_button *style, struct nk_input *in,
```
Hejsil: Removing `const` here (an other places) is a breaking change in cases where people use `-Werror`, because users might have written code simular to this:
```c
const struct nk_input *in = ...
...
nk_do_button(state, out, r, style, in, behavior, content)
```

Which would now throw this error:
```
error: passing 'const struct nk_input *' to parameter of type 'struct nk_input *' discards qualifiers [-Werror,-Wincompatible-pointer-types-discards-qualifiers]
```

Should the major version be incremented because of this?

Nielsbishere: I agree but it's the only straightforward way to fix it.

dumblob: Yeah, the missing `const`s are a pity, but I also don't see any more suitable way to solve it. Maybe we should increase the *minor* part of the version instead of the *bugfix* part to emphasize this breaking change (but in the context of the usage of Nuklear, I think this does not deserve a *major* version bump).

Hejsil: Hmm. Yea bumping major really is overkill, even though it is consistent with the documentation for the changelog. I think minor bump + emphasis is good enough.

PR #15, opened 2019-12-04
Fix radio buttons

Nielsbishere:
This fix changes the looks of the radio buttons (like https://github.com/vurtun/nuklear/pull/653 zhouxs1023's Modify the appearance for option and checkbox!).

With a few minor changes to make it clearer. And README has been changed accordingly and a more thorough example was added.


Readme.md
```diff
@@ -56,40 +56,58 @@ This is very important not doing it either leads to compiler errors or even wors
 
 ```c
 /* init gui state */
-struct nk_context ctx;
-nk_init_fixed(&ctx, calloc(1, MAX_MEMORY), MAX_MEMORY, &font);
+struct nk_context *ctx = (nk_context*) malloc(sizeof(nk_context));
```
dumblob: Nitpick: why on heap? It's a small struct and the less allocated memory you have to manage, the better (and more readable).

Nielsbishere: True, I'll fix this, it's because on my own project there's too much on the stack already.

Nielsbishere:
Updated.

dumblob:
Please bump version also in `src/CHANGELOG` (and regenerate `nuklear.h`).

Btw. why did you remove all the `const`s? They're quite useful IMHO. Or is it some residual after debugging/testing the lib?

Nielsbishere:
Will do, the consts are because of the merge with fix_click. The input struct can't be const, since clicking in the mouse rect will modify it (so you can't click through two objects and active both of them).

Nielsbishere:
Updated.

Issue #18, opened 2019-12-07
Invalid hover detection when windows are touching


luknw:
#### What?
When windows are pixel-perfect aligned one next to the other, the hover status isn't updated properly.

#### Minimum example
Replace the window in the default demo with the following:
```C
if (nk_begin(ctx, "Demo1", nk_rect(100, 100, 200, 200),
    NK_WINDOW_MOVABLE|NK_WINDOW_CLOSABLE|NK_WINDOW_MINIMIZABLE|NK_WINDOW_TITLE))
{
    nk_layout_row_dynamic(ctx, 80, 1);
    nk_button_label(ctx, "button");
}
nk_end(ctx);
if (nk_begin(ctx, "Demo2", nk_rect(300, 100, 200, 200),
    NK_WINDOW_MOVABLE|NK_WINDOW_CLOSABLE|NK_WINDOW_MINIMIZABLE|NK_WINDOW_TITLE))
{
    nk_layout_row_dynamic(ctx, 80, 1);
    nk_button_label(ctx, "button");
}
nk_end(ctx);
```
![nuklear-bug-report-bad](https://user-images.githubusercontent.com/13088361/70362380-15949f00-1885-11ea-9bc6-c456b0fb3351.gif)
The left button doesn't react on hover.

When windows are spread apart by 1 or more pixels, hover works fine:
```C
if (nk_begin(ctx, "Demo1", nk_rect(100, 100, 200, 200),
    NK_WINDOW_MOVABLE|NK_WINDOW_TITLE))
{
    nk_layout_row_dynamic(ctx, 80, 1);
    nk_button_label(ctx, "button");
}
nk_end(ctx);
if (nk_begin(ctx, "Demo2", nk_rect(300 + 1, 100, 200, 200),
    NK_WINDOW_MOVABLE|NK_WINDOW_TITLE))
{
    nk_layout_row_dynamic(ctx, 80, 1);
    nk_button_label(ctx, "button");
}
nk_end(ctx);
```
![nuklear-bug-report-good](https://user-images.githubusercontent.com/13088361/70362603-dc106380-1885-11ea-8516-b8beff502c9e.gif)

#### Suggested fix
The hover isn't updated, because windows are detected as intersecting each other:
https://github.com/Immediate-Mode-UI/Nuklear/blob/adc52d710fe3c87194b99f540c53e82eb75c2521/src/nuklear_window.c#L221-L242
The `NK_INTERSECT` is defined as:
[...]

PR #19, opened 2019-12-07
Fix off-by-one error in NK_INTERSECT

luknw:
Fixes intersect detection for the upper bound, see #18 for rationale.

### DIFF
--- a/nuklear.h
+++ b/nuklear.h
@@ -5622,7 +5622,8 @@ struct nk_context {
 #define NK_INBOX(px, py, x, y, w, h)\
     (NK_BETWEEN(px,x,x+w) && NK_BETWEEN(py,y,y+h))
 #define NK_INTERSECT(x0, y0, w0, h0, x1, y1, w1, h1) \
-    (!(((x1 > (x0 + w0)) || ((x1 + w1) < x0) || (y1 > (y0 + h0)) || (y1 + h1) < y0)))
+    ((x1 < (x0 + w0)) && (x0 < (x1 + w1)) && \
+    (y1 < (y0 + h0)) && (y0 < (y1 + h1)))
 #define NK_CONTAINS(x, y, w, h, bx, by, bw, bh)\
     (NK_INBOX(x,y, bx, by, bw, bh) && NK_INBOX(x+w,y+h, bx, by, bw, bh))
 
@@ -25475,6 +25476,7 @@ nk_tooltipfv(struct nk_context *ctx, const char *fmt, va_list args)
 ///    - [yy]: Minor version with non-breaking API and library changes
 ///    - [zz]: Bug fix version with no direct changes to API
 ///
+/// - 2019/12/10 (4.01.5) - Fix off-by-one error in NK_INTERSECT
 /// - 2019/10/09 (4.01.4) - Fix bug for autoscrolling in nk_do_edit
 /// - 2019/09/20 (4.01.3) - Fixed a bug wherein combobox cannot be closed by clicking the header
 ///                        when NK_BUTTON_TRIGGER_ON_RELEASE is defined.
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "nuklear",
-  "version": "4.01.4",
+  "version": "4.01.5",
   "repo": "vurtun/nuklear",
   "description": "A small ANSI C gui toolkit",
   "keywords": ["gl", "ui", "toolkit"],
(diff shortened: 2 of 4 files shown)


Nielsbishere:
Almost looks correct. But it should only apply for the lower x & y bounds, the upper should be excluded. For example if you'd have two cells next to each other (tightly packed), a click in the begin of the second cell shouldn't be included in the first cell.

luknw:
@Nielsbishere I'm not entirely sure if I understand you correctly, please tell me if I'm wrong.
Do you mean that the two inequalities in `NK_INTERSECT` check the lower bound and the other two the upper bound? So, following this logic, `NK_INTERSECT` should look like this?
```C
#define NK_INTERSECT(x0, y0, w0, h0, x1, y1, w1, h1) \
    (!(((x1 > (x0 + w0)) || ((x1 + w1) <= x0) || (y1 > (y0 + h0)) || (y1 + h1) <= y0)))
```

Nielsbishere:
You should probably test the scenario I explained; like have two cells next to each other without padding or anything in between. Clicking the exact first pixel of the second cell and the pixel before it. Two of those checks shouldn't use equal and two should, but I can't tell you which ones since I havent touched that code yet.

luknw:
Ok, the logic in `NK_INTERSECT` is a bit convoluted, but I think your reasoning is wrong here and all the checks should use equal.

I tested this to be sure and when only two checks use equals, the intersect is still detected wrong. The reason why is because all the checks involve comparison with either first cell or second cell upper bound.

Let's say we swap the order in which we provide first cell and second cell to `NK_INTERSECT`. The result should stay the same, because we test the same two cells. If two inequalities use equals and the other two don't, the checks would produce different results depending on the order of the cells.

Nielsbishere:
I mean the following:
cell 0 at 0,0->16,16
cell 1 at 16,0->32,16
If you click 16,0 it should only intersect 1.
If you click at 32,0 it should intersect none.

luknw:
@Nielsbishere I've tested the layout you came up with and there's no difference regarding the inequalities. I've peeked in the code and mouse clicks aren't detected using `NK_INTERCEPT`, but `NK_INBOX` instead.

dumblob:
I'd like to merge this, but I'll wait for the consensus as I want to avoid misunderstandings.

Nielsbishere:
What is using NK_INTERSECT then? What do you need to change this for? Does NK_INBOX use the correct equation?

luknw:
`NK_INTERSECT` detects if any two rectangles intersect. There are two types of usages in the codebase:
1. detect if drawn shape bounds intersect clip rectangle
1. detect if two windows intersect each other

I encountered the issue with the second type when I wanted to have two windows aligned side by side. When hovering a mouse over a window, the window should become active, but only when there isn't any other window intersecting it. When window is active, widgets inside that window should change color on mouse hover. Due to the bug, the tightly arranged windows don't become active on hover and the widgets don't change color.

You can see how this behavior looks like in the issue I've filed: #18 

When it comes to `NK_INBOX`, it uses the correct equation. It's defined as:
https://github.com/Immediate-Mode-UI/Nuklear/blob/a3f40348d572fad24af1e2d6c6b755171af481db/src/nuklear.h#L5402-L5404

You can see, that `NK_BETWEEN` has a non-strict inequality in the lower bound and a strict one in the upper bound.

Nielsbishere:
Seems good then, pretty good fix

luknw:
Ok, now when we have consensus, I'd like to propose one last tiny change. Let's simplify boolean expression inside `NK_INTERSECT` to get rid of `not` at the beginning.

dumblob:
Let's do this. One last thing is to bump version (in `src/CHANGELOG` and in `package.json` and then regenerate `nuklear.h` as described in `src/Readme.md`) :wink:.

luknw:
All right, I'll squash the changes into one commit.

luknw:
Ok, should be ready to merge.

Btw. `package.json` has a repo set to `vurtun/nuklear`, shouldn't it be updated to `Immediate-Mode-UI/Nuklear`?

dumblob:
>Btw. package.json has a repo set to vurtun/nuklear, shouldn't it be updated to Immediate-Mode-UI/Nuklear?

Good catch! Definitely needs an update. Feel free to do that in your PR. I'll merge this PR tomorrow at the latest.

dumblob:
Hope this didn't get too tedious :wink:. Thanks!

PR #24, opened 2019-12-12
Add pure C wayland demo

skoperst:
This implementation uses software rendering in native wayland client.

### DIFF
--- a/demo/wayland_rawfb/Makefile
+++ b/demo/wayland_rawfb/Makefile
@@ -1,0 +1,9 @@
+WAYLAND=`pkg-config wayland-client --cflags --libs`
+CFLAGS?=-std=c11 -Wall -Werror -O3 -fvisibility=hidden
+
+hello_wayland: main.c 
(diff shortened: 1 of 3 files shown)


dumblob:
Didn't test it, but overall the PR looks too good to not merge it :wink:. Thanks!

I can see there the following TODO items for future contributors:
1. change bubble sort at least to quick sort or better to radix sort (which is the fastest to me known method for sorting integers)
2. in `nk_wayland_line_horizontal()` there is plenty room for optimization - e.g.:
    1. avoid most of the pointers (especially `unsigned int *ptr` working with the same memory as `unsigned char *pixels`)
    2. pass `win->data` and `win->width` directly instead of just `win`
    3. use `uint8_t` everywhere instead of `unsigned int`
    4. use `restrict` for all pointers
    5. manually unroll first two loops (C macros could ease readability)
    6. make it inline
3. `nk_wayland_stroke_line()` could also get some optimization love similar to (2) :wink:
4. there are some unimplemented `NK_COMMAND_`...

PR #35, opened 2019-12-23
Window improvements: callbacks, ids

Nielsbishere:
nk_add_window has been added to support adding a window that isn't necessarily drawn and returning it. You then have to check via `nk_window_has_contents` to see if you can actually fill the window. This is very useful as this provides you with a callback whenever your window is closed or minimized.
It also takes in an id and a title instead of a name of the window. Meaning you don't need a name and title for a window anymore to keep them apart (so no nk_begin and nk_begin_titled equiv).

nk_begin now returns an nk_window, so you can get the location and dimension of the window (handy if it is resizable). All old functionality will remain the same (it used to return 1 instead of a pointer, but if(nk_begin) will remain the same, since it checks for 0).

An example for when this is used in pratice; abstraction for UI where you'd like to know physical location of the window and keep that up-to-date.

This branch depends on fix_radio_buttons.


Nielsbishere:
This is also useful if you want to have a max or minimum size for the window:
```cpp
if(nk_window *wnd = nk_begin(...)){

 if(wnd->bounds.w < minW)
   wnd->bounds.w = minW;
 if(wnd->bounds.h < minH)
  wnd->bounds.h = minH;

 //...
 nk_end();
}
```

src/nuklear.h
```diff
@@ -1256,10 +1256,10 @@ enum nk_panel_flags {
 /// __bounds__  | Initial position and window size. However if you do not define `NK_WINDOW_SCALABLE` or `NK_WINDOW_MOVABLE` you can set window position and size every frame
 /// __flags__   | Window flags defined in the nk_panel_flags section with a number of different window behaviors
 ///
-/// Returns `true(1)` if the window can be filled up with widgets from this point
-/// until `nk_end` or `false(0)` otherwise for example if minimized
+/// Returns a non-zero window pointer if the window can be filled up with widgets from this point
+/// until `nk_end` or `nullptr(0)` otherwise for example if minimized
 */
-NK_API int nk_begin(struct nk_context *ctx, const char *title, struct nk_rect bounds, nk_flags flags);
+NK_API struct nk_window* nk_begin(struct nk_context *ctx, const char *title, struct nk_rect bounds, nk_flags flags);
```
Hejsil: Heh, this return type change is also a breaking change :)

src/CHANGELOG
```diff
@@ -8,6 +8,12 @@
 ///    - [yy]: Minor version with non-breaking API and library changes
 ///    - [zz]: Bug fix version with no direct changes to API
 ///
+/// - 2019/12/23 (4.01.9) - Changed nk_begin to return a nk_window so you can get the position and other useful info.
```
Hejsil: This is a change that needs a minor bump :)

src/nuklear.h
Nielsbishere: if(nk_begin(...)) is generally used and that will still behave the exact same way. Only if the user stores int it will break. Otherwise it will now provide functionality to immediately get the window you opened

Hejsil: Agreed, this is not a problem for most users, but it is a breaking change for some. I'm only really pointing this out because the patch version was bumped, and not at least the minor version (maybe I could have been a little clearer, mb).

Nielsbishere: Okay, will get on bumping the minor versions and merging asap

Issue #51, opened 2020-01-19
Please update the version of stb_truetype.h and stb_rect_pack.h in nuklear.h


ronaaron:
The versions are very old, and in particular a number of bugs in TTF handling have been fixed in more recent versions.

dumblob:
Thanks for letting us know. There was a similar request and I think the goal was to automatically update them in each run of `paq.sh`/`paq.bat` (i.e. during the "amalgamation" process). PRs are welcome and because it sounds like an easy to review change, they'll be merged very quickly (couple of days probably).

ronaaron:
I was going to push to a new branch, but I don't seem to be able to do  that.  So I'm attaching the modified 'src/nuklear_font.h'.

**NOTE**: There are two functions which in the current version take an nk_allocator, but which have no such thing in the code I merge; I do not understand all the ramifications of *not* having the allocator, it seems to work as-is on my Linux box.

The functions in question are part of the baker:

- nk_tt_PackFontRangesRenderIntoRects
- nk_tt_PackEnd


[nuklear_font.zip](https://github.com/Immediate-Mode-UI/Nuklear/files/4086136/nuklear_font.zip)

ronaaron:
The code does crash on macOS, so apparently the allocator is needed. I was hoping it might be possible to allocate whatever's needed beforehand rather than piecemeal.  But I don't know the code, and am relying on you guys.

ronaaron:
Still relying on you guys... any feedback on the need for the allocator?

dumblob:
I'm still out of time to do reviews. But what caught my eye is that you can't submit a PR - that's weird. Could you please try again (step by step as [GitHub doc](https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/creating-a-pull-request ) says)? It would encourage others to look over your changes :wink:.

ronaaron:
Pinging again, since I put a PR in just after your last comment.  Looking for assistance on that issue, since the ancient TTF support is hindering me.

dumblob:
@Nielsbishere feel free to review the PR - I'd like to give you commit permission afterwards :wink:

SairesArt:
@ronaaron Someone shares my pain <3
I have been bothered by the old stb_library versions myself. Many thanks for for the PR and the contribution!

ronaaron:
Yeah; I hope it gets integrated (with appropriate changes) soon.

dumblob:
[...]

PR #53, opened 2020-01-22
first attempt at updating stb_truetype and stb_rect_pack

ronaaron:
In response to the [issue I reported](https://github.com/Immediate-Mode-UI/Nuklear/issues/51), I forked the repo and pushed my changes to a branch.

The code does compile and it seems to run 'ok' on Linux, but it crashes on macOS because apparently the allocator stuff in the baker percolates around. 

I don't understand all the ramifications so I'm asking for someone who knows the code to please assist.


ronaaron:
Updated the ttf branch so the stb files are separate.

src/paq.sh
```diff
@@ -1,3 +1,3 @@
...
+python build.py --macro NK --intro HEADER --pub nuklear.h --priv nuklear_internal.h,nuklear_math.c,nuklear_util.c,nuklear_color.c,nuklear_utf8.c,nuklear_buffer.c,nuklear_string.c,nuklear_draw.c,nuklear_vertex.c,stb_rect_pack.h,stb_truetype.h,nuklear_font.c,nuklear_input.c,nuklear_style.c,nuklear_context.c,nuklear_pool.c,nuklear_page_element.c,nuklear_table.c,nuklear_panel.c,nuklear_window.c,nuklear_popup.c,nuklear_contextual.c,nuklear_menu.c,nuklear_layout.c,nuklear_tree.c,nuklear_group.c,nuklear_list_view.c,nuklear_widget.c,nuklear_text.c,nuklear_image.c,nuklear_button.c,nuklear_toggle.c,nuklear_selectable.c,nuklear_slider.c,nuklear_progress.c,nuklear_scrollbar.c,nuklear_text_editor.c,nuklear_edit.c,nuklear_property.c,nuklear_chart.c,nuklear_color_picker.c,nuklear_combo.c,nuklear_tooltip.c --outro LICENSE,CHANGELOG,CREDITS
```
dumblob: Could you also update `paq.bat` (careful with `CRLF`)?

src/nuklear_font.c
```diff
@@ -376,1689 +17,33 @@ nk_rp_pack_rects(struct nk_rp_context *context, struct nk_rp_rect *rects, int nu
...
-        q->y0 = y;
-        q->x1 = x + b->xoff2 - b->xoff;
-        q->y1 = y + b->yoff2 - b->yoff;
-    } else {
-        q->x0 = *xpos + b->xoff;
-        q->y0 = *ypos + b->yoff;
-        q->x1 = *xpos + b->xoff2;
-        q->y1 = *ypos + b->yoff2;
-    }
-    q->s0 = b->x0 * ipw;
```
dumblob: With which fonts did you test this?

src/nuklear_font.c
```diff
@@ -2318,11 +303,13 @@ nk_font_bake(struct nk_font_baker *baker, void *image_memory, int width, int hei
         it = config_iter;
         do {const struct nk_font_config *cfg = it;
             struct nk_font_bake_data *tmp = &baker->build[input_i++];
-            nk_tt_PackSetOversampling(&baker->spc, cfg->oversample_h, cfg->oversample_v);
-            nk_tt_PackFontRangesRenderIntoRects(&baker->spc, &tmp->info, tmp->ranges,
-                (int)tmp->range_count, tmp->rects, &baker->alloc);
+            stbtt_PackSetOversampling(&baker->spc, cfg->oversample_h, cfg->oversample_v);
+			// TODO:
+            // was: stbtt_PackFontRangesRenderIntoRects(&baker->spc, &tmp->info, tmp->ranges, (int)tmp->range_count, tmp->rects, &baker->alloc);
+            stbtt_PackFontRangesRenderIntoRects(&baker->spc, &tmp->info, tmp->ranges, (int)tmp->range_count, tmp->rects);
```
dumblob: The stb allocator should use the Nuklear allocator. The following is from `stb_truetype.h`:

```
   // #define your own functions "STBTT_malloc" / "STBTT_free" to avoid malloc.h
￼   #ifndef STBTT_malloc
￼   #include <stdlib.h>
￼   #define STBTT_malloc(x,u)  ((void)(u),malloc(x))
￼   #define STBTT_free(x,u)    ((void)(u),free(x))
￼   #endif
```

src/paq.sh
ronaaron: Done

src/nuklear_font.c
ronaaron: Roboto-Regular, ProggyClean, and some random font I had.

src/nuklear_font.c
ronaaron: How do I bump the Nuklear version?

dumblob: See https://github.com/Immediate-Mode-UI/Nuklear/blob/master/src/CHANGELOG

ronaaron:
I've compiled on both macOS and Linux, I have no idea why this 'build failed'.

ronaaron:
Dont know why the build failed for github, it works fine for me on macOS and Linux

dumblob:
That's weird it compiled for you as the syntax `{0}` is not correct C99 (correctly it should be `(struct ...){0}` - i.e. with casting :wink:). Try to use the same arguments as CI uses, i.e. `-std=c99 -pedantic -O2`.

dumblob:
On the other hand, I'd still say we should stick with C89 for such core functionality (only demos and backends are allowed to use C99 or newer, but the core `nuklear.h` should stay for some years at C89 for compatibility reasons).

src/nuklear.h
```diff
@@ -5461,4 +5461,12 @@ template<typename T> struct nk_alignof{struct Big {T x; char c;}; enum {
 #define NK_ALIGNOF(t) ((char*)(&((struct {char c; t _h;}*)0)->_h) - (char*)0)
 #endif
 
+#ifdef NK_IMPLEMENTATION
+#define STB_RECT_PACK_IMPLEMENTATION
+#define STB_TRUETYPE_IMPLEMENTATION
+#endif
+
+#define STBTT_malloc(x,u) nk_malloc((nk_handle){0},0,x)
```
dumblob: Please change it to C89 (something like `#define STBTT_malloc(x,u) do{ nk_handle _ = {0}; nk_malloc( _, 0, x ) }while(0)`).

src/paq.bat
```diff
@@ -1 +1 @@
...
\ No newline at end of file
+build.py --macro NK --intro HEADER --pub nuklear.h --priv nuklear_internal.h,nuklear_math.c,nuklear_util.c,nuklear_color.c,nuklear_utf8.c,nuklear_buffer.c,nuklear_string.c,nuklear_draw.c,nuklear_vertex.c,stb_rect_pack.h,stb_truetype.h,nuklear_font.c,nuklear_input.c,nuklear_style.c,nuklear_context.c,nuklear_pool.c,nuklear_page_element.c,nuklear_table.c,nuklear_panel.c,nuklear_window.c,nuklear_popup.c,nuklear_contextual.c,nuklear_menu.c,nuklear_layout.c,nuklear_tree.c,nuklear_group.c,nuklear_list_view.c,nuklear_widget.c,nuklear_text.c,nuklear_image.c,nuklear_button.c,nuklear_toggle.c,nuklear_selectable.c,nuklear_slider.c,nuklear_progress.c,nuklear_scrollbar.c,nuklear_text_editor.c,nuklear_edit.c,nuklear_property.c,nuklear_chart.c,nuklear_color_picker.c,nuklear_combo.c,nuklear_tooltip.c --outro LICENSE,CHANGELOG,CREDITS > ..\nuklear.h
```
dumblob: I'm pretty sure this should be `CRLF` and not `LF`.

src/nuklear.h
ronaaron: Problem: the stb_* files have extensive "//" comments, which are disallowed by C89.

What's the rationale for not supporting C99? It's been 2 decades...

src/paq.bat
ronaaron: The file came from git as Unix file format, I didn't change that.  Also: even though I don't use Windows, I'm pretty sure that the batch processor doesn't get confused.

src/nuklear.h
dumblob: >Problem: the stb_* files have extensive "//" comments, which are disallowed by C89.

Let's try adding a filter changing `//` to `/* ... */` (linewise is enough even if it'll look ugly) to the python script.

>What's the rationale for not supporting C99? It's been 2 decades...

Yep, it's long time ago, but the embedded world, which is one of the targets of this lib is even slower :cry:. On the other hand, we could make a strict schedule when we'll switch over to C99. What about a poll running for 4 months in this repo issue tracker (as pinned issue)?

src/paq.bat
dumblob: Well, the file didn't have any line ending and you added one :wink:. Whether cmd.exe would or would not get confused depends also on other things, but CRLF is the only option known to me which is guaranteed to be safe.

src/nuklear.h
SairesArt: > What about a poll running for 4 months in this repo issue tracker (as pinned issue)?

Ahhhh yeah, now I remember the pain myself^^
Keeping the ANSI C description is definitely a fun argument to take a look at nuklear and is the reason I initially found nuklear.
Writing a big library such as nuklear to that standard is not trivial. That's quite the achievement in it's self.
However, leaving the C89 in favor of C99, just because **another** library needs comments replaced is not a good argument, especially in the light of @vurtun 's amazing work.
I too have come across this exact moment and have always transformed stb's libraries to C89 in the past, though I'm not sure I still have the regex script for that...
Lastly, sean berret has often stated, that he writes in C89 **except** for the comments, as far as I'm aware and from my experience with stb_image, stb_image_write and stb_dxt, all his libraries are fully C89 compliant, once the comments are replaced. (Since MSVC, which stb compiles with, still technically does not support anything beyond C89)

tl;dr:
-I'm ok with a poll in general, but considering changing to C99 because of the comments is horrible idea and quite the metaphorical fuck you to @vurtun 
-Changing comments is trivial and other graphics&openGL projects maintain the ANSI C compatibility, like https://github.com/recp/cglm

ronaaron:
OK, changed the comments to c89...

src/nuklear.h
dumblob: @SairesArt yeah, I fully agree with everything you wrote - I meant to propose 2025 or later in the poll as the last C89 year for Nuklear (I suppose libs like GL will also be "modern enough" at that point to make the switch feel natural and not become an outsider :wink:). But maybe the poll is not necessary.

Btw. without @vurtun's decision for ANSI C, Nuklear would never become such a good library. Respect!

dumblob:
@ronaaron I'm a bit reluctant to change the stb libs - I'd really like to have them 1:1. Would you rather change the python script?

ronaaron:
Hey, I need this to work for my own project...

ronaaron:
Different behavior from the previous version of nuklear.h:

I get SIGSEGV from stbtt_PackFontRangesGatherRects called by nk_font_atlas_bake from nk_glfw3_font_stash_end.

This happens if I have more than 3 or 4 fonts in the atlas. I assume it's due to the allocator, but ...

dumblob:
@ronaaron what does Valgrind verbose output say to the SIGSEGV?

ronaaron:
I'm running on macOS, no valgrind available.  I tried using ASAN, but it doesn't trip.

dumblob:
Well, yeah - the Nuklear comment in that function says it all:
```c
NK_INTERN int
nk_tt_PackFontRangesGatherRects(struct nk_tt_pack_context *spc,
    struct nk_tt_fontinfo *info, struct nk_tt_pack_range *ranges,
    int num_ranges, struct nk_rp_rect *rects)
{
    /* rects array must be big enough to accommodate all characters in the given ranges */
...
```

Is your app open source? If so, could we get it somewhere and run it with Valgrind? If not, you still can quickly run a VM with any mainstream Linux distro :wink:.

ronaaron:
It's not open source, sorry.  But here's relevant valgrind info:

==19314== Conditional jump or move depends on uninitialised value(s)
==19314==    at 0x602742: stbtt_PackFontRangesGatherRects (nuklear.h:15296)
==19314==    by 0x60BC4F: nk_font_bake_pack(nk_font_baker*, unsigned long*, int*, int*, nk_recti*, nk_font_config const*, int, nk_allocator*) (nuklear.h:16460)
==19314==    by 0x60AA77: nk_font_atlas_bake (nuklear.h:17397)
==19314==    by 0x63F757: nk_x11_font_stash_end() (nk_xlib_gl3.h:697)


==19314==  Uninitialised value was created by a heap allocation
==19314==    at 0x4C2FB0F: malloc (in /usr/lib/valgrind/vgpreload_memcheck-amd64-linux.so)
==19314==    by 0x5EE7FA: nk_malloc(nk_handle, void*, unsigned long) (nuklear.h:8026)
==19314==    by 0x60A8A4: nk_font_atlas_bake (nuklear.h:17382)
==19314==    by 0x63F757: nk_x11_font_stash_end() (nk_xlib_gl3.h:697)

==19314== Use of uninitialised value of size 8
==19314==    at 0x602783: stbtt_PackFontRangesGatherRects (nuklear.h:15296)
...

==19314== Invalid read of size 4
==19314==    at 0x602783: stbtt_PackFontRangesGatherRects (nuklear.h:15296)
...

ronaaron:
The problem goes away by increasing the amount of memory allocated in nk_font_baker_memory().
Specifically, I did this:
    *temp += (nk_size)*glyph_count * 2 * sizeof(stbtt_packedchar);
instead of 
    *temp += (nk_size)*glyph_count * sizeof(stbtt_packedchar);

Of course that's not likely the correct solution, but it points in a direction.

ronaaron:
It still works with that fix when I double or triple the number of fonts, so something along those lines is a good fix.

src/nuklear_font.c
```diff
@@ -2152,9 +137,9 @@ nk_font_baker_memory(nk_size *temp, int *glyph_count,
             *glyph_count += nk_range_glyph_count(i->range, range_count);
         } while ((i = i->n) != iter);
     }
-    *temp = (nk_size)*glyph_count * sizeof(struct nk_rp_rect);
-    *temp += (nk_size)total_range_count * sizeof(struct nk_tt_pack_range);
-    *temp += (nk_size)*glyph_count * sizeof(struct nk_tt_packedchar);
+    *temp = (nk_size)*glyph_count * sizeof(struct stbrp_rect);
+    *temp += (nk_size)*glyph_count * sizeof(stbtt_pack_range);
```
dumblob: Overall I think we're getting close to merging. But this change from `` to `` is overly suspicious to merge it :cry:.

I think stb_truetype started using some structs a bit differently in the newer versions. Maybe it's some off-by-one oversight or struct padding or whatever. I'd look at how the structs `nk_rp_rect` `nk_tt_pack_range` `nk_tt_packedchar` `stbrp_rect` `stbtt_pack_range` `stbtt_packedchar` `nk_font_bake_data` are being used everywhere and whether they're used everywhere with corresponding counts as I don't think `stbtt_pack_range` depends on glyph count, but rather truly on the calculated `total_range_count`.

Have you tried just adding `time += 1` or `time += 8` or similar (i.e. a very small fixed number)?

ronaaron:
body p { margin-bottom: 0cm; margin-top: 0pt; } 
  
  
    
    On 10/02/2020 17:41, dumblob wrote:
    
    
      
      @dumblob requested changes on this pull request.
      We're almost there 😉.
      
      In src/nuklear_font.c:
      > @@ -2152,9 +137,9 @@ nk_font_baker_memory(nk_size *temp, int *glyph_count,
             *glyph_count += nk_range_glyph_count(i->range, range_count);
         } while ((i = i->n) != iter);
     }
-    *temp = (nk_size)*glyph_count * sizeof(struct nk_rp_rect);
-    *temp += (nk_size)total_range_count * sizeof(struct nk_tt_pack_range);
-    *temp += (nk_size)*glyph_count * sizeof(struct nk_tt_packedchar);
+    *temp = (nk_size)*glyph_count * sizeof(struct stbrp_rect);
+    *temp += (nk_size)*glyph_count * sizeof(stbtt_pack_range);

      Overall I think we're getting close to merging. But this change
        from to is overly suspicious to merge it 😢.
      I think stb_truetype started using some structs a bit
        differently in the newer versions. Maybe it's some off-by-one
        oversight or struct padding or whatever. I'd look at how the
        structs nk_rp_rect nk_tt_pack_range
        nk_tt_packedchar stbrp_rect stbtt_pack_range
        stbtt_packedchar nk_font_bake_data
        are being used everywhere and whether they're used everywhere
        with corresponding counts as I don't think stbtt_pack_range
        depends on glyph count, but rather truly on the calculated total_range_count.
      Have you tried just adding time += 1 or time
          += 8 or similar (i.e. a very small fixed number)?
      
      
    
    Yes, and it crashed.
    
      [
{
"@context": "http://schema.org",
"@type": "EmailMessage",
"potentialAction": {
"@type": "ViewAction",
"target": "https://github.com/Immediate-Mode-UI/Nuklear/pull/53?email_source=notifications\u0026email_token=AECUVBSRIDLIS652OCP3U43RCFYSPA5CNFSM4KKHB4HKYY3PNVWWK3TUL52HS4DFWFIHK3DMKJSXC5LFON2FEZLWNFSXPKTDN5WW2ZLOORPWSZGOCU4LVCI#pullrequestreview-356039305",
"url": "https://github.com/Immediate-Mode-UI/Nuklear/pull/53?email_source=notifications\u0026email_token=AECUVBSRIDLIS652OCP3U43RCFYSPA5CNFSM4KKHB4HKYY3PNVWWK3TUL52HS4DFWFIHK3DMKJSXC5LFON2FEZLWNFSXPKTDN5WW2ZLOORPWSZGOCU4LVCI#pullrequestreview-356039305",
"name": "View Pull Request"
},
"description": "View this Pull Request on GitHub",
"publisher": {
"@type": "Organization",
"name": "GitHub",
"url": "https://github.com"
}
}
]
    
    
      
        
          Ron AARON | 
           CTO  Aaron High-Tech, Ltd |
          +1 516.373.0794 / +972 52.652.5543 |
          
            GnuPG Key: 91F92EB8

ronaaron:
I fixed the actual problem with the memory. It needs to zero the allocated memory.

src/nuklear_font.c
```diff
@@ -1179,6 +1179,7 @@ nk_font_atlas_bake(struct nk_font_atlas *atlas, int *width, int *height,
     tmp = atlas->temporary.alloc(atlas->temporary.userdata,0, tmp_size);
     NK_ASSERT(tmp);
     if (!tmp) goto failed;
+	memset(tmp,0,tmp_size);
```
dumblob: Please tabs to spaces and then I'm merging. This is a leap forward for Nuklear. Thanks!

dumblob:
Seems you're online - because of the importance of this change, feel free to also bump version in `package.json`. Tomorrow at the latest I'm merging this PR :wink:. Thanks again for your persistence!

dumblob:
Thanks @ronaaron for the tiresome work! It was totally worth it. Hope you'll become project member at some point - feel free to express your will any time :wink:.

src/nuklear_font.c
```diff
@@ -376,1689 +17,33 @@ nk_rp_pack_rects(struct nk_rp_context *context, struct nk_rp_rect *rects, int nu
...
-        nk_int ins, i,j=0,m,n, next_move, was_off=0, off, start_off=0;
-        nk_int x,y,cx,cy,sx,sy, scx,scy;
-        const nk_byte *points;
-        endPtsOfContours = (data + g + 10);
-        ins = nk_ttUSHORT(data + g + 10 + numberOfContours * 2);
-        points = data + g + 10 + numberOfContours * 2 + 2 + ins;
-
-        n = 1+nk_ttUSHORT(endPtsOfContours + numberOfContours*2-2);
-        m = n + 2*numberOfContours;  /* a loose bound on how many vertices we might need */
-        vertices = (struct nk_tt_vertex *)alloc->alloc(alloc->userdata, 0, (nk_size)m * sizeof(vertices[0]));
```
2asoft, line 792: The adapted `stb_truetype` library in Nuklear used to use the `nk_allocator` that `nk_font_baker` (which comes from `nk_font_atlas`) uses for all its memory allocations.
Removing these customizations, and insisting on not introducing them, has forced the use of separate methods for memory management in `stb_truetype`:
```C
#ifndef STBTT_malloc
static nk_handle fictional_handle = {0};

#define STBTT_malloc(x,u)  nk_malloc( fictional_handle, 0, x )
#define STBTT_free(x,u)    nk_mfree( fictional_handle , x)
#endif
```

`nk_malloc` and `nk_mfree` only exist when `NK_INCLUDE_DEFAULT_ALLOCATOR` is defined.

So the negative effect of this change is that the library no longer builds when `NK_INCLUDE_DEFAULT_ALLOCATOR` is not defined.

This presents a problem for libraries that wrap Nuklear and provide their own memory management, eg many bindings for other languages.


As an example, I am working on updating a binding from a Nuklear version before this PR to a Nuklear version after this PR.
It is preferred to not let Nuklear pull in stdlib for memory management, since the language in question (Rust, but the point of this comment is language-agnostic) can provide its own, and it makes no sense to have two separate ways of managing memory.
This was previously possible by creating a `nk_allocator` that points to custom alloc/free implementations that delegate to the language's implementations, and initializing a `nk_font_atlas` with this allocator; this allocator would be used for all font-related memory management in Nuklear, including in `stb_truetype`.

After this PR, `stb_truetype` no longer uses the `nk_allocator` of `nk_font_atlas`.
This means it must be provided with a separate implementation in the binding code.
This is not technically difficult (just implement a malloc and a free, and alias `STBTT_malloc` and `STBTT_free` to them), but is certainly cumbersome for the users of this library that actually prefer finer control over memory management (which, given the library's lightweight nature, is more than just those using bindings to other languages).

Another subtle difference is that, once `STBTT_malloc` and `STBTT_free` are dealt with as above, it is no longer possible to use two `nk_font_atlas` with two different allocators at the same time in the same manner as before this change, since `stb_truetype`'s memory management cannot switch between different allocators at run time (without manually implementing that, and introducing locking for multithreaded accesses if necessary, etc).
Using multiple `nk_font_atlas` in the same application simultaneously is not something that I have personally needed so far, and can't immediately think of a reason where this would be needed, but it is certainly worth noting.

dumblob: @asoft thanks for coming here and describing in detail what this change caused. What would you suggest we implement/change?

Btw. we're open to PRs (especially those increasing interoperability without much cost), so feel free to make one and we might discuss the details there.

2asoft: If allowing for finer control over mememory management is a goal of Nuklear, I think the current situation around memory management in stb_tt should be treated as a bug.
I'd like this bug to be fixed.

Options I see, in order of perceived implementation difficulty:
1. Add configuration to setup stb_tt library-wide to use a given allocator (via an alternative implementation of `STBTT_malloc` and `STBTT_free`).
1. Propagate `nk_font_atlas`'s allocator to stb_tt by _modifying Nuklear's copy of stb_tt_, adding the allocator parameter in all methods where `STBTT_malloc`/`STBTT_free` are used
1. Propagate `nk_font_atlas`'s allocator to stb_tt or `STBTT_malloc`/`STBTT_free` via some other means, without modifying the committed copy of stb_tt

Pros/cons that I see:
1. pros: simple to implement; simple to use\
cons: introduces yet another place for memory management; has all drawbacks mentioned in initial comment.
1. pros: simple to implement; simple to use; does not change amount of various memory management entry points \
cons: complicates stb_tt version updates
1. pros: simple to use; does not change amount of various memory management entry points \
cons: likely complicated implementation, cannot be implemented immediately, need to research options; 

I favor option 3, because it does not have the cons of options 1 and 2.
I'm playing locally with some options, don't have a good solution yet.

dumblob: Hm, how about reversing the dependency and propagate stb_tt's `STBTT_malloc/STBTT_free` to nk_font_atlas or something along these lines?

2asoft: I found a simple option 3 solution: #186.

PR #54, opened 2020-01-23
Fix 'large font' bug, and 'button image' bug

ronaaron:


ronaaron:
Fixed issue with 'large font' (e.g. CN with many glyphs at large font size), as well as the button image bug.

src/nuklear_button.c
```diff
@@ -198,7 +198,12 @@ nk_draw_button_symbol(struct nk_command_buffer *out,
     else if (state & NK_WIDGET_STATE_ACTIVED)
         sym = style->text_active;
     else sym = style->text_normal;
-    nk_draw_symbol(out, type, *content, bg, sym, 1, font);
+
+	/* don't let symbol escape button bounds */
+	struct nk_rect symbol_content = *content;
+	float f = NK_MIN(symbol_content.h, symbol_content.w);
```
dumblob, line 204: Hm, thinking about this a bit - do we need to force symbols to be rectangular?

src/nuklear_font.c
```diff
@@ -195,7 +195,12 @@ nk_font_bake_pack(struct nk_font_baker *baker,
         } while ((it = it->n) != config_iter);
     }
     *height = 0;
-    *width = (total_glyph_count > 1000) ? 1024 : 512;
+   
+	/* Be able to handle very large fonts */
+	int factor = total_glyph_count / 1024;
```
dumblob, line 200: Maybe also comment on the origin of this equation :wink:.

nuklear.h
```diff
@@ -5430,6 +5455,10 @@ struct nk_window {
     struct nk_window *next;
     struct nk_window *prev;
     struct nk_window *parent;
+
+	/* testing flags */
```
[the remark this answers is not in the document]
ronaaron: That change set isn't in the latest, it's old.

src/nuklear_button.c
ronaaron: Dunno.  I do know the existing behavior is not good.

src/nuklear_font.c
ronaaron: What origin? It was just a bit of guesswork. and it works to keep the display from turning all black when too big a font is loaded.

nuklear.h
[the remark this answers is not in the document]
dumblob: Yeah, don't know where it comes from - maybe a GitHub issue (mixing some old commits with new ones)? I didn't write it today :wink:.

src/nuklear_font.c
dumblob: Well, I meant the "guesswork" as you call it. I think there could be some dependency on font size etc., but don't know where to start (didn't do any analysis).

src/nuklear_button.c
dumblob: @hejsil thoughts?

src/nuklear_font.c
ronaaron: I'd love to tell you I did some deep analysis, but it was just a "hunch" that seems to play out. I was able to load huge font at a large size, when before I could only load some of the characters at a medium size.

It fixes the specific problem I was facing ,but I don't know if there's a better solution out there.

src/nuklear_button.c
Hejsil: I'm guessing symbols somehow ends up escaping the bounds of the button (a picture would be nice so I had an idea as to how this failure looks). This rectangle solution seems wrong to me tough. Shouldn't we just shrink the content by `style->button.border`? Ofc, this wont solve anything if the symbol clips at the corners and the corners are rounded.

dumblob: Shrinking sounds also reasonable (round corners are theoretically solvable by changing the symbol itself, so probably not a big issue).

src/nuklear_font.c
dumblob: @ronaaron you do not seem to be the only one having [a problem with bigger fonts](https://github.com/Immediate-Mode-UI/Nuklear/issues/77#issuecomment-605383306 ), so I'd definitely like to have a reasonable solution for this. I actually think even the original width computation was not correct. But instead of making it incorrect again, I'd rather suggest trying to consult the rect packing and font library from [STB](https://github.com/nothings/stb ).

Hejsil:
Any updates on this?

Issue #56, opened 2020-01-30
nk_window_is_closed no working


hbiblia:
https://github.com/vurtun/nuklear/issues/266

line: 15769
```
/* ... */
layout->flags |= NK_WINDOW_HIDDEN;
layout->flags |= NK_WINDOW_CLOSED;
layout->flags &= ~NK_WINDOW_MINIMIZED; // new line
/* ... */
```

**layout->flags |= NK_WINDOW_CLOSED;**

nk_window_is_closed does not work for me, looking in the previous Issue I found that someone reported it but this was not implemented.

Tested and it works, please test and implement.

PR #57, opened 2020-01-30
nk_window_is_closed no working 

hbiblia:
https://github.com/Immediate-Mode-UI/Nuklear/issues/56

nk_window_is_closed does not work for me, looking in the previous Issue I found that someone reported it but this was not implemented.


nuklear.h
```diff
@@ -15767,6 +15767,7 @@ nk_panel_begin(struct nk_context *ctx, const char *title, enum nk_panel_type pan
                 &style->window.header.close_button, in, style->font) && !(win->flags & NK_WINDOW_ROM))
             {
                 layout->flags |= NK_WINDOW_HIDDEN;
+                layout->flags |= NK_WINDOW_CLOSED;
```
Hejsil: Yea, this seems right. I actually think the `layout->flags |= NK_WINDOW_HIDDEN;` is wrong and should be removed as well.

Also this change needs to be done in [src/nuklear_panel.c](https://github.com/Immediate-Mode-UI/Nuklear/blob/master/src/nuklear_panel.c) so that the `paq` script does not override your change.

mgerhardy: After applying these changes, I got an assert in some other locations.

Hejsil:
This PR seem to have grown out of its original scope. I'm therefor gonna close it to keep the tracker decently clean. Feel free to open a new one.

PR #64, opened 2020-02-26
FIX: UTF-8 Backspace won't work in SFML OpenGL 3 Demo

generalzzd:
I recently get a problem in UTF8 SFML project that I will alway get a question mark when I input "Backspace" button. I'm using `nk_font_chinese_glyph_ranges` and customized ttf front style. After debugging the source code, I find out the TextEntered intercepter `if(evt->type == sf::Event::TextEntered) ` in "nuklear_sfml_gl3.h" will accept "Backspace" event and input unicode into textinput component. 

After checking old repositories, bharadwaj-raju has mentioned his solution.[https://github.com/vurtun/nuklear/pull/932](url) 
Just change and add this code:
	`if (evt->text.unicode != 8) {
			nk_input_unicode(ctx, evt->text.unicode);
		}`

And it works well after examining. So, I think I may put this question here and let more people know. 

Thanks.


demo/sfml_opengl3/nuklear_sfml_gl3.h
```diff
@@ -442,7 +442,9 @@ nk_sfml_handle_event(sf::Event* evt)
         } else nk_input_motion(ctx, evt->touch.x, evt->touch.y);
         return 1;
     } else if(evt->type == sf::Event::TextEntered) {
-        nk_input_unicode(ctx, evt->text.unicode);
+		if (evt->text.unicode != 8) {
```
dumblob: Please add explanation comment what this check means and then I'll merge this :wink:. Thanks!

demo/sfml_opengl3/nuklear_sfml_gl3.h
```diff
@@ -442,7 +442,7 @@ nk_sfml_handle_event(sf::Event* evt)
         } else nk_input_motion(ctx, evt->touch.x, evt->touch.y);
         return 1;
     } else if(evt->type == sf::Event::TextEntered) {
-		if (evt->text.unicode != 8) {  //this line will check user whether input BACKSPACE. 8 is the Unicode of BACKSPACE.  If true, the function will ignore text input.
+		if (evt->text.unicode != 8) {  //this line will check user whether input BACKSPACE. 8 is the Unicode of BACKSPACE.  If true, the function will ignore BACKSPACE input.
 			nk_input_unicode(ctx, evt->text.unicode);
 		}
```
generalzzd: This line will check user whether input BACKSPACE. 8 is the Unicode of BACKSPACE.  If true, the function will ignore BACKSPACE input.

demo/sfml_opengl3/nuklear_sfml_gl3.h
```diff
@@ -442,7 +442,9 @@ nk_sfml_handle_event(sf::Event* evt)
         } else nk_input_motion(ctx, evt->touch.x, evt->touch.y);
         return 1;
     } else if(evt->type == sf::Event::TextEntered) {
-        nk_input_unicode(ctx, evt->text.unicode);
+		if (evt->text.unicode != 8) {  //this line will check user whether input BACKSPACE. 8 is the Unicode of BACKSPACE.  If true, the function will ignore BACKSPACE input.
```
dumblob: It seems this header file uses C89 comments (`/* */`) - I'd prefer obeying that though I won't require that. Also I think a comment `8 ~ backspace` would be enough (all the rest is visible from the source itself or 100% clear from the context in which this everything happens) :wink:.

demo/sfml_opengl3/nuklear_sfml_gl3.h
```diff
@@ -442,7 +442,8 @@ nk_sfml_handle_event(sf::Event* evt)
         } else nk_input_motion(ctx, evt->touch.x, evt->touch.y);
         return 1;
     } else if(evt->type == sf::Event::TextEntered) {
-		if (evt->text.unicode != 8) {  //this line will check user whether input BACKSPACE. 8 is the Unicode of BACKSPACE.  If true, the function will ignore BACKSPACE input.
+		/* 8 ~ backspace */
+		if (evt->text.unicode != 8) {
```
generalzzd: 8 ~ backspace

dumblob:
Done. Thanks for your patience!

Issue #66, opened 2020-02-28
Building demo/sdl_opengl3 with msys2 causes conflicting types for 'SDL_main'


KenthJohan:
Building demo/sdl_opengl3 with MSYS2 causes:
```
$ make
rm -f bin/demo.exe
cc main.c -std=c99 -pedantic -O2 -o bin/demo.exe -lmingw32 -lSDL2main -lSDL2 -lopengl32 -lm -lGLU32 -lGLEW32
In file included from C:/msys64/mingw64/include/SDL2/SDL.h:32,
                 from main.c:14:
main.c:73:5: error: conflicting types for 'SDL_main'
   73 | int main(void)
      |     ^~~~
In file included from C:/msys64/mingw64/include/SDL2/SDL.h:32,
                 from main.c:14:
C:/msys64/mingw64/include/SDL2/SDL_main.h:121:29: note: previous declaration of 'SDL_main' was here
  121 | extern SDLMAIN_DECLSPEC int SDL_main(int argc, char *argv[]);
      |                             ^~~~~~~~
make: *** [Makefile:25: demo.exe] Error 1
```

It can be fixed adding arg params:
https://github.com/Immediate-Mode-UI/Nuklear/blob/e4d90b7db313aa2143aff275640a7532d5b07a8b/demo/sdl_opengl3/main.c#L73
```diff
- int main(void)
+ int main(int argc, char * argv[])
```

dumblob:
Yeah, please make a PR and I'll merge it ASAP. Thanks!

PR #67, opened 2020-02-28
Added main arg params in demo/sdl_opengl

KenthJohan:
#66 
This fixed the error: conflicting types for 'SDL_main' when using mingw


dumblob:
Done. Thanks!

PR #78, opened 2020-03-10
add info to Reviewers guide

crazyBaboon:
This info is missing from the reviewers guide.


Hejsil:
Well, I think there is a reasonable case to make that if you add a new API to nuklear a nice addition would be to use that API in the demos (though not nessesary).

For changes focused on the demoes I agree that nuklear probably shouldn't be touched.

dumblob:
I agree with @Hejsil . Could you @crazyBaboon rephrase it a little in the PR?

Hejsil:
I'm gonna close this as it seems we haven't been able to come up with something we're happy with. Feel free to open a new one if you come up with something.

PR #80, opened 2020-03-11
Fix bug where padding was subtracted from widget

### LINKED ISSUE #79: commit "Fix bug where width padding was applied twice" regression
Hi @Hejsil ,

Apparently, commit 14dd68ea0f4918ec3a77222ddb5c337dff4995f3 "Fix bug where width padding was applied twice" introduces a regression in the code. To visualise the bug introduced in this commit:

1 - build the demo glfw_opengl3 and run it;

2 - Go to overview -> Widgets -> Combo -> Fist

Note all the items text, Pistol, Fist, Shotgun are all squashed to the left, without any spacing and with the first capital letter disappearing halfway:

Ubuntu-Mate 18.04 - 64 bit.

Hejsil:
Fixes #79

The code deleted here seems to have been an effort to solve what 14dd68e solved so it makes sense that these would conflict. I'm pretty sure 14dd68e is the correct solution though and this was some old hack.

### DIFF
--- a/nuklear.h
+++ b/nuklear.h
@@ -22839,16 +22839,6 @@ nk_widget_fitting(struct nk_rect *bounds, struct nk_context *ctx,
     style = &ctx->style;
     layout = win->layout;
     state = nk_widget(bounds, ctx);
-
-    panel_padding = nk_panel_get_padding(style, layout->type);
-    if (layout->row.index == 1) {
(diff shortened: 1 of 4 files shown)


Hejsil:
Before:
image

After:
image

dumblob:
Thanks!

Issue #70, opened 2020-03-04
Can we add nk_property_ull for 64 bit unsigned integers?


KenthJohan:
I guess this would be the necessary step to support a `nk_property_ull(struct nk_context *ctx, const char *name, unsigned long long min, unsigned long long *val, unsigned long long max, unsigned long long step, float inc_per_pixel)` function:
```diff
enum nk_property_kind {
    NK_PROPERTY_INT,
    NK_PROPERTY_FLOAT,
	NK_PROPERTY_DOUBLE,
+	NK_PROPERTY_ULL
};
```
```diff
union nk_property {
    int i;
    float f;
    double d;
+	unsigned long long ull;
};
```

dumblob:
Generally I'm not against 64 (or more) bit widths nor against adding new property functions to Nuklear. I'm though quite concerned about *unsigned* as all other functions were deliberately signed. Would it work for your use case to use signed 64 bit integers?

There should be also a "friend" function analogous to `nk_propertyi()`.

PR #84, opened 2020-03-16
Added nk_property_ull for unsigned long long integers

KenthJohan:


src/nuklear_internal.h
```diff
@@ -1,6 +1,8 @@
 #ifndef NK_INTERNAL_H
 #define NK_INTERNAL_H
 
+#include "nuklear.h"
```
dumblob: Is this a typo or am I missing something?

src/nuklear_internal.h
```diff
@@ -305,14 +308,16 @@ enum nk_property_filter {
     NK_FILTER_FLOAT
 };
 enum nk_property_kind {
-    NK_PROPERTY_INT,
-    NK_PROPERTY_FLOAT,
-    NK_PROPERTY_DOUBLE
+	NK_PROPERTY_INT,
```
dumblob: Please don't change code style if possible (see our recommended [review guidelines](https://github.com/Immediate-Mode-UI/Nuklear#reviewers-guide ) which our reviewers of PRs are encouraged to follow).

KenthJohan: Ok I will fix that.

src/nuklear_internal.h
KenthJohan: Many definitions in `nuklear_internal.h` depends on `nuklear.h`. Maybe it is not supposed to be parsed by from IDE intellisense. I tried with qt-creator and it couldn't parse it.

dumblob: The point is, that the sources in `src/` are first being amalgamated (concatenated) into `nuklear.h` using `src/paq.sh` (or `src/paq.bat` on Windows). So you must not do anything like this. See the reviewers guide for an overview.

KenthJohan:
> 
> 
> Hi @KenthJohan, thanks for your effort. I'll start with a dumb, but quite important question.
> 
>     1. What's the whole purpose of this PR? Do you have any specific use cases? What fails for you without this PR?
I am using Nuklear with Kvaser CAN lib to send message every time the user changes a value.
The problem is that there is one CAN message that consist of one `uint64_t` value and I don't want to use 4 of `nk_property_int` to make up a `uint64_t` it would not be user friendly imo.

src/nuklear_internal.h
KenthJohan: I know there is two way to use Nuklear: header only `nuklear.h` or multiple source files `src/nuklear.h`. I understand that Nuklear is only supposed to be used as header only but for me it is easier to navigate around code with the source files than header only `nuklear.h`.

dumblob: To clarify:
a) There is only one way to use Nuklear - namely the `nuklear.h` header.
b) For the sole purpose of **development of Nuklear itself**, the sources are split and maintained in `src/`.

So you absolutely can not use for your app anything in `src/`. Nuklear is a **single header** library, nothing else :wink:.

dumblob:
>I am using Nuklear with Kvaser CAN lib to send message every time the user changes a value.
The problem is that there is one CAN message that consist of one uint64_t value and I don't want to use 4 of nk_property_int to make up a uint64_t it would not be user friendly imo.

Ok, the use case sounds plausible, but I'm still not sure we need this as `unsigned` integers are extremely dangerous (due to their typical property of overflow) and the world moves steadily away from them. Especially in the context of min-max, increments and decrements (which is exactly what Nuklear properties are designed for). Let's discuss it with other devs. @Hejsil ideas?

Btw. if we come to conclusion that it makes sense to incorporate, I'll definitely expect `assert()`s "everywhere" and also `nk_propertyu()` function (other _int _float _double functions have also a friend :wink:).

src/nuklear_internal.h
KenthJohan: I like header only files but Nuklear is too large for my taste. If Nuklear were split into multiple header only files I would use them instead.

dumblob: @keharriso ok, it's up to you. Look at `src/paq.sh` how it's being amalgamated and try to amalgamate your own small version. If you'll come up with something very simple, don't hesitate and share your work with us as it might be something to contemplate for the future.

Btw. have you seen https://github.com/rxi/microui ?

KenthJohan:
> 
> 
> > I am using Nuklear with Kvaser CAN lib to send message every time the user changes a value.
> > The problem is that there is one CAN message that consist of one uint64_t value and I don't want to use 4 of nk_property_int to make up a uint64_t it would not be user friendly imo.
> 
> Ok, the use case sounds plausible, but I'm still not sure we need this as `unsigned` integers are extremely dangerous (due to their typical property of overflow) and the world moves steadily away from them. Especially in the context of min-max, increments and decrements (which is exactly what Nuklear properties are designed for). Let's discuss it with other devs. @Hejsil ideas?
> 
> Btw. if we come to conclusion that it makes sense to incorporate, I'll definitely expect `assert()`s "everywhere" and also `nk_propertyu()` function (other _int _float _double functions have also a friend 😉).

Is signed integers were overflowing and unsigned do expected wrap around.

KenthJohan:
> 
> 
> > I am using Nuklear with Kvaser CAN lib to send message every time the user changes a value.
> > The problem is that there is one CAN message that consist of one uint64_t value and I don't want to use 4 of nk_property_int to make up a uint64_t it would not be user friendly imo.
> 
> Ok, the use case sounds plausible, but I'm still not sure we need this as `unsigned` integers are extremely dangerous (due to their typical property of overflow) and the world moves steadily away from them. Especially in the context of min-max, increments and decrements (which is exactly what Nuklear properties are designed for). Let's discuss it with other devs. @Hejsil ideas?
> 
> Btw. if we come to conclusion that it makes sense to incorporate, I'll definitely expect `assert()`s "everywhere" and also `nk_propertyu()` function (other _int _float _double functions have also a friend 😉).

The signed integers overflows and unsigned wraps around expectedly.

src/nuklear_internal.h
KenthJohan: No. Thanks for the tip. I have to think about this.

KenthJohan:
Maybe keep this discussion in https://github.com/Immediate-Mode-UI/Nuklear/issues/70?

PR #88, opened 2020-03-20
Fixed compiler warnings - bring your own implementations and unused functions

mgerhardy:


nuklear.h
```diff
@@ -5726,15 +5726,6 @@ static nk_handle fictional_handle = {0};
...
 
-#ifndef NK_MEMSET
-#define NK_MEMSET nk_memset
-#endif
-#ifndef NK_MEMCPY
-#define NK_MEMCPY nk_memcopy
-#endif
-#ifndef NK_SQRT
-#define NK_SQRT nk_sqrt
-#endif
```
dumblob: Shouldn't we extend this to all other `nk_` function which are separately `#ifdef`ed? I think those are `nk_sin()` `nk_cos()` `nk_strtod()` and `nk_dtoa()`.

mgerhardy: Can add those, too - if the way I did fix the others is what would be accepted.

dumblob: >if the way I did fix the others is what would be accepted.

I think it's OK as in the documentation they're not scattered, but on one place. So we don't need to have them at one place in the source.

mgerhardy:
I think I would call this done if nothing else comes up in a new review.

nuklear.h
```diff
@@ -3730,7 +3730,10 @@ NK_API int nk_stricmp(const char *s1, const char *s2);
 NK_API int nk_stricmpn(const char *s1, const char *s2, int n);
 NK_API int nk_strtoi(const char *str, const char **endptr);
 NK_API float nk_strtof(const char *str, const char **endptr);
+#ifndef NK_STRTOD
+#define NK_STRTOD nk_strtod
```
Hejsil, line 3734: `nk_strtod` only ifdefs out the declarations while `nk_memcpy` ifdefs out both the declaration and definition. Should they all ifdef out the definition or?

mgerhardy: Ah of course - sorry - will fix it.

mgerhardy: Fixed now - thanks for reporting.

src/CHANGELOG
```diff
@@ -8,6 +8,9 @@
 ///    - [yy]: Minor version with non-breaking API and library changes
 ///    - [zz]: Bug fix version with no direct changes to API
 ///
+/// - 2020/04/09 (4.02.1) - Removed unused nk_sqrt function to fix compiler warnings
+/// - 2020/04/09 (4.01.11) - Fixed compiler warnings if you bring your own methods for
+///                        nk_cos/nk_sin/nk_strtod/nk_memset/nk_memcopy/nk_dtoa
```
dumblob, line 13: According to the commit changelog, the version `4.01.11` is not separately "checkoutable", so it's really confusing to have it separately in the changelog. Please either make both versions (`4.01.11` `4.02.1`) separately checkoutable or bump the version in this PR just once and squash the version comment messages :wink:.

mgerhardy: Merged everything into one commit.

dumblob:
Thanks for your persistence!

PR #89, opened 2020-03-20
Use a shared makefile to set the compile flags and enable more warnings

mgerhardy:
projects that are using the library might have enabled compiler flags that would spam them with
warnings - so we should try to keep the warnings as low as possible

### DIFF
--- a/demo/Makefile.inc
+++ b/demo/Makefile.inc
@@ -1,0 +1,2 @@
+# Flags
+CFLAGS += -std=c99 -pedantic -Wall -Wextra -O2
--- a/demo/allegro5/Makefile
+++ b/demo/allegro5/Makefile
@@ -1,8 +1,8 @@
 # Install
 BIN = demo
 
-# Flags
-CFLAGS += -std=c99 -pedantic -O2
+PARENT_DIR := $(abspath $(dir $(lastword $(MAKEFILE_LIST)))/..)
+include $(PARENT_DIR)/Makefile.inc
 
 SRC = main.c
 OBJ = $(SRC:.c=.o)
--- a/demo/glfw_opengl2/Makefile
+++ b/demo/glfw_opengl2/Makefile
@@ -1,8 +1,8 @@
 # Install
 BIN = demo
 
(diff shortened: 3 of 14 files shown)


demo/allegro5/Makefile
```diff
@@ -1,8 +1,8 @@
 # Install
 BIN = demo
 
-# Flags
-CFLAGS += -std=c99 -pedantic -O2
+PARENT_DIR := $(abspath $(dir $(lastword $(MAKEFILE_LIST)))/..)
```
dumblob: Have you tested this on *nix (including macOS) platforms as well as Windows (MSVC)? I'm afraid this won't work well on all of them (or maybe I'm too pessimistic :wink:).

demo/Makefile.inc
```diff
@@ -0,0 +1,2 @@
+# Flags
+CFLAGS += -std=c99 -pedantic -Wall -Wextra -O2
```
dumblob: Wouldn't it be simpler to just add this as recommendation/requirement to the [Reviewers Guide](https://github.com/Immediate-Mode-UI/Nuklear#reviewers-guide ) instead of complicating the build process and introducing a new file for such a simple purpose?

mgerhardy: Is it wanted to print those warnings? If so, I can also include the flags in every makefile - no problem. But I would first like to discuss the flags.

dumblob: Generally the stricter the better. Though we do **not require** such flags for non-`nuklear.h` stuff as stated in the Reviewers Guide. It's just a preference. If it makes some `demo/`s or `example/`s issue a warning, then these should be fixed together with introduction of these flags.

Regarding each flag separately, I think some demos do not comply with C99, but maybe with newer version or some dialect (GNU extensions etc.) which means it might really need some rework.

mgerhardy:
Closing this one in favor of the new https://github.com/Immediate-Mode-UI/Nuklear/pull/101
I hope to simplify the steps that are needed to compile and test the code.

PR #100, opened 2020-04-09
Fixed compiler warnings - removed unused variables

mgerhardy:
Didn't bump the version as this is no real change.


Hejsil:
Hmm, yeah I wonder if we wonna bump the version for changes like this. On one hand, no the code did not semantically change, but on the other hand certain builds of nuklear will now succeed with `-Werror`.

dumblob:
>Hmm, yeah I wonder if we wonna bump the version for changes like this.

I don't think it's necessary. There are always exceptions to rules :wink:. On the other hand, I'm not against bumping it as it shouldn't cause any harm :wink:.

Hejsil:
Hmm. Alright, I have no strong feelings either way. This PR looks good to me!

dumblob:
Thanks @mgerhardy!

PR #101, opened 2020-04-09
Main makefile to perform the needed steps for each change.

mgerhardy:
It updates the `nuklear.h` and the `nuklear.html` files as well as tries to compile the example and some of the demos.

I would even go further and only provide one makefile - because that's easier to maintain. But that's on the maintainers of this project. I would opt-in to perform that change.

I've only included the glfw demos because all the others don't compile out of the box on my system. E.g. the SDL demos are using the wrong include paths for the headers. One should include `<SDL.h>` not `<SDL2/SDL.h>`. One global makefile could solve all this by providing the needed configuration of `CFLAGS` and `LDFLAGS` for every demo (to be fair - this would of course also be possible with shared makefiles - but would be a lot of code duplication). Adding `pkg-config` support and stuff like that would make contributors lifes a lot easier. Because it would increase the chance that the code builds right away on their system.


Hejsil:
I like the idea of having a makefile that does some of the work we require people to do, like auto generating `nuklear.h` and doing the docs (which I actually didn't know we had, and needed to generated).

The "one makefile to build all demos" might be a little too hard to acomplish. I wouldn't be able to build any of the windows only demos on my linux machine for an example. Having a makefile the automated tests can run to check all this would be pretty nice though.

mgerhardy:
Sharing the compiler flags and toolchain parameters would make the cross-compilation with e.g. mxe a lot easier, too. I could try to implement that, too (already did it for other projects and it worked quite nice)

mgerhardy:
... but that should not be part of this PR anyway. This one should just make the contributors life a little bit easier by not missing any needed step.

mgerhardy:
Any news on this one?

Makefile
```diff
@@ -0,0 +1,16 @@
+all: doc example demo
```
dumblob, line 1: I'd say any Makefile which does more than one thing (i.e. has more than one target) should have help as default operation. The help shouldn't be verbose - if target names are self-explanatory, then the whole help could be a one liner with invocation syntax listing all targets.

Makefile
```diff
@@ -0,0 +1,16 @@
...
+
+paq:
+	cd src && ./paq.sh > ../nuklear.h
+
+doc: paq
+	$(MAKE) -C doc
+	cd doc && ./build.sh
```
dumblob, line 8: `build.sh` should get a Windows `bat` equivalent.

Makefile
```diff
@@ -0,0 +1,16 @@
...
+
+example: paq
```
dumblob, line 10: Building example(s) shouldn't have `paq` as dependency as the current `nuklear.h` is always in this upstream repo. I understand that when testing, one needs to rebuild e.g. some demos, but that's not the point of this general Makefile and definitely not the standard use case of `examples/`.

Makefile
```diff
@@ -0,0 +1,16 @@
...
+	$(MAKE) -C example
+
+demo: paq
```
dumblob, line 13: Same as above - there shouldn't be any `paq` dependency.

Makefile
```diff
@@ -0,0 +1,16 @@
...
+	$(MAKE) -C demo/glfw_opengl2
+	$(MAKE) -C demo/glfw_opengl3
```
dumblob, line 15: This is not enough. I think on unices all non-Windows demos should be tried to build, but if any fails, it shouldn't stop `make` from proceeding, but instead issue a warning with recommendation for common issues (e.g. `WARN could not build glfw_opengl3 - do you have glfw and libc installed?`).

On Windows this target should try to build all Windows demos with same error handling like I described for unices.

PR #112, opened 2020-04-14
Arc Command Support for GDI & X + Bugfix

Vucay:
This adds arc support (both stroked and filled) to the X11 and GDI backends of Nuklear. 

A bugfix is also included for the GDI backend where stroked circles used to be filled with a black background. Now only the actual stroke is drawn. Since this was fixed with just one line of code I decided to put it in this PR.

### DIFF
--- a/demo/gdi/nuklear_gdi.h
+++ b/demo/gdi/nuklear_gdi.h
@@ -399,6 +399,7 @@ nk_gdi_stroke_circle(HDC dc, short x, short y, unsigned short w,
         SelectObject(dc, pen);
     }
 
+    HGDIOBJ br = SelectObject(dc, GetStockObject(NULL_BRUSH));
     SetDCBrushColor(dc, OPAQUE);
     Ellipse(dc, x, y, x + w, y + h);
 
--- a/demo/gdi/nuklear_gdi.h
+++ b/demo/gdi/nuklear_gdi.h
@@ -410,6 +410,61 @@ nk_gdi_stroke_circle(HDC dc, short x, short y, unsigned short w,
 }
 
 static void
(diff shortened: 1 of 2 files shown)


demo/gdi/nuklear_gdi.h
```diff
@@ -408,6 +409,61 @@ nk_gdi_stroke_circle(HDC dc, short x, short y, unsigned short w,
...
+    else
+    {
+    	DWORD pen_style = PS_SOLID | PS_ENDCAP_FLAT | PS_GEOMETRIC; /* without that endcap round caps are used which looks really weird for thick arcs */
+
+    	LOGBRUSH brush;
+		brush.lbStyle = BS_SOLID;
+		brush.lbColor = color;
+		brush.lbHatch = 0;
+
+    	pen = ExtCreatePen(pen_style, line_thickness, &brush, 0, NULL);
```
Hejsil, line 430: There seems to be some inconsistent spacing in this file generally. Idk if it's a mix of tabs and spaces.

Also, I wounder if there is a way we can reuse this `pen` or something to avoid allocating and freeing it on each call to this function (I assume `ExtCreatePen` allocates memory, since it has to be deleted). I don't think this is a showstopper for getting this merged, but it would be an improvement IMHO.

Vucay: Many functions in the GDI backend seem to do this but with CreatePen instead of ExtCreatePen; basically a simpler pen constructor that requires a deletion just like ExtCreatePen.
It might be a good idea to optimize this in an extra pull request if possible since it affects more than just the new functions.

Hejsil:
Going through PRs to get as many merged as I can. I wanna test this before merging, but there seem to be no demo that uses this. Do you have some code I can test with?

RobLoach:
https://github.com/Immediate-Mode-UI/Nuklear/pull/385

Issue #110, opened 2020-04-12
Possible to get nk_combo (Combobox) "DropUp"?


neonr-0:
It's possible to get nk_combo_begin_label (as example) get "drop up" beside default dropdown?
Example code:
```
	if (nk_begin(&ctx, "StatusBar", nk_rect(0, height-80, width, 80), NK_WINDOW_NO_SCROLLBAR))
	{
		nk_layout_row_dynamic(&ctx, 30, 3);
		nk_label(&ctx, g_GUIData.strPartsCount, NK_TEXT_LEFT);
		nk_label(&ctx, g_GUIData.strParamsCount, NK_TEXT_LEFT);
		static int selected_item = 0;
		static const char *items[] = { "Item 0","item 1","item 2" };
		if (nk_combo_begin_label(&ctx, items[selected_item], nk_vec2(nk_widget_width(&ctx), 35))) {
			nk_layout_row_dynamic(&ctx, 35, 1);
			for (int i = 0; i < 3; ++i)
				if (nk_combo_item_label(&ctx, items[i], NK_TEXT_LEFT))
					selected_item = i;
			nk_combo_end(&ctx);
		}
		nk_end(&ctx);
	}
```
On screenshot dropdown combobox. Can I set layout row upper uses absolute coordinates, it's possible?
![example](https://user-images.githubusercontent.com/6906978/79077107-8ce0b080-7d07-11ea-98d1-19474aabf674.png)

dumblob:
Didn't have this idea before, but why not to try it? There might be some clipping, etc. issues, but generally it should be possible. Nuklear is quite hackable, so don't hesitate and check the guts :wink:. PRs welcome!

neonr-0:
Found in guts really easy solution but need to be somehow marked (flag it in before call function?) because I don't find any windows size to check if combobox start overlapping on bottom on the screen (window).
In nk_combo_begin:
```
body.y = header.y + header.h-ctx->style.window.combo_border;
```
I change to:
```
body.y = header.y - ctx->style.window.combo_border - size.y;
```
Yep as result it's seems to work just fine :)
![example](https://user-images.githubusercontent.com/6906978/79228614-6ab37380-7e6a-11ea-99ac-4026d7d133de.png)

PR #113, opened 2020-04-14
Add to ComboBox "DropUp" if no space available to DropDown.

neonr-0:
Add ComboBox "DropUp" if no space available in layout height. 
#110

### DIFF
--- a/nuklear.h
+++ b/nuklear.h
@@ -28188,7 +28188,10 @@ nk_combo_begin(struct nk_context *ctx, struct nk_window *win,
     popup = win->popup.win;
     body.x = header.x;
     body.w = size.x;
-    body.y = header.y + header.h-ctx->style.window.combo_border;
+	if (win->bounds.h < header.h + size.y) //DropUp
+		body.y = header.y - ctx->style.window.combo_border - size.y;
+	else //DropDown
+		body.y = header.y + header.h - ctx->style.window.combo_border;
     body.h = size.y;
 
     hash = win->popup.combo_count++;


dumblob:
Thanks for the prompt response. Before we'll dive into this, could you please check how [PRs are getting reviewed](https://github.com/Immediate-Mode-UI/Nuklear#reviewers-guide )?

Hejsil:
Hmm. I'm not sure if this is quite the right approach. I guess that depends on the wanted behavior. Currently, the dropdown is allowed to go out of the window, and this isn't really a problem unless it actually escapes the programs window.
out

I'd say the best approach might be to break the combo box api to allow specifying whether the thing drops down or up. Maybe something like this:
```c
enum nk_drop_direction {
    NK_DROP_UP,
    NK_DROP_DOWN,
    NK_DROP_AUTO = NK_DROP_AUTO_PREFER_DOWN,

    /* Drops down, unless there isn't room in the window, in which case it drops up */
    NK_DROP_AUTO_PREFER_DOWN,
    /* Drops up, unless there isn't room in the window, in which case it drops down */
    NK_DROP_AUTO_PREFER_UP,
}
```

neonr-0:
It not break if we set before call nk_combo_begin
As example nk_combo_set_direction(NK_DROP_AUTO_PREFER_UP)
But it will create more complexity...

> 
> 
> Thanks for the prompt response. Before we'll dive into this, could you please check how [PRs are getting reviewed](https://github.com/Immediate-Mode-UI/Nuklear#reviewers-guide)?

yep commentary in code excessless...

Hejsil:
> It not break if we set before call nk_combo_begin
> As example nk_combo_set_direction(NK_DROP_AUTO_PREFER_UP)
> But it will create more complexity...

Hmm, yea that could work. I do think passing it in directly is a better API choice, and we're are not afraid of doing breaking changes as long as it is versioned appropriately. I'll leave it up to you whether you want to break API or not. I think both approaches are fine.

neonr-0:
If we Break API need more changes for some reason, not just add ComboBox to display mode....
But I think change API also better. 
(In my plans also very needed function add ctrl+a (Select All) combination to nk_edit_string but it's seems to be not breaking API changes)

neonr-0:
Something like this.
output.

Hejsil:
Very cool. I'll see if I can take a look at this code over the weekend. I'll assume this new API is good unless someone else complains :)

neonr-0:
Found something doing wrong in my code...
If we pass to combobox bigger height it clip in "ctx->popup" layout.
On drop down works fine because it's clip it by height (start point at y is correct) but on drop up start point at y based input height.
I need figure out and understand how to move start y position on all combobox "popup-ed" layouts based on clipping height...
tmp2
Of course we can precalculate height based on elements in combo-box, but this it not solution.
Well need some more time then I expected :(
But in my setup it work... but it need to work not only in my setup :)

PR #120, opened 2020-04-29
fix: 'nk_edit' frame border drawing

exchg:
Draw frame border after background to prevent overlapping.

At least for `x11` and `gdi` back-ends.

Before:
x11demo-before

After:
x11demo-after

### DIFF
--- a/nuklear.h
+++ b/nuklear.h
@@ -29091,6 +29091,7 @@ nk_tooltipfv(struct nk_context *ctx, const char *fmt, va_list args)
 ///    - [yy]: Minor version with non-breaking API and library changes
 ///    - [zz]: Bug fix version with no direct changes to API
 ///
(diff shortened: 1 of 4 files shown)


Hejsil:
Looks good. You only need to bump the minor number in [package.json](https://github.com/Immediate-Mode-UI/Nuklear/blob/master/package.json) and add your change to the [CHANGELOG](https://github.com/Immediate-Mode-UI/Nuklear/blob/master/src/CHANGELOG).

exchg:
> Looks good. You only need to bump 

Ah, done.

Hejsil:
Yep, looks good. If @dumblob agrees then this can be merged

dumblob:
Sure, LGTM. Thanks!

PR #124, opened 2020-04-30
nuklear_glfw_gl3.h without implementation

riri:
Using this header without the implementation is not possible because needed structs are not defined in the header only part. This PR solves this problem by making `struct nk_glfw` and `struct nk_glfw_device` public.

Note that a source implementing this backend still needs to include nuklear implementation too, because it uses some macros defined only in the implementation part of nuklear (i.e. `NK_MEMSET`). I have not been as deep as reviewing that part (not being sure of what should be done).

### DIFF
--- a/demo/glfw_opengl3/main.c
+++ b/demo/glfw_opengl3/main.c
@@ -75,7 +75,7 @@ static void error_callback(int e, const char *d)
 int main(void)
 {
     /* Platform */
-    struct nk_glfw glfw = {0};
+    struct nk_glfw* glfw = {0};
     static GLFWwindow *win;
     int width = 0, height = 0;
     struct nk_context *ctx;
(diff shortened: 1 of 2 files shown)


riri:
I have just changed how to fix it. Instead of exposing the nk_glfw struct, I made it allocated on the heap by nk_glfw3_init(), and freed by nk_glfw3_shutdown().

That makes an interface change in the init function, but that's not a big deal.

riri:
That makes sense too :) Would you merge the PR at the selected commit, or I have to push back myself on my fork?

Hejsil:
It would be easiest for me if you dropped the commit on your fork. That way I can just do the merge through github :+1:

riri:
Done :)

Hejsil:
Thanks!

PR #137, opened 2020-05-20
fix nk_do_edit: Keep scroll position when re-activating edit widget.

lundril:
How to test:
Construct two edit widgets.
Enter a long text into the first edit widget, so that the widget autoscrolls.
Click into second (still empty) edit widget.
Click into first edit widget again.
=> Text jums back, so that first character of text in first edit widget is visible again.
This patch keeps the scroll position of the edit widget constant.

### DIFF
--- a/nuklear.h
+++ b/nuklear.h
@@ -26535,7 +26535,10 @@ nk_do_edit(nk_flags *state, struct nk_command_buffer *out,
     if (!prev_state && edit->active) {
         const enum nk_text_edit_type type = (flags & NK_EDIT_MULTILINE) ?
             NK_TEXT_EDIT_MULTI_LINE: NK_TEXT_EDIT_SINGLE_LINE;
+        /* keep scroll position when re-activating edit widget */
+        struct nk_vec2 oldscrollbar = edit->scrollbar;
         nk_textedit_clear_state(edit, type, filter);
+        edit->scrollbar = oldscrollbar;
         if (flags & NK_EDIT_AUTO_SELECT)
             select_all = nk_true;
         if (flags & NK_EDIT_GOTO_END_ON_ACTIVATE) {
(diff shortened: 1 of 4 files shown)


Hejsil:
Haven't tried to reproduce this and check if the solution is right. I might look into this over the weekend. There are a few things the PR is missing regardless:
* You need to run the `paq` script in `src` to regenerate `nuklear.h`
* The version needs to be bumped in [package.json](https://github.com/Immediate-Mode-UI/Nuklear/blob/master/package.json) and [CHANGELOG](https://github.com/Immediate-Mode-UI/Nuklear/blob/master/src/CHANGELOG).

lundril:
Hello,

> * You need to run the `paq` script in `src` to regenerate `nuklear.h`
> * The version needs to be bumped in
> [package.json](https://github.com/Immediate-Mode-UI/Nuklear/blob/master/package.json) and
> [CHANGELOG](https://github.com/Immediate-Mode-UI/Nuklear/blob/master/src/CHANGELOG).

I updated this and force-pushed the update (I hope this is the right way).


> Haven't tried to reproduce this and check if the solution is right. I might look into this over the weekend.

If you want to try out:
On my fork I created a branch, which contains Test code for X11 (Linux).

  git repo: https://github.com/lundril/Nuklear-1.git
  branch:   edittest_orig

Then:
  cd demo/x11
  make
  ./bin/zahnrad

You will see a sub-window called "Edit Test", which should contain two edit widgets.
1) Enter a LOOOOOONG string into the upper edit widget (so that this edit widget autoscrolls)
2) Click into the lower edit widget (don't enter anything)
3) Click into the upper edit widget again
   => upper edit widget auto-scrolls back to initial position.

That's what my patch tries to fix.

so long
  Ingo

Hejsil:
I've tested this and confirmed that it works. All the changes look good to me. If @dumblob agrees then we can merge this.

dumblob:
LGTM, please merge. Thanks for this PR and review guys!

PR #139, opened 2020-05-26
Added image support to allegro5 backend

lockie:
Hello!
This PR adds support for `NK_COMMAND_IMAGE` draw command in allegro5 backend along with two helper functions to create and destroy allegro images stored in `nk_image` struct.


Hejsil:
Looks good!

PR #140, opened 2020-05-27
Fixed naming bug in allegro5 backend image support

lockie:
Hello!
This is follow-up to #139 . I am so very sorry, I messed up the function name to load the bitmap, so it is not exported when Nuklear is built as dynamic library (and hence is unusable). This PR fixes it.


Hejsil:
Np, I missed it to so I'm partly to blame. Good thing it was found early :+1:

PR #145, opened 2020-06-05
Added ability to use user fonts after initialization in allegro5 backend

lockie:
Hello!
Currently Allegro5 backend allows to set custom font on initialization using [`nk_allegro5_font_create_from_file`](https://github.com/Immediate-Mode-UI/Nuklear/blob/bffe1037b2de63bd05f040f58b32bfca9877689e/demo/allegro5/nuklear_allegro5.h#L34) that wraps `nk_user_font` struct into [`NkAllegro5Font`](https://github.com/Immediate-Mode-UI/Nuklear/blob/bffe1037b2de63bd05f040f58b32bfca9877689e/demo/allegro5/nuklear_allegro5.h#L53) custom bookkeeping struct, but after that the font only could be set globally using [`nk_allegro5_font_set_font`](https://github.com/Immediate-Mode-UI/Nuklear/blob/bffe1037b2de63bd05f040f58b32bfca9877689e/demo/allegro5/nuklear_allegro5.h#L36) (which [wraps `nk_style_set_font`](https://github.com/Immediate-Mode-UI/Nuklear/blob/bffe1037b2de63bd05f040f58b32bfca9877689e/demo/allegro5/nuklear_allegro5.h#L142-L148)). This small PR adds helper function that allows to get `nk_user_font` struct from that `NkAllegro5Font` bookkeeping struct to use it in different contexts, e.g. in `nk_style_push_font`.

### DIFF
--- a/demo/allegro5/nuklear_allegro5.h
+++ b/demo/allegro5/nuklear_allegro5.h
@@ -33,7 +33,8 @@ NK_API void                   nk_allegro5_del_image(struct nk_image* image);
 /* Fonts. We wrap normal allegro fonts in some nuklear book keeping */
 NK_API NkAllegro5Font*        nk_allegro5_font_create_from_file(const char *file_name, int font_size, int flags);
 NK_API void                   nk_allegro5_font_del(NkAllegro5Font *font);
-NK_API struct nk_user_font*   nk_allegro5_font(NkAllegro5Font *allegro5font);
+/* NOTE : just use NkAllegro5Font instead of nk_user_font,
+    since the former just extends the latter*/
 NK_API void                   nk_allegro5_font_set_font(NkAllegro5Font *font);
 
 #endif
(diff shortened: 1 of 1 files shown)


Hejsil:
Shouldn't this logic just be in `nk_allegro5_font_create_from_file`? Also why does `NkAllegro5Font` have a `height` field? Couldn't the `height` field in `nk_user_font` just be used?

lockie:
> Shouldn't this logic just be in nk_allegro5_font_create_from_file?

I had this thought, but decided to leave original backend code as intact as possible.
All right, I'll make those changes a bit later.

lockie:
All right, so I've removed unnecessary `height` field from `NkAllegro5Font` and moved the `nk_user_font` initialization code right into `nk_allegro5_font_create_from_file`. So now there's no need in my initial helper `nk_allegro5_font`, since `NkAllegro5Font` struct starts with `struct nk_user_font` and thus is interchangable with `nk_user_font`, so I've also removed the helper.

I've battle-tested the code in my application, and it works without hiccup.

Hejsil:
Looks good!

lockie:
Thanks!

PR #149, opened 2020-06-11
use integer for integer bound

wyattscarpenter:
I got a warning about this from my compiler (cl.exe). I tested it both ways, and the behavior is the same, except for the compiler warning no longer occurs after my change.


Hejsil:
Looks good!

PR #151, opened 2020-06-13
fix bugs in computing pooled allocation sizes

pmcmorris:
- Buffer size was under counted by using nk_page_data rather than nk_page_element.
- Buffer size was over counted by sizing for an unused element (page header contains the first one).
- Original code was not unsafe as the overcounting was larger than the under counting for the default pool size.
- Sized pages based on pool capacity rather than hard-coded value since it's configurable.
- Added and assert for safety.

### DIFF
--- a/demo/glfw_opengl3/nuklear_glfw_gl3.h
+++ b/demo/glfw_opengl3/nuklear_glfw_gl3.h
@@ -492,7 +492,7 @@ void nk_glfw3_shutdown(struct nk_glfw* glfw)
     nk_font_atlas_clear(&glfw->atlas);
     nk_free(&glfw->ctx);
     nk_glfw3_device_destroy(glfw);
-    memset(glfw, 0, sizeof(glfw));
+    memset(glfw, 0, sizeof(*glfw));
 }
 
 #endif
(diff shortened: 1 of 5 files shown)


pmcmorris:
I don't know why the changes from the previous PR showed up in here. :P

dumblob:
Hi and thanks for spotting this! We try to review PRs at least according to our [Review Guidelines](https://github.com/Immediate-Mode-UI/Nuklear#reviewers-guide ) - could you please adjust your PR accordingly before we'll take a look?

Commits from any other PR shouldn't do any harm if squashing them in the end ;)

pmcmorris:
I've read the guidelines and added missing entry in the changelog.  Let me know if anything else is missing.

dumblob:
Have taken a look and have one more plea - namely to add a comment as for a newcomer it's not apparent why the offset has to be off by `1`. Otherwise it LGTM :wink:.

@Hejsil could you also take a look as a second pair of eyes?

Hejsil:
The CHANGELOG doesn't appear in nuklear.h, which means that paq.sh was run after CHANGELOG was edited. Otherwise this looks good to me :+1:

pmcmorris:
Sorry for the delay, added a comment about the sizing, re-ran the paq script.

Hejsil:
Thanks!

PR #152, opened 2020-06-16
Added processed event flag as a return value in allegro5 backend

lockie:
Hello!
This PR adds return value to `nk_allegro5_handle_event`, which now returns boolean (actually integer) indicating whether the event was processed or not. This is similar to other backends; it allows to properly process events in games (i.e. not handle the event in-game if it was handled by Nuklear).


dumblob:
LGTM, thanks!

lockie:
Thank you :slightly_smiling_face:

PR #153, opened 2020-06-16
Add Newline Character Handling to nk_text/nk_label

zgcrowno:
This is a newline handling implementation originally written by @AntiBlueQuirk, but which was never merged into the old Vurtun library for some reason. I've been very pleased with the results so far, but would obviously love feedback on potential issues I may have overlooked when bringing over this implementation.


Hejsil:
I haven't tested the code, but I've read through it. The code looks good to me though :+1:

zgcrowno:
Thanks for taking the time to look this over, @Hejsil!

zgcrowno:
@dumblob, is this functionality something you're still interested in?

krupitskas:
Ehm, it's pretty important feature and this is march 2021, feels like this repo died

dumblob:
I'll chime in quickly just to say I'm sorry for not devoting any time lately to Nuklear (I'm too busy with family life).

I'd like to say though, that I don't think this repo is dead. It's just slow. Any new reviewers are welcome (after a good review or two they'll get full access to this repository and are encouraged to recruit other new reviewers).

nuklear.h
```diff
@@ -5927,7 +5935,7 @@ struct nk_text {
     struct nk_color text;
 };
 NK_LIB void nk_widget_text(struct nk_command_buffer *o, struct nk_rect b, const char *string, int len, const struct nk_text *t, nk_flags a, const struct nk_user_font *f);
-NK_LIB void nk_widget_text_wrap(struct nk_command_buffer *o, struct nk_rect b, const char *string, int len, const struct nk_text *t, const struct nk_user_font *f);
+NK_LIB void nk_widget_text_wrap(struct nk_command_buffer *o, struct nk_rect b, const char *string, int len, const struct nk_text *t, const struct nk_user_font *f, int do_wrap);
```
RobLoach, line 5938: Having the name `nk_widget_text_wrap()`, one would assume it should wrap, regardless of the parameter. In order to avoid the breaking change, could we have an `if` statement prior to the call and call either `nk_widget_text()` or `nk_widget_text_wrap()` based on whether the wrap is required?

RobLoach (changes requested):

dumblob:
Seeing this at this time I'm playing with the idea to declare the current behavior of `nk_text_wrap` and `nk_widget_text_wrap` as faulty and just change their behavior to not only wrap on newlines but also on word boundaries. Instead of providing yet another API (`multiline` functions) which'll presumably subsume probably all existing use cases of the `wrap` functions.

Thoughts?

rswinkle:
Can we resurrect this?  I agree with @dumblob that we should probably just make the existing wrap functions wrap on newlines too.  I could maybe see a function for pre-formatted text that handles newlines but doesn't doesn't wrap otherwise so the user is responsible for making sure the window is big enough or horizontally scrollable but I feel like that's less important than being able to handle newlines at all.

Currently the only way to display a large multiline block of text efficiently (loc-wise) is with a large nk_edit_string and that is less than ideal.

If we can agree on the API I can adapt this code into a new pull request.

StrikerTheHedgefox:
I second the desire to bring this back. I need newline support for a credits page in a source port I'm developing.

PR #156, opened 2020-07-01
Fix layout bounds not accounting for padding

### LINKED ISSUE #132: Padding is not correctly figured for the bottom of a window
Setting padding to, say, {20,20} works fine on the left, top, and right.  But if you then do nk_window_get_content_region(), you'll find that the height is 40 too high (e.g. 2 times the top padding).

I assume this is an error...

Hejsil:
fixes #132

This is not quite done. Currently having this funny bug in the node editor demo.
out

### DIFF
--- a/clib.json
+++ b/clib.json
@@ -1,6 +1,6 @@
 {
   "name": "nuklear",
-  "version": "4.09.1",
+  "version": "4.09.3",
   "repo": "Immediate-Mode-UI/Nuklear",
   "description": "A small ANSI C gui toolkit",
   "keywords": ["gl", "ui", "toolkit"],
--- a/nuklear.h
+++ b/nuklear.h
@@ -19547,6 +19547,8 @@ nk_panel_begin(struct nk_context *ctx, const char *title, enum nk_panel_type pan
     layout->bounds = win->bounds;
     layout->bounds.x += panel_padding.x;
     layout->bounds.w -= 2*panel_padding.x;
+    layout->bounds.y += panel_padding.y;
+    layout->bounds.h -= 2*panel_padding.y;
     if (win->flags & NK_WINDOW_BORDER) {
         layout->border = nk_panel_get_border(style, win->flags, panel_type);
         layout->bounds = nk_shrink_rect(layout->bounds, layout->border);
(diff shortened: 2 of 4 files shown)


ronaaron:
I get what looks correct in my own demo with

````  layout->bounds.h -= panel_padding.y;````

Not adjusting the bounds.y.  For some reason that looks already adjusted.

**Comment by Hejsil** [comment]
Do you have a header on your window? If so, then that would take up some room so it might look like padding is applied. With this change, the repro in #132 gives the same number for both width and height, which is correct for a window with no header. I'll look into this a little more later once I have more time (probably in a weeks time).

**Comment by ronaaron** [comment]
No, the flags are just NO_SCROLLBAR. I set the style.window.padding to [20,20] and the content_bounds with your change looks like it has a double-padding on the top of the window.

I changed the window.header.normal color to something obvious, and it doesn't show up so... no header AFAICT.

**Comment by Hejsil** [comment]
:thinking: Strange. I'd need a test case as I can't reproduce with my simple example:
```
ctx->style.window.padding.x = 50;
ctx->style.window.padding.y = 50;
if (nk_begin(ctx, "Demo", nk_rect(50, 50, 200, 200), NK_WINDOW_NO_SCROLLBAR))
{
    struct nk_rect rect = nk_window_get_content_region(ctx);
    printf("%f %f\n", rect.w, rect.h);
}
nk_end(ctx);
```
This prints `100,000000 100,000000` which is correct for this example.

**Comment by Hejsil** [comment]
Expanding the test case a little to show x and y:
```
ctx->style.window.padding.x = 50;
ctx->style.window.padding.y = 50;
if (nk_begin(ctx, "Demo", nk_rect(0, 0, 200, 200), NK_WINDOW_NO_SCROLLBAR))
{
    struct nk_rect rect = nk_window_get_content_region(ctx);
    printf("%f %f %f %f\n", rect.x, rect.y, rect.w, rect.h);
}
nk_end(ctx);
```

Prints `50,000000 50,000000 100,000000 100,000000` which is correct

ronaaron:
It must be something odd w/ my code.  Not all that interesting, the fix looks proper.

Hejsil:
Alright, I'll just look into the bug in the node editor demo then :+1:

RobLoach (changes requested):
Sounds like some small changes in the demo remain?

RobLoach:
@Hejsil This good to merge now? Thanks a lot!

Hejsil:
@RobLoach Ooh shoot. I did not manage to fix the node_editor demo. I was looking into it earlier after I did a rebase but didn't find the cause

RobLoach:
Revert it until it's fixed?

Hejsil:
This seems like a bug in node_editor and not in nuklear, but not a 100% sure. We could also open an issue for this. I'll have another look at this whenever I have time over the holiday

PR #160, opened 2020-07-06
Feature subpixel api

Nielsbishere:
Multiple PRs:
- https://github.com/Immediate-Mode-UI/Nuklear/pull/50
- https://github.com/Immediate-Mode-UI/Nuklear/pull/35
- https://github.com/Immediate-Mode-UI/Nuklear/pull/15
- https://github.com/Immediate-Mode-UI/Nuklear/pull/14
- https://github.com/Immediate-Mode-UI/Nuklear/pull/13

Various changes such as:
- Fixing clicking through multiple overlapping windows
- Allowing subpixel rendering
- Allowing add_window to return a Window pointer to read out information useful to an abstraction that isn't immediate mode over nuklear.
- Fixed radio buttons style
- Fixed compile issues with C++17 (-Werr -W3)
- Trees can now have 'child' nodes, which are supposed to be the final stop for childless nodes. This isn't fully true, but it can be used to have another tree node type. It doesn't have a symbol as a file should as opposed to a folder.


RobLoach:
Looks like there's a few git conflicts :eyes:

dumblob:
I'm sorry for us not getting earlier to this. I think this PR is quite valuable and was a lot of work to put it all together. We should merge it after a review. Would anyone be willing to update it to fix the merge conflicts? I'm constantly fighting my tighter and tighter schedule...

Nielsbishere:
There we go

dumblob:
I've skimmed it (really no review yet) and it actually looks surprisingly good to me. @RobLoach could you please thoroughly review it? @Hejsil would you also find some time to take a look at this rather sensitive PR? The more eyes the better.

doc/nuklear.html
```diff
@@ -2313,6 +2314,17 @@
...
+                        Fixed radio buttons' style.
+                        Fixed click cascading through multiple buttons; it should only click the upper button.
+                        Strict c++17 now compiles: only declaring memset, memcpy if they are used.
+                        Only asserting index range if ushorts are used for indices.
+                        Made paq.sh identical to paq.bat (outputs to nuklear.h).
+                        Replaced NK_ASSERT(0 && ...) with NK_ERROR(), to ensure it compiles on strict C++17.
+                        Replaced const char** by const char * const * in combobox, to ensure const List&lt;const char*&gt; can be used.
+                        Added nk_tree_type for allowing the symbol next to a tree element to be hidden (for children) if it's not an image.
 - 2020/04/09 (4.02.1) - Removed unused nk_sqrt function to fix compiler warnings
                       - Fixed compiler warnings if you bring your own methods for
```
Hejsil, line 2329: Hmm. The versioning in here seems quite out of date

Readme.md
```diff
@@ -59,37 +59,55 @@ This is very important; not doing it either leads to compiler errors, or even wo
 struct nk_context ctx;
```
Hejsil, line 59: Not sure why we are updating the example in the readme. IMO readme examples should be very small and give the reader a taste of how nuklear is to use. Not sure this update really makes it any better

clib.json
```diff
@@ -1,8 +1,14 @@
...
-  "src": ["nuklear.h"]
+  "keywords": [
+    "gl",
+    "ui",
+    "toolkit"
+  ],
+  "src": [
+    "nuklear.h"
+  ]
 }
```
Hejsil, line 14: Let's not change the formatting of the `clib.json`. Only bump the version

doc/nuklear.html
```diff
@@ -1,4 +1,4 @@
-<meta charset='utf-8' emacsmode='-*- markdown -*-'>
+﻿<meta charset='utf-8' emacsmode='-*- markdown -*-'>
```
Hejsil, line 1: What changed on this line?

src/nuklear_combo.c
```diff
@@ -778,7 +778,7 @@ nk_combo_string(struct nk_context *ctx, const char *items_separated_by_zeros,
     return nk_combo_separator(ctx, items_separated_by_zeros, '\0', selected, count, item_height, size);
 }
 NK_API int
-nk_combo_callback(struct nk_context *ctx, void(*item_getter)(void*, int, const char**),
+nk_combo_callback(struct nk_context *ctx, void(*item_getter)(void*, int, const char* const *),
```
Hejsil, line 781: Pretty sure this is a breaking change as old `void (*)(void *, int, const char**)` functions passed here will now be a compiler error.

src/nuklear_combo.c
```diff
@@ -831,7 +831,7 @@ nk_combobox_separator(struct nk_context *ctx, const char *items_separated_by_sep
 }
 NK_API void
 nk_combobox_callback(struct nk_context *ctx,
-    void(*item_getter)(void* data, int id, const char **out_text),
+    void(*item_getter)(void* data, int id, const char * const *out_text),
```
Hejsil, line 834: Same here

src/nuklear_draw.c
```diff
@@ -556,3 +556,397 @@ nk_draw_text(struct nk_command_buffer *b, struct nk_rect r,
...
+    b->clip.h = r.h;
+    cmd = (struct nk_command_scissor*)
+        nk_command_buffer_push(b, NK_COMMAND_SCISSOR, sizeof(*cmd));
+
+    if (!cmd) return;
+    cmd->x = r.x;
+    cmd->y = r.y;
+    cmd->w = NK_MAX(0, r.w);
+    cmd->h = NK_MAX(0, r.h);
+}
```
Hejsil, line 580: There are a lot of these new subpixel apis that are just copies of the original but without the casts to `short`. Their signature are the same, so should `nk_push_scissor` just support subpixels if enabled instead of having two functions?

src/nuklear_style.c
```diff
@@ -490,8 +490,9 @@ nk_style_from_table(struct nk_context *ctx, const struct nk_color *table)
     tab->background         = nk_style_item_color(table[NK_COLOR_TAB_HEADER]);
     tab->border_color       = table[NK_COLOR_BORDER];
     tab->text               = table[NK_COLOR_TEXT];
-    tab->sym_minimize       = NK_SYMBOL_TRIANGLE_RIGHT;
-    tab->sym_maximize       = NK_SYMBOL_TRIANGLE_DOWN;
+    tab->sym_minimize       = NK_SYMBOL_PLUS;
+    tab->sym_maximize       = NK_SYMBOL_MINUS;
+    tab->sym_child          = NK_SYMBOL_NONE;
```
Hejsil, line 495: I think this is a breaking change visually for people who use the default style

src/nuklear_toggle.c
```diff
@@ -7,8 +7,8 @@
  *
  * ===============================================================*/
 NK_LIB nk_bool
-nk_toggle_behavior(const struct nk_input *in, struct nk_rect select,
-    nk_flags *state, nk_bool active)
+nk_toggle_behavior(struct nk_input *in, struct nk_rect select,
+    nk_flags *state, int active)
```
Hejsil, line 11: Pretty sure this should be a bool and not an int

src/paq.sh
```diff
@@ -1,3 +1,11 @@
...
 
+rm ../nuklear.h
+
+python3 build.py --macro NK --intro HEADER --pub nuklear.h --priv1 nuklear_internal.h,nuklear_math.c,nuklear_util.c,nuklear_color.c,nuklear_utf8.c,nuklear_buffer.c,nuklear_string.c,nuklear_draw.c,nuklear_vertex.c --extern stb_rect_pack.h,stb_truetype.h --priv2 nuklear_font.c,nuklear_input.c,nuklear_style.c,nuklear_context.c,nuklear_pool.c,nuklear_page_element.c,nuklear_table.c,nuklear_panel.c,nuklear_window.c,nuklear_popup.c,nuklear_contextual.c,nuklear_menu.c,nuklear_layout.c,nuklear_tree.c,nuklear_group.c,nuklear_list_view.c,nuklear_widget.c,nuklear_text.c,nuklear_image.c,nuklear_9slice.c,nuklear_button.c,nuklear_toggle.c,nuklear_selectable.c,nuklear_slider.c,nuklear_progress.c,nuklear_scrollbar.c,nuklear_text_editor.c,nuklear_edit.c,nuklear_property.c,nuklear_chart.c,nuklear_color_picker.c,nuklear_combo.c,nuklear_tooltip.c --outro LICENSE,CHANGELOG,CREDITS > ../nuklear.h
+
+if [ -s ../nuklear.h ] ; then
+	echo -- nulear exported
+else
+	echo -- nuklear export failed
+fi
```
Hejsil, line 11: When would this script fail, so that we would need these error messages? Also, I think these messages should be emitted by `build.py` instead of here

Hejsil (changes requested):
I tried, but it is very hard to review this as I have to context switch all the time to figure out what each change relates to. I would personally prefer we don't do these "merge 5 things together" pull requests for a few reasons:

* It's very hard to review
* The git history is a mess, and doing a "Squash and Merge" is not really a good idea as we want multiple commits, one for each fix/feature.
* Change log for this version is a bit of a mess. I would rather we have patch and minor bumps have just one thing they do. Major bumps can be a mess.

Nielsbishere:
@Hejsil I agree, but it was requested I merge this a long time ago iirc.

src/nuklear_combo.c
Nielsbishere: That's right, but that's why there's a major bump and not a minor one. If this change isn't enforced then it won't work for some cases.

clib.json
Nielsbishere: Any reason? To me it seems easier to read this way

Readme.md
Nielsbishere: It shows elements of nuklear that are often going to be used in applications, but I guess it's preference

doc/nuklear.html
Nielsbishere: Unsure, hopefully not line endings

src/nuklear_draw.c
Nielsbishere: Not sure if this was a bad auto merge or for backwards compatibility.

src/paq.sh
Nielsbishere: If python3 is not installed

Readme.md
dumblob: I usually tend to prefer readme examples to be **very** small. In our case we're pretty much covered by everything in `example/` directory. So here I won't advocate for extending the current example.

I'm though fine with extending if it's just about 1-5 SLOC which add and I quote "often used elements". That's not the current case though.

doc/nuklear.html
dumblob: I think we'll fix the versioning after we finally know which of the (many) other PRs get merged first.

clib.json
dumblob: I'm indifferent here. But generally speaking if such changes shall further extend big PRs I'd beg everyone to leave them out for the sake of clarity.

Hejsil: I would personally also like it to be formatted this way, as then it would be inline with my json formatter (4 spaces would be nice too). If enough people agree, then sure we can change it, but I don't want to end up en a situation where we change the clib.json format in every PR, if you know what I mean :)

src/nuklear_combo.c
Hejsil: Bumping to 4.10 is a minor bump. We would need a 5.0 to be a major one.

dumblob: There is actually no major bump (only minor) in this PR as of now :cry:.

But more importantly I'd definitely prefer if major bumping features (i.e. those affecting backwards compatibility) were in a separate PR. It's true that this PR is quite convoluted though I think I'd merge it after it passes our review. But the split due to a major bump is something important for several other reasons (one of which is release engineering).

src/nuklear_draw.c
dumblob: >so should `nk_push_scissor` just support subpixels if enabled instead of having two functions?

Interesting idea. I think it's worth exploring. @Nielsbishere would you find some time to try to rewrite this PR to see if it solves some of the issues we have here and if it makes the PR leaner?

src/nuklear_style.c
dumblob: Generally I'm indifferent on this one. A quick google search seems to suggest that triangles are a tiny bit more modern than `+` `-` though. So I think this can be left out of this big PR.

src/paq.sh
dumblob: Generally I think the only valuable change to `paq.sh` could be to print an error message in case of a failure. But only in that case and it has to be to stderr. I.e.:

```sh
#!/bin/sh

set -e

python build.py --macro NK --intro HEADER --pub nuklear.h --priv1 nuklear_internal.h,nuklear_math.c,nuklear_util.c,nuklear_color.c,nuklear_utf8.c,nuklear_buffer.c,nuklear_string.c,nuklear_draw.c,nuklear_vertex.c --extern stb_rect_pack.h,stb_truetype.h --priv2 nuklear_font.c,nuklear_input.c,nuklear_style.c,nuklear_context.c,nuklear_pool.c,nuklear_page_element.c,nuklear_table.c,nuklear_panel.c,nuklear_window.c,nuklear_popup.c,nuklear_contextual.c,nuklear_menu.c,nuklear_layout.c,nuklear_tree.c,nuklear_group.c,nuklear_list_view.c,nuklear_widget.c,nuklear_text.c,nuklear_image.c,nuklear_9slice.c,nuklear_button.c,nuklear_toggle.c,nuklear_selectable.c,nuklear_slider.c,nuklear_progress.c,nuklear_scrollbar.c,nuklear_text_editor.c,nuklear_edit.c,nuklear_property.c,nuklear_chart.c,nuklear_color_picker.c,nuklear_combo.c,nuklear_tooltip.c --outro LICENSE,CHANGELOG,CREDITS > ../nuklear.h ||
    printf 'ERR nuklear.h wrongly generated (or not generated at all)\n' >&2
```

dumblob:
@Nielsbishere yep, we thought that merging small changes will make it more appropriate. As it turns out, not all the changes are as small as anticipated. Do you think you could help us with splitting the PR (at least the major bump versus other changes) and cleaning it as discussed above?

src/nuklear_combo.c
Nielsbishere: @Hejsil my bad, it's major.minor.bugfix? Then I might have to bump it

aganm:
I double checked this branch to see if all works well in case it ends up being pulled in. There's only one thing that bugs me, I'm not sure if it's on purpose but the color picker widget isn't following my mouse cursor anymore.

You see when I click, I'm holding the mouse, but the color cursor isn't following the mouse cursor:
GIF 12-24-2021 9-35-00 PM

This is the expected behaviour:
GIF 12-24-2021 9-36-15 PM

Nielsbishere:
Oh yeah I will get to this PR in a bit, just busy.
@aganm I'm guessing the event is being consumed before it reaches it for some reason.
@dumblob I think I should probably apply those fixes. I'm not sure how to split it in different branches without rewriting it.

dumblob:
@aganm thanks for the quick test! That's very important to do. Feel free to continue your endeavor if you find some time for it. We'll be thankful for that!

>@dumblob I think I should probably apply those fixes. I'm not sure how to split it in different branches without rewriting it.

Yeah, different fixes will be necessary. Regarding splitting, what do you mean "rewriting it"? Just take few related commits into a separate branch and then consolidate the result with one new commit in that new branch (fixing the discrepancies which are to be expected). Do this until you run out of all the commits from this PR :wink:. This PR will then get closed as superseeded by all the new PRs (for each new branch). Sounds tedious but in this case it's really valuable and we'll be greateful if you managed. Take your time.

Thanks!

mecwerks:
> I double checked this branch to see if all works well in case it ends up being pulled in. There's only one thing that bugs me, I'm not sure if it's on purpose but the color picker widget isn't following my mouse cursor anymore.

Found a fix that worked for me. I'm not sure what else this may affect but it seemed to retain the click through fix

```diff
NK_API nk_bool
nk_input_has_mouse_click_in_rect(const struct nk_input *i, enum nk_buttons id,
    struct nk_rect b)
{
    const struct nk_mouse_button *btn;
-    if (!i || i->mouse.clicked) return nk_false;
+    if (!i) return nk_false;
    btn = &i->mouse.buttons[id];
    if (!NK_INBOX(btn->clicked_pos.x,btn->clicked_pos.y,b.x,b.y,b.w,b.h))
        return nk_false;
    return nk_true;
}
```

aganm:
@mecwerks Thanks for that, I have looked into it and this change comes from commit 

> 8e89387e2d946545a19bbe61bd1bafaa6da90c7d  Merge branch 'fix_click' of https://github.com/Nielsbishere/Nuklear into fix_radio_buttons

 It seems like the reason for this change was to fix click related to radio buttons, but, I have tested your proposed fix and it seems to work, both color pickers and radio buttons seem to work perfectly with this fix!

Fix which is a revert of this one specific line https://github.com/Immediate-Mode-UI/Nuklear/blob/8e89387e2d946545a19bbe61bd1bafaa6da90c7d/src/nuklear_input.c#L157
back to what it was before this PR
https://github.com/Immediate-Mode-UI/Nuklear/blob/3a0aafb9cc117b66d6f1cfcc36bb6d0c8dc2ea08/src/nuklear_input.c#L151

With this fix I would be happy if we could merge this PR.

Nielsbishere:
Oh hey it's my old branch. @aganm @mecwerks the reason for this change was quite important; if you have a dropdown window over a radio button, it'll push through the window into the radio button. The proper fix would be to figure out why the color picker has this behavior. Maybe this check is in the wrong place; preventing that behavior? But letting the radio button be clicked while the input event is already consumed is not good imo

aganm:
@Nielsbishere Si

> Oh hey it's my old branch. @aganm @mecwerks the reason for this change was quite important; if you have a dropdown window over a radio button, it'll push through the window into the radio button. The proper fix would be to figure out why the color picker has this behavior. Maybe this check is in the wrong place; preventing that behavior? But letting the radio button be clicked while the input event is already consumed is not good imo

The reason the color picker does this is because the color picker calls `nk_button_behavior`, which in turn calls `nk_input_has_mouse_click_in_rect`. Both the radio button and color picker rely on the same underlying functions.

Maybe the function should be copied with a different name so that the color picker can have its own behavior instead of reusing `nk_button_behavior`? Maybe `nk_pick_behavior` which would be a copy of `nk_button_behavior` except for this little difference?

aganm:
> Oh hey it's my old branch. @aganm @mecwerks the reason for this change was quite important; if you have a dropdown window over a radio button, it'll push through the window into the radio button. The proper fix would be to figure out why the color picker has this behavior. Maybe this check is in the wrong place; preventing that behavior? But letting the radio button be clicked while the input event is already consumed is not good imo

It seems both color picker and radio button rely on the same underlying `nk_button_behavior` function. But the difference is that the radio button calls `nk_button_behavior` with a behavior parameter of `NK_BUTTON_DEFAULT` and the color picker  calls `nk_button_behavior` with a behavior parameter of `NK_BUTTON_REPEATED`. The function `nk_button_behavior` seems to not handle the repeated mode correctly.

Nielsbishere:
@aganm sounds like a solution is to only check if it's not repeated then. I'm not continuing this branch as my current project doesn't use it. Maybe if I ever use it again, but I don't have any time left to do this rn

RobLoach:
3 years :wink:  Thanks for your push on this. Are there any individual PRs you believe we should bring in? A large change like this can be difficult to review, as demonstrated. Smaller code changes are easier to bring in.

Nielsbishere:
yeah I was going through my open PRs and noticed this was still open. I'm not sure what happened since then in terms of development, but the original was smaller PRs and was asked to be merged into one bigger PR.
- https://github.com/Immediate-Mode-UI/Nuklear/pull/50
- https://github.com/Immediate-Mode-UI/Nuklear/pull/35
- https://github.com/Immediate-Mode-UI/Nuklear/pull/15
- https://github.com/Immediate-Mode-UI/Nuklear/pull/14
- https://github.com/Immediate-Mode-UI/Nuklear/pull/13
So it might be worth looking what's worth salvaging from there if these features haven't made it in in these 3 years.

PR #176, opened 2020-08-11
Fixed offset in the documentation

crazyBaboon:
The function prototypes now appear above the respective text description.


aganm:
Why above? It's kind of a convention that documentation comes before the thing that is being documented.

RobLoach (changes requested):
Yeah, standard code documentation usually appears above the declaration. I'd give a :-1: on this one.

dumblob:
I appreciate the effort you put into this @crazyBaboon . So far I think there are not enough of convincing arguments to merge this. Feel free to describe what this change would achieve and which existing problem it would solve.

I'll close this one for now. Thanks anyway!

PR #185, opened 2020-09-04
Change boolean int to nk_bool

aganm:
Nuklear uses `int` as boolean type which creates ambiguity between boolean values and actual integers.

This adds nk_bool and replaces all occurrences of `int` booleans by nk_bool.
nk_bool typedefs to `int` by default so it's a drop in replacement change. For users, it's as if nothing ever changed.

I also added NK_INCLUDE_STANDARD_BOOL. If defined, `stdbool.h` is included and nk_bool typedefs to `bool`.

### DIFF
--- a/src/nuklear.h
+++ b/src/nuklear.h
@@ -177,6 +177,15 @@ extern "C" {
   #endif
 #endif
 
+#ifndef NK_BOOL
+  #ifdef NK_INCLUDE_STANDARD_BOOL
+    #include <stdbool.h>
+    #define NK_BOOL bool
+  #else
+    #define NK_BOOL int /* could be char, use int for drop-in replacement backwards compatibility */
+  #endif
+#endif
+
 typedef NK_INT8 nk_char;
 typedef NK_UINT8 nk_uchar;
 typedef NK_UINT8 nk_byte;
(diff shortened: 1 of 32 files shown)


src/nuklear.h
```diff
@@ -203,6 +213,11 @@ NK_STATIC_ASSERT(sizeof(nk_flags) >= 4);
 NK_STATIC_ASSERT(sizeof(nk_rune) >= 4);
 NK_STATIC_ASSERT(sizeof(nk_size) >= sizeof(void*));
 NK_STATIC_ASSERT(sizeof(nk_ptr) >= sizeof(void*));
+#ifdef NK_INCLUDE_STANDARD_BOOL
+NK_STATIC_ASSERT(sizeof(nk_bool) == sizeof(bool));
+#else
+NK_STATIC_ASSERT(sizeof(nk_bool) == 4);
```
Hejsil, line 219: `nk_bool` is defined as `int` here and `int` is defined in the C standard as having a minimum size of 16 bits, so this assert is incorrect. Hmm, either this assert should be `sizeof(nk_bool) >= 2` or `nk_bool` should be defined as `nk_int` (which would technically be a breaking change)

Hejsil (commented):
Haven't looked through every replacement of `int -> nk_bool` but this PR seems good to me at a glance. Only one thing:

dumblob:
Good catch @Hejsil! I think I'd prefer the assert solution which seems to be already implemented. So I'm merging this.

PR #187, opened 2020-09-06
Fix warning: 'fictional_handle' defined but not used

aganm:


aganm:
Doesn't compile in C89.

PR #199, opened 2020-10-07
Fix nk_combo return type wrongly changed to nk_bool

aganm:


aganm:
Also `nk_combo_begin` changed to nk_bool, this one should be nk_bool.

dumblob:
@aganm please read our Reviewers' Guidelines and fix the missing bits before we'll dive into review :wink:.

@exchg could you please review so that we have at least two reviewers?

exchg:
> @exchg could you please review so that we have at least two reviewers?

sure

exchg:
looks like it's ok

dumblob:
LGTM. Thanks for catching this inconsistency @exchg and for the second review!

PR #200, opened 2020-10-07
Add a rawfb backend which uses SDL_Surface's for output.

wziard:
The other sdl  demo backends all use OpenGL, which defeats the
purpose of using SDL_Renderer for portabilty.
This is maybe not as good as a full SDL_Renderer backend, but it
is more flexible. For most applications software rendering of the
UI is more than fast enough.

### DIFF
--- a/demo/sdl2surface_rawfb/Makefile
+++ b/demo/sdl2surface_rawfb/Makefile
@@ -1,0 +1,9 @@
+CFLAGS=`sdl2-config --cflags --libs`   -std=c11 -Wall  -O0 -g -fvisibility=hidden -Wno-unused `pkg-config SDL2_ttf --cflags --libs`
+
+.PHONY: clean
+
+demo: main.c sdl2surface_rawfb.h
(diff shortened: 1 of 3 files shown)


dumblob:
Thanks!

Just out of curiosity - how fast is it on your platform (fps, your CPU type, CPU use during rendering)?

wziard:
> Thanks!
> 
> Just out of curiosity - how fast is it on your platform (fps, your CPU type, CPU use during rendering)?

I haven't really checked. I don't have an fps counter in my simple game. It feels smooth enough. 

In my game I render the ui to a SDL_Surface and then overlay that transparently over my 2d game.
It's not the most efficient, streaming a 1920x1080 texture every frame, but it feels smooth. And as I only use it to display menus etc while the game is paused a bit of a drop in fps doesn't really bother me.

It's not efficient, but, hey, it's very easy! :+1: 

I love the simplicity of Nuklear. I used it for a glfw project before. But this is a simple 2d SDL game which I'd like to keep as platform agnostic as possible, hence this rawfb approach.

Issue #231, opened 2020-12-19
ISO ansi pedantic compilation error (gcc)


riskpeep:
```
(base) rskerr@penguin:~/dev/Nuklear/demo/sdl_opengl3$ make
rm -f bin/demo 
cc main.c -std=c89 -pedantic -O2 -o bin/demo -lSDL2 -lGL -lm -lGLU -lGLEW
In file included from main.c:25:
../../nuklear.h:6056:8: warning: extra tokens at end of #endif directive [-Wendif-labels]
 #endif // STBTT_malloc
        ^
../../nuklear.h:6058:8: warning: extra tokens at end of #endif directive [-Wendif-labels]
 #endif // NK_INCLUDE_FONT_BAKING
        ^
In file included from main.c:25:
../../nuklear.h: In function ‘nk_font_bake_pack’:
../../nuklear.h:16436:9: warning: ISO C90 forbids mixed declarations and code [-Wdeclaration-after-statement]
         struct stbtt_fontinfo *font_info = &baker->build[i++].info;
         ^~~~~~
```
```
(base) rskerr@penguin:~/dev/Nuklear/demo/sdl_opengl3$ gcc --version
gcc (Debian 8.3.0-6) 8.3.0
Copyright (C) 2018 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
```

riskpeep:
PR Accepted, closing the issue.

PR #232, opened 2020-12-20
Fix C89 compilation errors

riskpeep:
This PR seeks to resolve issue #231 ISO ansi pedantic compilation error (gcc).

It makes changes to src/nuklear_internal.h and to src/nuklear_font.h to allow them to compile w/ gcc flags '-std=c89 -pedantic'


Hejsil:
We bump up the version number on every change, so if you could do that in [package.json](https://github.com/Immediate-Mode-UI/Nuklear/blob/master/package.json) and add a small changelog to [CHANGELOG](https://github.com/Immediate-Mode-UI/Nuklear/blob/master/src/CHANGELOG), then this can be merged :+1:

riskpeep:
Glad to do.  Added an update to package.json and the CHANGELOG

Issue #236, opened 2020-12-28
NK_UINT_DRAW_INDEX not working for D3D11 implementation


Hecklezz:
If I draw a tonne of things at once, I reach the limit of the maximum number of indices or vertexes of the default ushort size. However, I read that defining NK_UINT_DRAW_INDEX will increase this limit to a uint which should be way more than enough for what I need it for. 

However, if I have this define added, even when barely drawing anything for e.g. just using the nuklear demo, this happens (defined)- 
<details>
  <summary>Click to expand</summary>

![image](https://user-images.githubusercontent.com/1632632/103206973-e269f580-4948-11eb-88e1-db3ccad334f0.png)

</details>

when it should look normal like this (not defined)- 
<details>
  <summary>Click to expand</summary>
  
![image](https://user-images.githubusercontent.com/1632632/103206942-d2eaac80-4948-11eb-90af-7beec5baa08c.png)

</details>

Does anyone know how to properly fix this in the D3D11 implementation, or perhaps what I am doing wrong? 

The demo used is from here- 
https://github.com/Immediate-Mode-UI/Nuklear/tree/master/demo/d3d11

Thank you.

Hejsil:
Hi. I seems you're hitting the problem stated in [this comment](https://github.com/Immediate-Mode-UI/Nuklear/blob/5570172553ce48af8fe0513c8091f9bb04d3e0e4/src/nuklear_vertex.c#L227-L237). Specifically:
```
     * and don't forget to specify the new element size in your drawing
     * backend (OpenGL, DirectX, ...). For example in OpenGL for `glDrawElements`
     * instead of specifing `GL_UNSIGNED_SHORT` you have to define `GL_UNSIGNED_INT`.
     * Sorry for the inconvenience. */
```

Hejsil:
Now, I cannot be 100% sure, but I think you need to fix [this line](https://github.com/Immediate-Mode-UI/Nuklear/blob/daecaaa5af124d4c7365883e12c2ad5b18760d67/demo/d3d11/nuklear_d3d11.h#L91):
```diff
-    ID3D11DeviceContext_IASetIndexBuffer(context, d3d11.index_buffer, DXGI_FORMAT_R16_UINT, 0);
+    ID3D11DeviceContext_IASetIndexBuffer(context, d3d11.index_buffer, DXGI_FORMAT_R32_UINT, 0);
```

Hecklezz:
[...]

PR #237, opened 2020-12-29
Made D3D11 implementation work when using NK_UINT_DRAW_INDEX

Hecklezz:
Issue #236 

When defining NK_UINT_DRAW_INDEX to increase the size of nk_draw_index from ushort to a uint to allow drawing a lot more on a window, the D3D11 implementation doesn't handle the change properly and exhibits the behavior explained in the Issue. 

This Pull Request implements a change that allows the D3D11 implementation to work regardless if you have the NK_UINT_DRAW_INDEX defined or not.


Hejsil:
:confetti_ball:

Issue #222, opened 2020-12-02
checkbox behaviour is reversed and wrong


crazyBaboon:
When the checkbox is active, its square should be coloured and when is inactive it should be uncoloured.

Currently, this behaviour is reversed and wrong.

You can see that in overview -> Widgets -> Inactive.

ridams:
I confirm that but with another style its OK

crazyBaboon:
@ridams any other style?

ridams:
use the included style.c file then use the dark style; just uncomment

crazyBaboon:
Unfortunately, this issue is still unresolved.

aganm:
I think the confusion is coming from the color scheme. You can change the color to red for example.

```
ctx->style.checkbox.cursor_hover = nk_style_item_color(nk_rgb(255,0,0));
ctx->style.checkbox.cursor_normal= nk_style_item_color(nk_rgb(255,0,0));
```

Off = Blank square
![image](https://user-images.githubusercontent.com/20380758/111610001-431f0680-87b1-11eb-98a3-d0d589ecfc8f.png)
On = Square with a square in it
![image](https://user-images.githubusercontent.com/20380758/111610023-487c5100-87b1-11eb-8b76-dc5eeb758cec.png)

crazyBaboon:
@aganm good point.

I have added three new colors to the checkbox, to make it independent from the radio button colors.

I came up with another [solution](https://github.com/crazyBaboon/Nuklear/tree/fix_checkbox_appearance):

Inactive:

![Screenshot from 2021-03-18 19-53-37](https://user-images.githubusercontent.com/19549703/111696882-bf0a6480-882c-11eb-8650-34adc3c67428.png)

Active:

![Screenshot from 2021-03-18 20-55-11](https://user-images.githubusercontent.com/19549703/111697047-f11bc680-882c-11eb-9126-f4e0009741fa.png)

Hovered:

![Screenshot from 2021-03-18 19-53-33](https://user-images.githubusercontent.com/19549703/111697071-f9740180-882c-11eb-9ff1-c3f13a0e0df0.png)

ConnorBP:
why on earth is this issue still not resolved lmao had me questioning my sanity for a while there thinking i somehow implemented it wrong

RobertLemmens:
I ported the PR from @zhouxs1023 to here in #468 , it looks good

TomJGooding:
[...]

PR #238, opened 2021-01-03
Fix the behaviour of the checkbox

crazyBaboon:
This fixes issue #222 - the default style checkbox is reversed ('on' if 'off')


crazyBaboon:
This solution is not perfect because of side effects: the radio buttons become light grey when toggled, which IMO is aesthetically unpleasant. So i'm closing this.

PR #254, opened 2021-01-22
Fix behaviour of checkbox (issue #222)

crazyBaboon:
this solution fixes the issue #222 of the checkbox for the default style.

### DIFF
--- a/nuklear.h
+++ b/nuklear.h
@@ -24172,7 +24172,7 @@ nk_do_toggle(nk_flags *state,
     if (style->draw_begin)
         style->draw_begin(out, style->userdata);
     if (type == NK_TOGGLE_CHECK) {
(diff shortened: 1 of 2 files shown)


aganm:
Behaviour functions correctly.

image

It's not the checkbox that is inactive, it's the button.

The inactive checkbox is active, so it is colored.

aganm:
Here, I printed the value of inactive to show that it indeed works correctly.

GIF 3-18-2021 2-32-10 AM

crazyBaboon:
@aganm I have updated the issue description with pics, to avoid confusion.

crazyBaboon:
As can be seen in issue #222, this is not a perfect solution, and also it breaks the use of styles.c

PR #275, opened 2021-03-15
Update documentation for nk_bool return functions and change nk_propertyi back to int

aganm:


src/CHANGELOG
```diff
@@ -8,6 +8,8 @@
 ///    - [yy]: Minor version with non-breaking API and library changes
 ///    - [zz]: Bug fix version with no direct changes to API
 ///
+/// - 2020/12/19 (4.06.4) - Change nk_propertyi back to int
+/// - 2020/12/19 (4.06.3) - Update documentation for functions that now return nk_bool
```
Hejsil, line 12: It seems the `CHANGELOG` and `nuklear.h` are out of sync. Plz regenerate the `nuklear.h` :)

Also, `package.json` needs to have the version updated too.

Hejsil (commented):
One small thing

src/CHANGELOG
aganm: Thanks!

Hejsil:
:+1:

PR #283, opened 2021-04-08
Fixed copy and paste in glfw_opengl3 demo

### LINKED ISSUE #179: Clipboard issue with glfw implementation
Hi. Recently, i create a little program using glfw, opengl3 and nuklear. I've had problems with the clipboard feature. When a tried to copy or paste text into some edit_string widget, i get a sigfault. This is because, in the `nk_glfw3_clipboard_paste` function, the `nk_handled` that it gets is **NULL**. I saw in the `nk_glfw3_init` function and found that the clip user data is set to NULL:

image

So, this can be fixed setting the user data to the glfw handler:
image

I don't know if this is an error or if it's set to NULL for some reason. I searched in the issues and i didn't found anyone that comment this problems.

marcinzelent:
Fixes #179.


Hejsil:
Thanks!

Issue #136, opened 2020-05-18
Version tracking?


oxr463:
There are no tags or releases. How can I install a specific version of this library?

dumblob:
Unfortunately you'll need to check out each historical commit and look for version number either in `src/CHANGELOG` or `package.json`.

PRs are highly welcome (there was an ongoing effort to automate our release management, but I can't find it right now in all the PRs and Issues).

oxr463:
Interesting. So typically, even with a change log or a `package.json` file, we would still use git tags or even GitHub Releases so that we can easily download specific versions.

Let me fork and show you what I mean.

oxr463:
Okay, here is an example tag, https://github.com/oxr463/Nuklear/releases/tag/v4.02.4

If you wish to apply it to the main branch, you can follow these commands:

```sh
git remote add oxr463-fork https://github.com/oxr463/Nuklear.git
git pull --all # pulls in the tag(s)
git tag -l # list the tag(s)
git push --tags # seems self explanatory
```
Each release needs a tag. Not sure if someone wants to go back and tag each one, or if we just want to start tagging on each release moving forward.

# Reference:

- https://git-scm.com/book/en/v2/Git-Basics-Tagging
- https://help.github.com/en/github/administering-a-repository/about-releases

dumblob:
I see at least 2 things here:
1. we need to decide when to tag (shall we do it for each `.patch` bump? or just for `.minor`? or even somehow for backends & demos (e.g. through some additional hash to the existing semver - I wouldn't want to start versioning our backends&demos as we do not require them to have such quality as `nuklear.h` library itself)?)
2. this has to be automated (e.g. part of our `paq` scripts etc. - including "safety" measures to allow for removal of temporary tags if any etc.) or it has to be well-described in `Readme.md` under [Reviewers guide](https://github.com/Immediate-Mode-UI/Nuklear#reviewers-guide )

Btw. I prefer the automated method e.g. with `paq`, but feel free to make PR and we'll discuss the details there.

dumblob:
@Hejsil thoughts?

oxr463:
> 1. we need to decide when to tag

You should tag each time the version changes in your `package.json` file.

> 2. ...or it has to be well-described in Readme.md under Reviewers guide.

It is simple enough to add a step here that says how to do it.

Here are two simple ways to tag:

1. Using git:

```sh
git tag v4.02.5
git push --tags
```

[...]

PR #287, opened 2021-04-22
create-tag github action for future tags

epsilon-0:
Closes: https://github.com/Immediate-Mode-UI/Nuklear/issues/136

Signed-off-by: Aisha Tammy <floss@bsd.ac>


epsilon-0:
This is very useful for package maintainers to keep versions synced across different operating systems.

RobLoach:
Seems like a good idea. Are we able to change to the new clib.json file?

dumblob:
Yep, tagging shall follow `clib.json` (which is assumed to be in sync with the version in sources).

RobLoach:
`        root: clib.json` should do it. Added.

PR #288, opened 2021-04-29
Fix canvas example not rendering fully on macOS

plat-phoenix:
On macOS (or just retina/hi-dpi displays), the window size is not necessarily the render size so, glfwGetFramebufferSize must be used instead of glfwGetWindowSize in the render loop.

So originally it would look like this:
Screen Shot 2021-04-29 at 9 01 08 AM

But with my change it now looks like this:
Screen Shot 2021-04-29 at 9 04 39 AM

### DIFF
--- a/example/canvas.c
+++ b/example/canvas.c
@@ -477,7 +477,7 @@ int main(int argc, char *argv[])
         canvas_end(&ctx, &canvas);}
 
         /* Draw */
-        glfwGetWindowSize(win, &width, &height);
+        glfwGetFramebufferSize(win, &width, &height);
         glViewport(0, 0, width, height);
         glClear(GL_COLOR_BUFFER_BIT);
         glClearColor(0.2f, 0.2f, 0.2f, 1.0f);
--- a/example/canvas.c
+++ b/example/canvas.c
@@ -477,6 +477,9 @@ int main(int argc, char *argv[])
         canvas_end(&ctx, &canvas);}
 
         /* Draw */
+        /* Framebuffer size is used instead of window size because the window size is in screen coordinates instead of pixels.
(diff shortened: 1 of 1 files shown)


dumblob:
Please add comment to that line why the `...WindowSize...` function wasn't used, so that people know it.

I'd be also interested how this behaves on other systems (Windows, Linux X11) - do you have any opportunity to test it or somehow prove (e.g. by pointing to some doc) it resizes with the window when running and that it always reports a correct "window" size (even with different multimonitor setups etc.)?

Hejsil:
> I'd be also interested how this behaves on other systems (Windows, Linux X11) 

I've tested this on Linux X11 by resizing and dragging the window around on multiple monitors and I get the same values for width and height as on master.

Platinum-Phoenix:
In [glfw's docs on window size](https://www.glfw.org/docs/latest/window_guide.html#window_size) it actually says:
> Note
Do not pass the window size to glViewport or other pixel-based OpenGL calls. The window size is in screen coordinates, not pixels. Use the framebuffer size, which is in pixels, for pixel-based calls.

And here are the docs on both functions:
 [`glfwGetWindowSize`](https://www.glfw.org/docs/latest/group__window.html#gaeea7cbc03373a41fb51cfbf9f2a5d4c6):
> This function retrieves the size, in **screen coordinate**s, of the content area of the specified window. If you wish to retrieve the size of the framebuffer of the window in **pixels**, see glfwGetFramebufferSize.

[`glfwGetFramebufferSize`](https://www.glfw.org/docs/latest/group__window.html#ga0e2637a4161afb283f5300c7f94785c9):
> This function retrieves the size, in pixels, of the framebuffer of the specified window. If you wish to retrieve the size of the window in screen coordinates, see glfwGetWindowSize.

PR #292, opened 2021-05-12
Xcb cairo

riri:
As announced on the gallery thread, I adapted a backend xcb+cairo to be included in nuklear demos, and fixed the rect multi-color (thanks to [twidgets](https://github.com/taiwins/twidgets) project).

Two more little additions (I forgot to remove them before doing the commits, so if they are ok, so far so good ^^):
* added `.vscode` to .gitignore (that way auto-generated IDE directory does not mess up the git tree)
* added a style selector through a combo to the demo (with INCLUDE_STYLE defined)

If you don't want them, I will make a new commit to revert.

**Note**: this configuration is pretty interesting for little desktop apps, by reducing both CPU usage (blocking on events) and memory (at least on Linux, where cairo is mostly loaded in shared library as part of gtk loaded elsewhere). I come with a demo using only about 8MB of RSS)


.gitignore
```diff
@@ -8,3 +8,4 @@ docs/src
 *.tmp
 *.swo
 *.swp
+.vscode
```
Hejsil, line 11: Text editor specific ignores should be in your own global `.gitignore` :)

demo/xcb_cairo/main.c
```diff
@@ -0,0 +1,159 @@
...
+#include <assert.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <stdarg.h>
+#include <string.h>
+/*#include <limits.h>*/
+/*#include <math.h>*/
+/*#include <sys/time.h>*/
+/*#include <unistd.h>*/
+/*#include <time.h>*/
```
Hejsil, line 11: Reason why these a just commented out? Do the need to be removed or+

demo/xcb_cairo/nuklear_xcb.h
```diff
@@ -0,0 +1,852 @@
...
+ *
+ ****************************************************************************/
+
+#ifndef NK_XCB_CAIRO_H
+#define NK_XCB_CAIRO_H
+
+/* just to make my IDE happy */
+#ifndef NK_NUKLEAR_H_
+#include "../../nuklear.h"
+#endif
```
Hejsil, line 41: Hmm, I would pref we not have IDE specific "hacks", but I'm note quite sure what the "ideomatic" way of depending on the nuklear header here is. Maybe:
```c
#undef NK_IMPLEMENTATION
#include "../../nuklear.h"
```

Still feels a little gross

demo/xcb_cairo/nuklear_xcb.h
```diff
@@ -0,0 +1,852 @@
...
+
+/* Xcb part: work on windows */
+NK_API struct nk_xcb_context *nk_xcb_init(const char *title, int pos_x, int pos_y, int width, int height);
+NK_API void nk_xcb_free(struct nk_xcb_context *xcb_ctx);
+
+NK_API int nk_xcb_handle_event(struct nk_xcb_context *xcb_ctx, struct nk_context *nk_ctx);
+NK_API void nk_xcb_render(struct nk_xcb_context *xcb_ctx);
+NK_API void nk_xcb_size(struct nk_xcb_context *xcb_ctx, int *width, int *height);
+
+/* TODO: copy/paste */
```
Hejsil, line 63: Copy paste what? :)

demo/xcb_cairo/nuklear_xcb.h
```diff
@@ -0,0 +1,852 @@
...
+
+#endif /* NK_XCB_CAIRO_H */
+
+/*****************************************************************************
+ *
+ *                           IMPLEMENTATION
+ *
+ ****************************************************************************/
+
+/*#ifdef NK_XCB_CAIRO_IMPLEMENTATION*/
```
Hejsil, line 86: Seems like this needs to be uncommented

demo/xcb_cairo/nuklear_xcb.h
```diff
@@ -0,0 +1,852 @@
...
+        | XCB_EVENT_MASK_EXPOSURE
+        /*| XCB_EVENT_MASK_VISIBILITY_CHANGE*/
+        | XCB_EVENT_MASK_STRUCTURE_NOTIFY
+        /*| XCB_EVENT_MASK_RESIZE_REDIRECT*/
+        /*| XCB_EVENT_MASK_SUBSTRUCTURE_NOTIFY*/
+        /*| XCB_EVENT_MASK_SUBSTRUCTURE_REDIRECT*/
+        /*| XCB_EVENT_MASK_FOCUS_CHANGE*/
+        /*| XCB_EVENT_MASK_PROPERTY_CHANGE*/
+        /*| XCB_EVENT_MASK_COLOR_MAP_CHANGE*/
+        /*| XCB_EVENT_MASK_OWNER_GRAB_BUTTON*/
```
Hejsil, line 186: I'm guessing the commented out events are to show what is available, but the demo doesn't need it?

Hejsil (commented):
Not familiar with the cairo apis, so I've just skimmed over, looking for some some general improvements that could be made. I might try to compile and run this later if I remember.

.gitignore
riri: lol I never thought of this, you just made my day :)

demo/xcb_cairo/main.c
riri: Actually I don't remember, I usually try to reduce the amount of includes as most as possible. It works without them but was included in x11 demo... I will remove the lines

demo/xcb_cairo/nuklear_xcb.h
riri: clipboard :)

riri:
To be honest, I am not familiar either with the cairo API, learning as I adapted an external implementation to nuklear and from various other projects.

I think the font stuff needs improvement, but those are mostly demo implementation to serve as a base (what I did for my [pugl/cairo](https://gitlab.com/riribreizh/nk_pugl_cairo) backend).

RobLoach:
Would it be okay if we kept this in a repo somewhere? Sounds like not many people are familiar with the APIs behind these. They seem pretty awesome though! Having it in your own repo will give you a lot more control over changes too, which is nice.

dumblob:
Our policy is to accept backends despite them being incomplete or buggy or whatnot. Just document the downsides and we'll merge it :wink:. We appreciate the effort. Thanks a lot!

PR #294, opened 2021-05-13
Fix uninited variable.

wziard:


Hejsil:
Before this can be merged, the nuklear version needs to be updated. Any changes made to code in `src` needs to be versioned in the [package.json](https://github.com/Immediate-Mode-UI/Nuklear/blob/master/package.json) and [CHANGELOG](https://github.com/Immediate-Mode-UI/Nuklear/blob/master/src/CHANGELOG).

Also, just wondering. Was this something found by a compiler warning? Which compiler triggered this?

BieHDC:
This has been fixed in https://github.com/Immediate-Mode-UI/Nuklear/commit/4be905921b10ca76a2ac24d22ba6766a761083c0

PR #298, opened 2021-05-16
Change NK_MEMSET to memset() in demos, to prevent compilation errors when Nuklear.h is split or NK_IMPLEMENTATION is undefined

FrostKiwi:
All the backend implementations demos make use of the NK_MEMSET macro.
However NK_MEMSET is not defined in Nuklear's public header, but in it's private implemenation. Specifically in nuklear_internal.h As the name suggests, NK_MEMSET is meant for nuklear's internal implementation.
Even the single header build file makes this obvious: nuklear.h is --pub, but nuklear_internal.h is --priv1 .

- But why change it?
Although Nuklear is a single header library, it sometimes makes sense to split it for faster compilation. The Implementation can be precompiled into an object file and the header can be precompiled into a precompiled header object file. This allows blazing fast recompilation and instant response from linting. However, because the backend implementations mix public and internal definitions, it breaks compilation after the split occurs.
- Why not make use of internal definitions for consistency with Nuklear?
Because it is not consitent in the first place. The backend implementations use malloc(), which is not internal, but NK_MEMSET, which is.

Also it straight up contradicts:

> The implementation mode requires to define  the preprocessor macro NK_IMPLEMENTATION in * one * .c/.cpp file before #includeing this file

### DIFF
--- a/demo/glfw_opengl2/nuklear_glfw_gl2.h
+++ b/demo/glfw_opengl2/nuklear_glfw_gl2.h
@@ -136,7 +136,7 @@ nk_glfw3_render(enum nk_anti_aliasing AA)
             {NK_VERTEX_COLOR, NK_FORMAT_R8G8B8A8, NK_OFFSETOF(struct nk_glfw_vertex, col)},
             {NK_VERTEX_LAYOUT_END}
         };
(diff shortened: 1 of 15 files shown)


Hejsil:
Thanks 👍

PR #304, opened 2021-05-31
Add support for 9-slice/9-patch -> merge with latest upstream version + remove build.py related code

dav64:
It seems that the PR #69 is abandoned, but the 9patch feature would be nice to have in Nuklear and as @Hejsil suggests, it look like very close to validation 🙂
So I've done the merge to latest master commit (at the time of writing) + removed changes related to build.py

No changes have been made on @Michael-Kelley's code 🙂

I think that Nuklear could be used also as an end-user GUI and not only as developer's one and I think this PR is a nice step toward it
(And I'm more convinced when I saw the last demo screenshot with the army-fps-ish options menu)


dumblob:
@FrostKiwi could you maybe find some time for this? There seems to be quite some demand to prioritize this one and I'm still out of time for this project :cry:.

FrostKiwi:
> @FrostKiwi could you maybe find some time for this? There seems to be quite some demand to prioritize this one and I'm still out of time for this project .

Will dive into it :]

src/nuklear_button.c
```diff
@@ -98,11 +98,17 @@ nk_draw_button(struct nk_command_buffer *out,
...
+            nk_draw_image(out, *bounds, &background->data.image, nk_white);
+            break;
+        case NK_STYLE_ITEM_9SLICE:
+            nk_draw_9slice(out, *bounds, &background->data.slice, nk_white);
+            break;
+        case NK_STYLE_ITEM_COLOR:
+            nk_fill_rect(out, *bounds, style->rounding, background->data.color);
+            nk_stroke_rect(out, *bounds, style->rounding, style->border, style->border_color);
+            break;
     }
```
Hejsil, line 112: Seems we have this switch in quite a few places. I wonder if an `nk_draw_background` function would be something nice to have 🤔

Hejsil (approved):
Just took a look over it again. Haven't tested it, but the code looks good 👍 I've left some nitpick comments :)

FrostKiwi:
Learning to properly use skinning in this new context took me *forever* and properly exporting skinning designs from Photoshop has also pitfalls.
Next PR after this should be an updated documentation :S
Have an exam coming up this week, will give my review with some designs to test on next week.

lockie:
Is there a chance this would be merged anytime soon?

dumblob:
@lockie we strongly prefer [review of more than one person](https://github.com/Immediate-Mode-UI/Nuklear#reviewers-guide ) but here only @Hejsil found his precious time to review it so far (I'm too busy for this :cry:). If you could review it, we'll **immediately** merge this.

And if you'll review few other PRs, I'll give you full rights to this repo afterwards as we need more of interested people.

lockie:
@dumblob thank you for your trust! I'll have a deeper look into the code of this PR a bit later.

src/nuklear_9slice.c
```diff
@@ -0,0 +1,106 @@
...
+ *
+ *                          9-SLICE
+ *
+ * ===============================================================*/
+NK_API struct nk_9slice
+nk_sub9slice_ptr(void *ptr, nk_ushort w, nk_ushort h, struct nk_rect rgn, nk_ushort l, nk_ushort t, nk_ushort r, nk_ushort b)
+{
+    struct nk_9slice s;
+    nk_zero(&s, sizeof(s));
+    struct nk_image *i = (struct nk_image*)&s;
```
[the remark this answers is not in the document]
2asoft: Can we get this changed, @dav64 ? It makes the code less clear, as it is, IMO.
Other that this, the rest of the change looks good.
It would be good to have a usage example in the demos -- @FrostKiwi, are you still up for writing a new skinning example?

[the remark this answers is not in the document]
dav64: After code investigating, I didn't found if there were specific reason for the original author (Michael) to not using something like @Hejsil solution as it also look cleaner to me
Maybe that only was Nuklear casting fashion at the time of writting :thinking: 
I'll update that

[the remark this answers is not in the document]
dav64: Updated, let me know if you need anything else :slightly_smiling_face:

nuklear.h
```diff
@@ -480,7 +480,8 @@ struct nk_rect {float x,y,w,h;};
 struct nk_recti {short x,y,w,h;};
 typedef char nk_glyph[NK_UTF_SIZE];
 typedef union {void *ptr; int id;} nk_handle;
-struct nk_image {nk_handle handle;unsigned short w,h;unsigned short region[4];};
+struct nk_image {nk_handle handle; nk_ushort w, h; nk_ushort region[4];};
+struct nk_9slice {struct nk_image img; nk_ushort l, t, r, b;};
```
lockie, line 484: The naming here might be problematic for other language bindings. In those the `nk_` prefix is usually removed and moved to e.g. package name (or whatever target language provides), so in those bindings we got something like `nk.image` instead of `nk_image`. I'm author of Common Lisp bindings, and there probably wouldn't be the problem there - CL allows for names like `9slice`, but other languages might not.

lockie:
I've took the time to review this patch, and besides the possible issue with naming and language bindings, it is pretty straightforward and LGTM :+1:

nuklear.h
dumblob: Maybe rename to [`scale9` or `grid9` or `slice9`](https://en.wikipedia.org/wiki/9-slice_scaling )?

lockie: Something like `nine_slice` might do the trick as well.

dav64: Renamed 9slice to nine_slice
Let me know if you need anything else :slightly_smiling_face:

lockie (approved):
Looks like it is good to merge :slightly_smiling_face:

Hejsil:
Nice!

RUSshy:
Could the skinning example be updated please? i don't quite understand how to use it

FrostKiwi:
@Michael-Kelley @dav64 
Anyone know what [`nk_ushort l, nk_ushort t, nk_ushort r, nk_ushort b`](https://github.com/Immediate-Mode-UI/Nuklear/blob/master/src/nuklear_9slice.c#L25) stands for? It's a public API, but not documented as to how you are supposed to use it. Anytime you interact with anything 9-slice you have to specify it. I figured it's the square corner size of the 9 slice and have treated it as such, however, that doesn't actually seem to be *quite* true.

What does l t r b stand for? Would love to properly document it.

dav64:
If I remember correctly, it stands for **L**eft, **T**op, **R**ight and **B**ottom

But I cannot remember exactly the specifics/inner working as its been a while since I didn't used Nuklear

Hope this helps

Ashe:
Funnily enough I found myself here also needing help with this.

The letters are indeed top-left-right-bottom, but to my surprise the entire image is used in each segment of the nineslice.

I was EXPECTING it to say 'the top `t` pixels of the image is used for the top 3 segments of the nine-slice', but actually it's 'the entire texture is stretched as each part of the nineslice'

It looks very bizarre, maybe I'm missing something but I don't know how this is supposed to be used other than

```cpp
                nk_nine_slice slice;
                slice.img = *skin->Get();
                slice.b = 50;
                slice.t = 50;
                slice.l = 50;
                slice.r = 50;
                m_context->style.window.fixed_background = nk_style_item_nine_slice(slice);
```
This results in a copy of the image being in the top-left corner as a 50x50 image, then a stretched image of 50 px in height at the top, then another 50x50 in the top right.. Same for all the other sides.

revolucas:
Seemed to work when I tested it. I couldn't get it to look right with smaller resolution images I created but had it working with the example image from here:  https://devforum.roblox.com/t/9-slice-editor-studio-beta/1534739

nineslice


Are you setting the image width, height and region property in your 'Get'?

Ashe:
`Get` is just returning my `nk_image`, which I create in my object like so:

```cpp
image = std::make_unique<struct nk_image>(nk_image_id(*textureID));
```

So yeah, you're right I didn't set the sections.

What is 'region'? Could you give me a snippet of code for where those values come from?

lockie:
I can't get this feature working. I've tried the following

```c
    nk_style_push_style_item(ctx, &ctx->style.window.fixed_background,
                             nk_style_item_nine_slice(nk_nine_slice_handle(
                                 image->handle, 32, 32, 32, 32)));
```

but the result is most bizzare:
2024-09-16-170014_278x275_scrot


Could someone please post a small snippet of the code featuring nine-slice?

lockie:
NVM, I've managed to stumble upon a bug in allegro5 backend. For future readers, 9-slice is as easy as

```
    nk_style_push_style_item(
        ctx, &ctx->style.window.fixed_background,
        nk_style_item_nine_slice(nk_sub9slice_ptr(
            image->handle.ptr, image->w, image->h,
            (struct nk_rect){0, 0, image->w, image->h}, 30, 30, 30, 30)));
```
followed by `nk_style_pop_style_item(ctx)`.

Michael-Kelley:
> @Michael-Kelley @dav64 Anyone know what [`nk_ushort l, nk_ushort t, nk_ushort r, nk_ushort b`](https://github.com/Immediate-Mode-UI/Nuklear/blob/master/src/nuklear_9slice.c#L25) stands for? It's a public API, but not documented as to how you are supposed to use it. Anytime you interact with anything 9-slice you have to specify it. I figured it's the square corner size of the 9 slice and have treated it as such, however, that doesn't actually seem to be _quite_ true.
> 
> What does l t r b stand for? Would love to properly document it.

Hey, sorry for the very _very_ late reply. I somehow missed the notification for this. Those are values for left/top/right/bottom margins. You probably already know that by now seeing as this post is quite old, but I thought I'd still at least respond.

PR #306, opened 2021-06-18
Replace usage of memset in nk_font_atlas_bake with NK_MEMSET

Hejsil:
This wild `memset` prevented me from compiling nuklear without a c standard library :)


lockie:
LGTM :ok_hand:
Its weird how that `memset` slipped through :slightly_smiling_face:

Issue #94, opened 2020-03-31
Macros NK_OFFSETOF, NK_CONTAINER_OF and NK_ALIGNOF rely on undefined behavior


rivten:
When compiling with clang-9 with the option `-fsanitize=undefined` the basic nuklear example, I get the runtime errors : 

`../code/nuklear.h:19196:30: runtime error: member access within null pointer of type 'union nk_page_data'`
`../code/nuklear.h:19197:34: runtime error: member access within null pointer of type 'struct nk_page_element'`

Indeed, the macros `NK_OFFSETOF`, `NK_CONTAINER_OF`, `NK_ALIGNOF` all dereference a null pointer.
One possibility might be to have a `NK_NO_UB` flag or something so that people who want to catch UB errors don't get them when using nuklear.

exchg:
>Indeed, the macros NK_OFFSETOF, NK_CONTAINER_OF, NK_ALIGNOF all dereference a null pointer.

Nope.  Dereferencnig assumes that you will read some data under the pointer. Currently here is pointer arithmetic only.

rivten:
> Nope. Dereferencnig assumes that you will read or write some data under the pointer. Currently here is pointer arithmetic only.

Fair enough. Still, there is an issue somewhere. I tried to modify the macros locally to use `offsetof` in `stddef.h` and change some arithmetic, and then clang does not detect any issue anymore.

Here are my new macros : 

`#define NK_OFFSETOF(st,m) ((nk_ptr)offsetof(st,m))`
`#define NK_CONTAINER_OF(ptr,type,member) (type*)((void*)((char*)(1 ? (ptr) : (ptr) - NK_OFFSETOF(type,member)` (not completely sure about this one)
`#define NK_ALIGNOF(t) NK_OFFSETOF(struct {char c; t _h;}, _h)`

With those, I don't get any runtime error. So either clang-9 is reporting false positive ([but they report that false positive are unlikely](https://clang.llvm.org/docs/UndefinedBehaviorSanitizer.html#issue-suppression)), or either something is fishy with the code. In any case, other users might hit the same issues as I did, so it's probably good to be aware of that.

exchg:
>So either clang-9 is reporting false positive
>or either something is fishy with the code.

Both are possible 

Common 'offsetof' implementation looks like:
```
#define offsetof(st, m) \
    ((size_t)&(((st *)0)->m))

#define offsetof(st, m) \
[...]

PR #309, opened 2021-06-20
Use compiler builtins for NK_OFFSETOF when possible

Hejsil:
My attempt at fixing #94 by using compiler builtins when available.


gpetersson:
Visual C++ compilers older than 6.0 currently cannot compile nuklear due to a lack of alignof/offsetof, so anything that could replace that or make its use unnecessary, would be quite helpful to me.

PR #333, opened 2021-09-13
Update skinning example to use new 9-patch feature

RUSshy:
Still some missing controls that uses colors and borders (ugly), but i figured it would be nice to at least have some controls that make use of the new feature

skinning_NMcKUEaugi

### DIFF
--- a/example/skinning.c
+++ b/example/skinning.c
@@ -44,25 +44,29 @@
 
 struct media {
     GLint skin;
-    struct nk_image menu;
-    struct nk_image check;
-    struct nk_image check_cursor;
-    struct nk_image option;
-    struct nk_image option_cursor;
-    struct nk_image header;
-    struct nk_image window;
-    struct nk_image scrollbar_inc_button;
-    struct nk_image scrollbar_inc_button_hover;
-    struct nk_image scrollbar_dec_button;
-    struct nk_image scrollbar_dec_button_hover;
-    struct nk_image button;
-    struct nk_image button_hover;
-    struct nk_image button_active;
-    struct nk_image tab_minimize;
-    struct nk_image tab_maximize;
-    struct nk_image slider;
-    struct nk_image slider_hover;
-    struct nk_image slider_active;
+    struct nk_nine_slice menu;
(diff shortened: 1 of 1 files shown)


RUSshy:
I found issues with the way it renders 9-slice

Here all good, has enough room to draw the slice:

image


But here, problem:

image


It should check if each slice is inside the rect before rendering it, so we can hide the slices that are too big

RUSshy:
ping @dav64

dav64:
I'll take a look, but I'm not available rn

dav64:
Tested on my sample code + your PR one and cannot reproduce on my side :slightly_frowning_face:

image

Can you send me your modified skinning.c file that demonstrate the problem (or even a simple one with only the progress bar sample if you prefer) ? It will easier to reproduce/fix on my side :slightly_smiling_face: 

If you are using another/modified skin, can you send it also please? 
(thin glossy green progress bars not seems to be part of gwen and it'll help if I have exact same setup as you)

RUSshy:
> Tested on my sample code + your PR one and cannot reproduce on my side 🙁
> 
> image
> 
> Can you send me your modified skinning.c file that demonstrate the problem (or even a simple one with only the progress bar sample if you prefer) ? It will easier to reproduce/fix on my side 🙂
> 
> If you are using another/modified skin, can you send it also please?
> (thin glossy green progress bars not seems to be part of gwen and it'll help if I have exact same setup as you)

The one in the PR is untouched, i encountered the issue so i didn't include the change

Here is the full skinning.c file with the progress bar that is using the 9-patch

https://gist.github.com/RUSshy/cd7cc435819c71a74653d3364bc07d22

dav64:
Sorry for delay, not much time at the moment (private reasons)
Here is a temp fix of nuklear.h (hacky one at the moment, I don't like it) : If the slice is bigger than the region, I draw the original image stretched as-is (without slices)
https://gist.github.com/dav64/b2aa5c5f5bc3d2ad2cf7edd4e7fc8a5d

Once I have more time, I will work on a proper fix to merge that will use the slices and get rid of the center if region too small

PR #336, opened 2021-09-15
Fix "'num_len' may be used uninitialized" in nk_do_property

gpetersson:
Re-created PR#27 (not me) since this is still an issue, gcc 11.1 on Linux here.


Hejsil:
Thanks :+1:

PR #337, opened 2021-09-16
Added nk_spacer to help adjusting layout

joeld42:
I'm working on a fork of nuklear with some customizations for theming. I'll submit PRs for changes I make that I think will be generally useful. But feel free to ignore or reject if it doesn't fit.

This is a tiny little fix I use all over the place, `nk_spacer(ctx)` will allocate space like any other widget, but not do anything with it. It's useful for adding space between rows or between widgets in a row.

### DIFF
--- a/nuklear.h
+++ b/nuklear.h
@@ -2577,6 +2577,21 @@ NK_API struct nk_rect nk_layout_space_rect_to_screen(struct nk_context*, struct 
 /// Returns transformed `nk_rect` in layout space coordinates
 */
 NK_API struct nk_rect nk_layout_space_rect_to_local(struct nk_context*, struct nk_rect);
+
+/*/// #### nk_spacer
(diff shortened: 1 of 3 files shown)


Hejsil:
One can get equivalent behaviro by using `nk_label(ctx, "", 0)`, but the intent is probably a lot less clear and `nk_label` is a more complicated function to execute, just for an empty slot. What's your opinion on a name like `nk_dummy` or `nk_empty`. I'm not sure what I would call this myself.

Also, to get this merged we need the minor version to be bumped in `package.json` and `CHANGELOG`

Hejsil:
Merged here ff62698de8e7e94797cff380f4e84ecb9ec19d96

@lockie Thanks for the review. The version bump matches the standard (new feature = minor bump), so this is good

PR #338, opened 2021-09-16
Add separate padding for progress bar background

joeld42:
Added a separate outer padding for progress bars. This is useful when you need some padding for your background image as well as for the inner filled progress image. I named this `padding_outer` rather than calling it `padding` and renaming the existing one to `padding_cursor` or something like that so that this will be backwards compatible if people are already using padding on progress bar, but I'd be happy to rename it.

progress_bar

### DIFF
--- a/src/nuklear_progress.c
+++ b/src/nuklear_progress.c
@@ -96,25 +96,33 @@ nk_do_progress(nk_flags *state,
     float prog_scale;
     nk_size prog_value;
     struct nk_rect cursor;
+    struct nk_rect outer;
 
     NK_ASSERT(style);
     NK_ASSERT(out);
     if (!out || !style) return 0;
 
+    /* calculate progressbar outer rect */
+    outer.w = NK_MAX(bounds.w, 2 * (style->padding_outer.x + style->border ) );
+    outer.h = NK_MAX(bounds.h, 2 * (style->padding_outer.y + style->border ) );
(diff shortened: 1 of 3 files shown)


Hejsil:
Looking into it a little, `margin` seems to be the correct term for spacing outside a box (progress bar in this case). Seems there are quite a few elements where margin might be a useful concept (probably not in scope for this PR).

Also, to get this merged, we need the minor version to be bumped in `package.json` and `CHANGELOG`

src/nuklear_progress.c
```diff
@@ -96,25 +96,33 @@ nk_do_progress(nk_flags *state,
...
     /* update progressbar */
     prog_value = NK_MIN(value, max);
-    prog_value = nk_progress_behavior(state, in, bounds, cursor,max, prog_value, modifiable);
+    prog_value = nk_progress_behavior(state, in, outer, cursor,max, prog_value, modifiable);
     cursor.w = cursor.w * prog_scale;
 
     /* draw progressbar */
     if (style->draw_begin) style->draw_begin(out, style->userdata);
-    nk_draw_progress(out, *state, style, &bounds, &cursor, value, max);
+    nk_draw_progress(out, *state, style, &outer, &cursor, value, max);
```
RobLoach, line 125: Sounds like a a change from padding_outer and outer to "margin" makes sense. Thanks! Looking great so far.

RobLoach (changes requested):

PR #343, opened 2021-09-22
GCC __builtin_offsetof only exists in version 4 and later

gpetersson:
Early OSX compiler toolchains used gcc 2.95.2, 3.1 and 3.3, and offsetof was not added until gcc 4.


Hejsil:
Thanks!

PR #347, opened 2021-09-26
GCC __printf__ format attribute only exists in version 3.2 and later

gpetersson:
The format attribute `__printf__` is equivalent to `printf` but the latter works in gcc versions older than 3.2.


gpetersson:
Actually, versions under 3.2 won't work anyways, because they don't support math operations in the parameter list (`syntax error, found "+"`). Would you suggest just turning the attribute off entirely for < 3.2 ?

PR #348, opened 2021-09-26
Fix "may be used uninitialized" warnings in nk_widget

aganm:


src/nuklear_widget.c
```diff
@@ -21,6 +21,7 @@ NK_API struct nk_vec2
 nk_widget_position(struct nk_context *ctx)
 {
     struct nk_rect bounds;
+    bounds.x = 0;
```
Hejsil, line 24: Hmm. This doesn't seem like the right way to fix this warning. Where exactly does the warning point to?

aganm: This is the warnings I got:

```
/home/aganm/Projects/Nuklear/nuklear.h: In function ‘nk_widget_position’:
/home/aganm/Projects/Nuklear/nuklear.h:22851:12: warning: ‘bounds.x’ may be used uninitialized in this function [-Wmaybe-uninitialized]
22851 |     return nk_vec2(bounds.x, bounds.y);
      |            ^~~~~~~~~~~~~~~~~~~~~~~~~~~

/home/aganm/Projects/Nuklear/nuklear.h: In function ‘nk_widget_width’:
/home/aganm/Projects/Nuklear/nuklear.h:22875:18: warning: ‘bounds.w’ may be used uninitialized in this function [-Wmaybe-uninitialized]
22875 |     return bounds.w;
      |            ~~~~~~^~

/home/aganm/Projects/Nuklear/nuklear.h: In function ‘nk_widget_height’:
/home/aganm/Projects/Nuklear/nuklear.h:22887:18: warning: ‘bounds.h’ may be used uninitialized in this function [-Wmaybe-uninitialized]
22887 |     return bounds.h;
      |            ~~~~~~^~
```
Compiled with `gcc -Wall -Wextra`

aganm: It might make more sense to initialize the entire struct to 0? I just did the minimum to silence the warnings.

Hejsil: I see. So because gcc has analyzed that `nk_layout_peek` can (in theory, not in practice) return before initializing `bounds` it gives this warning. Do you know if this warning can be silenced without initializing the struct?

Hejsil: If not, then maybe we can initialize the `bounds` to zero in `nk_layout_peek` before it does its early return

aganm: I changed it to your suggestion (initialize `bounds` to zero in `nk_layout_peek` before it does early return). Indeed a better solution since the code shouldn't reach that anyway and it fixes the warnings.

Hejsil:
:+1:

PR #354, opened 2021-10-07
Add command to draw polyline from float*

Andersama:
Adds a command which passes a pointer to use for a polyline. Skips pushing data to buffer as an array of shorts and back.

Note: Assumes pointer is user managed and lives until frame is complete*
Usage: When an array of floats is already in memory as an array of `nk_vec2{x,y}`, same call signature as `nk_stroke_polyline()`

Can improve on programs which have a smaller fixed buffer for commands.
Names and enum values tentative, feel free to edit. Similar functions could be added for polygons.

### DIFF
--- a/src/nuklear.h
+++ b/src/nuklear.h
@@ -4203,7 +4203,8 @@ enum nk_command_type {
     NK_COMMAND_POLYLINE,
     NK_COMMAND_TEXT,
     NK_COMMAND_IMAGE,
-    NK_COMMAND_CUSTOM
+    NK_COMMAND_CUSTOM,
+    NK_COMMAND_POLYLINE_FLOAT
 };
 
 /* command base and header of every command inside the buffer */
@@ -4335,6 +4336,14 @@ struct nk_command_polyline {
     unsigned short line_thickness;
     unsigned short point_count;
     struct nk_vec2i points[1];
+};
+
+struct nk_command_polyline_float {
+    struct nk_command header;
+    struct nk_color color;
+    unsigned short line_thickness;
+    unsigned short point_count;
+    struct nk_vec2 *points;
 };
 
 struct nk_command_image {
--- a/src/nuklear_draw.c
+++ b/src/nuklear_draw.c
@@ -392,6 +392,23 @@ nk_stroke_polyline(struct nk_command_buffer *b, float *points, int point_count,
     }
 }
 NK_API void
+nk_stroke_polyline_float(struct nk_command_buffer *b, float *points, int point_count,
+    float line_thickness, struct nk_color col)
+{
+    nk_size size = 0;
+    struct nk_command_polyline_float *cmd;
+
+    NK_ASSERT(b);
+    if (!b || col.a == 0 || line_thickness <= 0) return;
+    size = sizeof(*cmd);
+    cmd = (struct nk_command_polyline_float*) nk_command_buffer_push(b, NK_COMMAND_POLYLINE_FLOAT, size);
+    if (!cmd) return;
+    cmd->color = col;
+    cmd->point_count = (unsigned short)point_count;
+    cmd->line_thickness = (unsigned short)line_thickness;
+    cmd->points = (struct nk_vec2*)points;
+}
+NK_API void
 nk_draw_image(struct nk_command_buffer *b, struct nk_rect r,
     const struct nk_image *img, struct nk_color col)
 {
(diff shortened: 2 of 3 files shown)


Hejsil:
Aah, right. I see what you are going for now. You want a "non-owning" (better name pending) API, as you have an array of floats in memory already and don't want to pay the memory cost of it twice and the cost of the copy. I think there could be a place for such things. One problem I see though is that most current users probably don't care about this "non-owning" API, but will now have to implement code in their backends to handle commands they don't actually use.

One thing we could look into is to have an option to enable "non-owning" commands (`NK_INCLUDE_NON_OWNING_COMMANDS`). Having this options allows the change to be non-breaking if it is off by default. This option then includes these new commands:
```
    NK_COMMAND_POLYGON_NON_OWNING,
    NK_COMMAND_POLYGON_FILLED_NON_OWNING,
    NK_COMMAND_POLYLINE_NON_OWNING,
    NK_COMMAND_TEXT_NON_OWNING,
```

These commands contain pointers to their data instead of embedding the data inside the command.

@dumblob Any opinions on this?

Andersama:
Is there a particular reason why `nk_vertex_draw` is written as a switch block in a loop vs a bunch of loops in a switch? I just gained maybe a 2x boost by just doing this:
```c
    if (format == NK_FORMAT_FLOAT) {
        NK_MEMCPY(attribute, values, sizeof(float)*value_count);
        return;
    }
 ```
 It looks like this is a pattern in a good bit of the call stack before it gets there.

dumblob:
Sorry again for the delay.

>@dumblob Any opinions on this?

First I acknowledge that there is need for this (as resource/memory management with immediate mode libs is easy/seamless and thus such copying is not necessary and shouldn't really decrease app safety almost in any way nor increase memory management burden).

I think adding corresponding new commands is the way to go. In practice I think we're totally free to do it even without updating any of our backend demos as they're officially not part of Nuklear. On the other hand I feel compassion with those "just wanting to update Nuklear to its latest" and suddenly needing to accommodate to these changes.

I'd say we should start planning for a new **major** version bump of Nuklear. Create a tracker issue linking all the problems which are better solved by rather less backwards-compatible solution (like this one) than adding "hacks" (what is a "hack" is subjective and YMMV but you get the idea I believe). Then make separate PRs for each of them and then do the major bump with all of them at once.

For example I'd consider adding the enum at the end as a hack (basically signalling "yeah, upstream thinks & willingly supports you not updating and coming back with issues if this hack won't work in your super specific weird setup".

So to wrap up - I'd go with adding new commands while bumping the major version, but not hastily but rather in conjunction with some other ideas/issues we have or know of.

I don't know yet how to best structure these new commands to not introduce too many forcing everybody to implement them. Maybe even some meta-command(s) (imagine USR1 & USR2 as analogy) meant as fully app-specific (i.e. backends/demos would be obliged to **not** implement them themself but rather throw it at the app - which could opt-in or opt-out e.g. by defining some macro etc.) could make sense. This needs discussion. We must not forget about bindings to many languages which further constraints what we can do (e.g. ABI compatibility should have IMHO lower priority than bindings-friendliness).

Thoughts?

(I don't guarantee I'll fully participate in the discussion due to lack of time, but also if @Andersama wants to join our organization, then why not - just review maybe 2 existing PRs and you'll get full admin rights here as we need reviewers and increased bus factor of the org :wink:)

> It looks like this is a pattern in a good bit of the call stack before it gets there. Edit: since I only noticed a bit ago that a lot of assembly was unoptimized I realized I was missing some key defines, I'll have to recheck what the performance is like for doing this.

Please report back your findings - if those improvements won't make the code less understandable and the gains are quite significant (IDK, let's say >1.3x speedup overall or of bottle necks), there is a good chance of merging them :wink:.

Andersama:
Had a thought, is there anything conceptually about nuklear that would prevent commands from being optimized as they're added in? For example with the charts api, is there anything particularly outstanding that would prevent the api from checking the state of the command buffer and modifying it? EG: if the api could recognize that two lines were drawn in succession w/ matching endpoints the api could transform the command into a polyline command?

Andersama:
I completely forgot I wrote this PR, was anything like this added?

RobLoach:
Nothing this this was implemented. Couple thoughts...
1. I like the idea, would be a nice way to improve fidelity on vector-based graphics
2. Could you implement the polyline float command on some of the renderers to get a few examples of how to use it? The SDL3 one may be a good start.
3. Do you have plans on having some of the components in Nuklear use this? May be good for the graphs maybe?

PR #359, opened 2021-10-19
Add non-owning draw api command

Andersama:
Rough idea of a non-owning draw call, not user friendly, expects the buffers provided are ready to draw as is.


dumblob:
Yep, this was the "base" of the idea I believe.

The question is for which commands a non-owning variant would make sense from pratical point of view (not every command potentially benefiting from a non-owning counterpart for performance reasons would be used in practice IMHO). Any guesses?

Also we should then compare it to other approaches discussed in the linked discussion...

PR #365, opened 2021-11-15
Rename package.json to clib.json

Aloz1:
The clib package manager has deprecated package.json in favor of clib.json (probably to reduce confusion between clib and other package managers, e.g. npm). As far as I can tell, there's no difference in file format, only in name.

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -185,7 +185,7 @@ Reviewing changes to `src/*` and `nuklear.h`:
   * Variable/function name casing.
   * Indentation.
   * Curly bracket (`{}`) placement.
-* Ensure that the contributer has bumped the appropriate version in
+* Ensure that the contributor has bumped the appropriate version in
   [clib.json](https://github.com/Immediate-Mode-UI/Nuklear/blob/master/clib.json)
   and added their changes to the
   [CHANGELOG](https://github.com/Immediate-Mode-UI/Nuklear/blob/master/src/CHANGELOG).
--- a/Readme.md
+++ b/Readme.md
@@ -186,7 +186,7 @@ Reviewing changes to `src/*` and `nuklear.h`:
   * Indentation.
   * Curly bracket (`{}`) placement.
 * Ensure that the contributer have bumped the appropriate version in
-  [package.json](https://github.com/Immediate-Mode-UI/Nuklear/blob/master/package.json)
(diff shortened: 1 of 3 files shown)


Aloz1:
Note, it seems that irrespective of this change, acquiring nukular via clib is currently broken.

It seems that clib now expects a tag to exist for the version specified in package.json/clib.json.

PR #287 will need to be merged to un-break clib support.

RobLoach (approved):
Seems appropriate.

dumblob:
Yep, to me too. I wouldn't bump the version as it's only repository-related and not Nuklear (the library/code) related. What do you think?

E.g. if you have a counter example (because of some build tool or whatever dependency), then I'm fine merging this with the bump :wink:.

src/CHANGELOG
```diff
@@ -8,6 +8,7 @@
 ///    - [yy]: Minor version with non-breaking API and library changes
 ///    - [zz]: Bug fix version with no direct changes to API
 ///
+/// - 2021/11/16 (4.09.1) - Renamed package.json to clib.json
```
RobLoach, line 11:
``` suggestion
```

clib.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "nuklear",
-  "version": "4.09.0",
+  "version": "4.09.1",
```
RobLoach, line 3:
``` suggestion
  "version": "4.09.0",
```

RobLoach (changes requested):
Removing the version bump in clib.json

Aloz1:
No counter examples or objections. Have reverted version bump in 8052d27.

clib.json
Aloz1: Fixed in 8052d27

src/CHANGELOG
Aloz1: Fixed in 8052d27

Readme.md
```diff
@@ -185,8 +185,8 @@ Reviewing changes to `src/*` and `nuklear.h`:
   * Variable/function name casing.
   * Indentation.
   * Curly bracket (`{}`) placement.
-* Ensure that the contributor has bumped the appropriate version in
-  [package.json](https://github.com/Immediate-Mode-UI/Nuklear/blob/master/package.json)
+* Ensure that the contributer has bumped the appropriate version in
```
RobLoach, line 188:
``` suggestion
* Ensure that the contributor has bumped the appropriate version in
```

PR #374, opened 2021-12-15
clib: Add MIT license

RobLoach:


dumblob:
It's actually dual-licensed - MIT & Unlicense (public domain). Please adjust it accordingly :wink:.

clib.json
```diff
@@ -4,5 +4,6 @@
   "repo": "Immediate-Mode-UI/Nuklear",
   "description": "A small ANSI C gui toolkit",
   "keywords": ["gl", "ui", "toolkit"],
+  "license": "MIT",
```
RobLoach, line 7:
``` suggestion
  "license": "MIT, Unlicense",
```
Looks like it just loads a string, so I suppose we can stick any comma-seperated SPDX text in here?
https://github.com/clibs/clib/blob/master/src/common/clib-package.c#L481

dumblob:
Thanks!

PR #375, opened 2021-12-15
Update stb_rect_pack and stb_truetype

RobLoach:
Some updates to stb_rect_pack and stb_truetype over from https://github.com/nothings/stb .

### DIFF
--- a/clib.json
+++ b/clib.json
@@ -1,6 +1,6 @@
 {
   "name": "nuklear",
-  "version": "4.09.0",
+  "version": "4.09.1",
   "repo": "Immediate-Mode-UI/Nuklear",
   "description": "A small ANSI C gui toolkit",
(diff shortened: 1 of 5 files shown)


dumblob:
Did you do some tests (ideally with several different fonts) to see how the fonts look like, if there is no blurriness (not more than it already is), no overlaps, no cut-offs, etc.?

Fonts are PITA but one of the selling points of any GUI lib, so we have to be extra careful...

RobLoach:
There's the Allegro demo using Comic Sans italic.... Because.

Screenshot at 2021-12-16 12-00-11

dumblob:
LGTM - please bump version again and feel free to merge it if @Hejsil is not against :wink:.

Hejsil (approved):

PR #376, opened 2021-12-15
Add EditorConfig to manage editor whitespace

RobLoach:
[EditorConfig](https://editorconfig.org) is a utility for various code editors which will automatically configure the editor to manage whitespace as desired.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -9,3 +9,6 @@ end_of_line = lf
 indent_size = 4
 insert_final_newline = true
 trim_trailing_whitespace = true
+
+[clib.json]
+indent_size = 2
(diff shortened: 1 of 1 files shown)


dumblob:
Interesting idea. I'm quite indifferent on this. I see no strong arguments for having it but also no strong argument for not having it :wink:.

What do others think? @Hejsil ?

Hejsil:
I haven't really used editor configs myself, but I don't see a reason not to have it. We should probably also have this editor config specify that the `clib.json` file have `indent_size = 2` for completeness

RobLoach:
I find editorconfig handy across both VSCode and vim, as they both pick up the settings when you have the plugins installed, and then automagically fixes your whitespace for you. Absolutely not needed though. Just helpful.

Hejsil:
:+1: Let's give it a go

dumblob:
Thanks!

PR #383, opened 2021-12-19
docs: Rename nuklear.html to index.html

RobLoach:
`index.html` is the standard main document that is loaded in folder structures. Given that, it would make sense to use `index.html` rather than `nuklear.html` to that when you load up the /doc directory in your web browser, you get the actual documentation.

This also retains the `nuklear.html` file as a redirect to the new `index.html`.


dumblob:
Thanks!

PR #384, opened 2021-12-19
changelog: Use x.y.z for versioning, rather than x.yy.zz

### LINKED ISSUE #382: Version strategy: Do not prefix with 0
Right now, we use a [semver](https://semver.org/)-esque strategy, but we are prefixing `#.##.#`, for example `4.09.1`. This is fine for having the versions align in nuklear.h, but it doesn't really make sense for semantic version numbering. It should likely be `#.#.#`.

I'm proposing removing the prefixing `0` from the minor number in the middle to more align with actual semantic versioning.

RobLoach:
This changes the versioning strategy from `x.yy.zz` to `x.y.z` to more align with [semantic versioning](https://semver.org/).

Fixes #382

### DIFF
--- a/nuklear.h
+++ b/nuklear.h
@@ -29613,13 +29613,12 @@ nk_tooltipfv(struct nk_context *ctx, const char *fmt, va_list args)
 
 /// ## Changelog
 /// ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~none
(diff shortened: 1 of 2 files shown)


dumblob:
While at it, maybe it's time to mention it's Semver?

dumblob:
Awesome, thanks! And thanks for cleaning the issue tracker!

PR #385, opened 2021-12-22
gdi: x11: Add ARC Command

RobLoach:
Fixes #112

The arc isn't quite right on x11. The bottom right on the canvas.

Screenshot at 2021-12-22 12-40-44

### DIFF
--- a/demo/gdi/nuklear_gdi.h
+++ b/demo/gdi/nuklear_gdi.h
@@ -75,7 +75,7 @@ nk_create_image(struct nk_image * image, const char * frame_buffer, const int wi
         image->region[1] = 0;
         image->region[2] = width;
         image->region[3] = height;
-        
+
         bi.bmiHeader.biSize = sizeof(BITMAPINFOHEADER);
         bi.bmiHeader.biWidth = width;
         bi.bmiHeader.biHeight = height;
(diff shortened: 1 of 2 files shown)


RobLoach:
@bmuesslg Mind giving it a review? I'm unable to test this one.

bmuessig:
I've built the changes and the demo looks like this on Win32/GDI: 
grafik

I'll give the same demo a try with SDL and post the result.

bmuessig:
This is how it looks like using SDL:
image

So this fix is unfortunately not quite working. I will have a look what's going wrong.

bmuessig:
Alright, I fixed it:
grafik

RobLoach:
Thanks for fixing that up! Looks like from this Pull Request, there's the demo/x11/nuklear_xlib.h changes too. :thinking:

riri:
> Thanks for fixing that up! Looks like from this Pull Request, there's the demo/x11/nuklear_xlib.h changes too. thinking

I have to update my nk working copy, but I can make such a test.

riri:
What's the status of this?
I see in master the command for gdi is coded from commit cacdf6b
I have implemented the x11 part in commit d63f106
Beside some whitespace fixes, are there improvements to add?
If not maybe that PR could be closed?

RobLoach:
I think this is done, yeah. thanks.

PR #386, opened 2021-12-22
demo: Add clean to Makefiles, and don't use bin directory

RobLoach:
This adds a `make clean` command to all the demo Makefiles, and skips the use of a `bin` directory to clean up the build. It ends up building into the same directory instead to make it a simpler build process.


RobLoach:
I'm conflicted on the build directory :thinking:

PR #387, opened 2021-12-22
Revert "Fix layout bounds not accounting for padding"

RobLoach:
Reverts Immediate-Mode-UI/Nuklear#156

@aganm @Hejsil This change reverts the layout bounds change. May be worth considering removing the change until we fix it for good. What do you think?

### DIFF
--- a/clib.json
+++ b/clib.json
@@ -1,6 +1,6 @@
 {
   "name": "nuklear",
-  "version": "4.09.3",
+  "version": "4.09.1",
   "repo": "Immediate-Mode-UI/Nuklear",
   "description": "A small ANSI C gui toolkit",
   "keywords": ["gl", "ui", "toolkit"],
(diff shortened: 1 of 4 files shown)


RobLoach:
@Hejsil Considering https://github.com/Immediate-Mode-UI/Nuklear/issues/390 , there may be other regressions we're seeing out of this. Mind if we revert and then fix it up afterwards?

aganm:
I just tried #160 and it had none of the regressions we found with #156.
Some pieces must be missing for it to work correctly.
Personally I would revert #156, double check #160 and pull it in.

Hejsil:
> @Hejsil Considering #390 , there may be other regressions we're seeing out of this. Mind if we revert and then fix it up afterwards?

Sure. I dont have much time until January to look into this so lets revert. Sorry for the breakage 🙏

RobLoach:
No worries! Can be difficult to check all features. Hope we can get some automated tests going at some point to check.

PR #400, opened 2022-01-25
Integrate file browser with glfw opengl2 demo

crazyBaboon:


demo/glfw_opengl2/Makefile
```diff
@@ -2,7 +2,7 @@
 BIN = demo
 
 # Flags
-CFLAGS += -std=c89 -Wall -Wextra -pedantic -O2
+CFLAGS += -std=c99 -Wall -Wextra -pedantic -O2
```
RobLoach, line 5: We do target C89 across all the other demos. Mind if we keep this?

RobLoach:
This is looking great! Nicely done. Would love to see this filebrowser become more generic, so thanks a lot for cleaning it up.

demo/glfw_opengl2/Makefile
dumblob: Let me clarify that C89 is just a nice-to-have when it comes to code which won't end up in `nuklear.h`. If there are no other issues with this PR and if @crazyBaboon will not want to make it C98-compatible, we'll merge it.

crazyBaboon: The problem are the C++ style comments in the stb_image.h.

Ideally, these should be replaced by C style comments, or else it won't compile (with the C89 flag) using some mainstream compilers. But that is a PITA. 

Does anyone know if there is a way of using C++ comments with the C89 flag?

RobLoach: Awesome! Sounds good with me :+1:

crazyBaboon: I would say that for now we change to C99.

If we find a way to replace the C++ comments from the stb_image.h, then we would do that later.

demo/glfw_opengl2/main.c
```diff
@@ -115,6 +148,23 @@ int main(void)
...
+    media.icons.home = icon_load("/home/pinto-pc/Documents/Nuklear/demo/glfw_opengl2/icon/home.png");
+    media.icons.directory = icon_load("/home/pinto-pc/Documents/Nuklear/demo/glfw_opengl2/icon/directory.png");
+    media.icons.computer = icon_load("/home/pinto-pc/Documents/Nuklear/demo/glfw_opengl2/icon/computer.png");
+    media.icons.desktop = icon_load("/home/pinto-pc/Documents/Nuklear/demo/glfw_opengl2/icon/desktop.png");
+    media.icons.default_file = icon_load("/home/pinto-pc/Documents/Nuklear/demo/glfw_opengl2/icon/default.png");
+    media.icons.text_file = icon_load("/home/pinto-pc/Documents/Nuklear/demo/glfw_opengl2/icon/text.png");
+    media.icons.music_file = icon_load("/home/pinto-pc/Documents/Nuklear/demo/glfw_opengl2/icon/music.png");
+    media.icons.font_file =  icon_load("/home/pinto-pc/Documents/Nuklear/demo/glfw_opengl2/icon/font.png");
+    media.icons.img_file = icon_load("/home/pinto-pc/Documents/Nuklear/demo/glfw_opengl2/icon/img.png");
+    media.icons.movie_file = icon_load("/home/pinto-pc/Documents/Nuklear/demo/glfw_opengl2/icon/movie.png");
```
RobLoach, line 163: These should likely be relative paths.

RobLoach:
``` suggestion
    media.icons.home = icon_load("icon/home.png");
    media.icons.directory = icon_load("icon/directory.png");
    media.icons.computer = icon_load("icon/computer.png");
    media.icons.desktop = icon_load("icon/desktop.png");
    media.icons.default_file = icon_load("icon/default.png");
    media.icons.text_file = icon_load("icon/text.png");
    media.icons.music_file = icon_load("icon/music.png");
    media.icons.font_file =  icon_load("icon/font.png");
    media.icons.img_file = icon_load("icon/img.png");
    media.icons.movie_file = icon_load("icon/movie.png");
```

RobLoach (changes requested):
These should likely be relative :wink:

crazyBaboon:
This needs a bit more work IMO, since when you comment **#define INCLUDE_FILE_BROWSER** in main.c, this crashes nuklear. I have started moving all the file browser code to the file_browser.c file, but haven't got it right yet, plus my computer power supply collapsed, so I am unsure when this will be ready.

demo/glfw_opengl2/main.c
crazyBaboon: Ooops - I forgot about that!

crazyBaboon: Done!

crazyBaboon:
I used pre-processor guards to seperate the components of the file browser. I have tested it, and it works :)

Let me know what you think!

dumblob:
Could we go a step further and create `demo/_shared/filebrowser/` and
1. move the `demo/glfw_opengl2/icon/` folder into it (to end up as `demo/_shared/filebrowser/icon/`)
2. move `demo/glfw_opengl2/stb_image.h` into it (to end up as `demo/_shared/filebrowser/stb_image.h`)
3. move all `demo/*.c` files (incl. `file_browser.c`) into `demo/_shared/` (to end up as `demo/_shared/*.c`)
4. adjust all `demo/*/*.c` files accordingly

Or do you have any other ideas how to avoid duplications and structure it logically yet simply?

crazyBaboon:
@dumblob 

That seems to be a good idea. Why `_shared` and not simply `shared`?

dumblob:
>Why `_shared` and not simply `shared`?

Just a poor man's attempt to distinguish a proper "shared" demo (which doesn't exist but might exist once - maybe it'll be something with clipboard or whatever) and a "structural constraints of the repository itself" :wink:. Feel free to come up with a better yet simple naming scheme.

crazyBaboon:
@dumblob 

What about `common`?

dumblob:
I'm not against - it's just the feeling I have that anything taken from a natural language (or from an IT slang) would evoke anything but "not a demo, only common utilities & helpers". Therefore I thought that an underscore - unofficially **the** way to denote some *private* or otherwise *special* identifiers in C - might convey what we want. But IDK - maybe really a different directory structure would be needed. Or maybe (starts to sound most probable to me now when I'm writing it) I'm exaggerating it and it doesn't matter much.

`lib` or `common` `_lib` `_common` etc. all would be acceptable IMHO.

What do others think?

crazyBaboon:
@dumblob 

I decided to use 'common'. Everything seems to be in the right place!

dumblob:
Merged, thanks!

PR #403, opened 2022-01-29
SDL demos: Favor switch over if-else chain in event handler.

lib-omar:


RobLoach (approved):
I'm liking this clean up.

PR #413, opened 2022-02-21
Android SDL opengles2 implementation

leonelhs:
Hello friends!
I have written a Nuklear Android demo based on SDL opengles2. 
this PR has zero modifications  to the main core. It just contain a new demo  source code.

I would like to share this demo with other devs in order to get feedbacks and keep improving mobile implementations.

My congratulations for this awesome project.


dumblob:
I totally want to merge this PR! Thanks for the effort!

I'm now contemplating the following:

1. The PR is huge because of a lot of generated files. I think we should definitely provide a way (a simple script in shell/Python/... or in the worst case a maximally detailed `TODO.md` file) describing how to generate the files instead of the files themself.
2. In addition to (1), should we also provide **some of** the pre-generated files in case the script from (1) stops working (an evergreen when it comes to programming for Android), so that one still has at least something to quickly start with (before someone corrects the script in this repo)?

What do people programming for recent Android think about this?

And what do others think?

leonelhs:
I made an script to build the demos, SDL sources are symbolic links into the projects. Anyways you most have a SDL sources copy.

I think this project has to much potential, specially for game devs, its hard to write a decent UI from the scratch for mobile devices since you can’t mix platform widgets whit hardware surfaces.

All ready there is a lot of people looking at the SDL forums for info and how to implement Nuklear in his projects. 
Actually SDL forums lead me tho here.

demo/android/sdl2surface_rawfb/build.sh
```diff
@@ -0,0 +1,83 @@
+#!/bin/bash
+#
+# This script is based on SDL sources => androidbuild.sh
+#
+##
+
+SOURCES=()
+MKSOURCES=""
+CURDIR=`pwd -P`
```
dumblob, line 9: ``` `pwd -P` ``` is dangerous here. I'd prefer `"$( pwd -P )"` instead :wink:.

demo/android/sdl2surface_rawfb/build.sh
```diff
@@ -0,0 +1,83 @@
...
+    echo "Please set the SDLPATH directory to the path of the SDL2 source code"
+    exit 1
+fi
+
+if [ -z "$ANDROID_HOME" ];then
+    echo "Please set the ANDROID_HOME directory to the path of the Android SDK"
+    exit 1
+fi
+
+if [ ! -d "$ANDROID_HOME/ndk-bundle" -a -z "$ANDROID_NDK_HOME" ]; then
```
dumblob, line 22: Should be `if [ ! -d "$ANDROID_HOME/ndk-bundle" ] && [ -z "$ANDROID_NDK_HOME" ]; then` as more than one comparison will never be supported according to the latest POSIX (yes, we're in bash here, but while I'm at nitpicking... :wink:).

demo/android/sdl2surface_rawfb/build.sh
```diff
@@ -0,0 +1,83 @@
...
+    exit 1
+fi
+
+APP="com.sdl2rawfb.nuklear"
+APPARR=(${APP//./ })
+BUILDPATH="$CURDIR/../build/$APP"
+
+# Start Building
+
+rm -rf $BUILDPATH
```
dumblob, line 33: -> `"$BUILDPATH"`

demo/android/sdl2surface_rawfb/build.sh
```diff
@@ -0,0 +1,83 @@
...
+mkdir -p $BUILDPATH
```
dumblob, line 34: Same goes here and in all the many following lines :wink:.

demo/android/sdl2surface_rawfb/build.sh
```diff
@@ -0,0 +1,83 @@
...
+import org.libsdl.app.SDLActivity;
+
+public class $ACTIVITY extends SDLActivity
+{
+
+}
+__EOF__
+
+# Update project and build
+cd $BUILDPATH
```
dumblob, line 82: ...until here :smile:.

demo/android/sdl_opengl2/.idea/workspace.xml
```diff
@@ -0,0 +1,350 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
dumblob, line 1: Is this `.idea/` folder needed at all?

demo/android/sdl_opengl2/build.sh
```diff
@@ -0,0 +1,97 @@
+#!/bin/bash
```
dumblob: Same comments as for the script above apply also here :smile:.

dumblob:
I feel we're on the right path :wink:.

Btw. could you also create e.g. `demo/android/README.md` with a detailed code snippet which would fetch the newest released version of https://github.com/libsdl-org/SDL/tree/main/android-project , rename it to `project-template.new/`, move all the needed stuff from the current `project-template/` to the new one, then rename `project-template/` to `project-template.old/`, then rename `project-template.new/` to `project-template/` and then finally remove `project-template.old/`?

demo/android/sdl2surface_rawfb/build.sh
leonelhs: done  😉

demo/android/sdl2surface_rawfb/build.sh
leonelhs: done  😉

demo/android/sdl_opengl2/.idea/workspace.xml
leonelhs: no needed and deleted  😉

demo/android/sdl2surface_rawfb/build.sh
leonelhs: Sorry I don’t understand  ...until here😳, I hope the Readme.md can clarify how the script is working

dumblob:
Ok, I've taken another look and gave a thought to it. I think we need the following and then I will merge this cool PR!

1. remove SDL sources from this PR altogether
2. if the SDL sources in this PR are patched by you, then provide only the patches **and** a makefile or at least a detailed description how to apply them

Let me know if there are any obstackles.

I would also like to apologize for commenting on the `build.sh` script - I've compared it to the SDL upstream's `androidbuild.sh` and almost fell over as they don't use `""` around variables etc. But because it's much more important to stay as close to upstream as possible, I will not require these to be fixed in your PR to minimize the diff to the upstream file. I'll instead consider filing a bug in the SDL upstream tracker once I will recover from the fell over :wink:.

leonelhs:
SDL sources are removed 😁.
Feel free to comment or improve scripts.

MaxGyver83:
Great work! I know that it's old but both implementations still compile!

PR #414, opened 2022-02-25
Windows (GDI): Native like nuklear windows

Ohjurot:
This is a preview PR to show my progress on building a C++ class that allows a nuklear window to behave the same as any other Windows window. The class is coming from a project that I’m currently writing so it is still rough. I will spend some more time in converting the style to the nuklear style (Maybe even converting the code from C++ to C). Already looking forward for some feedback and suggestions.

### DIFF
--- a/demo/gdi_native_nuklear/build.bat
+++ b/demo/gdi_native_nuklear/build.bat
@@ -1,0 +1,6 @@
+@echo off
+
+rem This will use VS2015 for compiler
+call "C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\vcvarsall.bat" x86
+
+cl /nologo /W3 /O2 /fp:fast /Gm- /D_CRT_SECURE_NO_DEPRECATE /Fedemo.exe main.cpp user32.lib gdi32.lib Msimg32.lib /link /incremental:no 
--- a/demo/gdi_native_nuklear/main.cpp
+++ b/demo/gdi_native_nuklear/main.cpp
@@ -1,0 +1,37 @@
+#include <Windows.h>
+#include <iostream>
+
+#pragma comment(linker,"\"/manifestdependency:type='win32' \
+name='Microsoft.Windows.Common-Controls' version='6.0.0.0' \
+processorArchitecture='*' publicKeyToken='6595b64144ccf1df' language='*'\"")
+
+#define NK_INCLUDE_FIXED_TYPES
+#define NK_INCLUDE_STANDARD_IO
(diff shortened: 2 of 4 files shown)


dumblob:
I'm out of time now to look at the code, but just want to mention that in my opinion the "Nuklear style" is probably less interesting than the C++ -> C conversion itself :wink:.

But either way, I'm willing to merge it once we'll take a quick look at it.

Thanks a lot!

dumblob:
@Ohjurot any further plans with this? Shall I merge it as "under construction" piece of code?

Ohjurot:
@dumblob I'm still not happy with how Windows handles the nuklear window preview after minimizing... However all the rest is working as intended. So I would be OK with mergin the PR. I might come up with a followup PR when I got a fix to this issue.

lockie:
I've had a look at it using mingw compiler and wine. Floating windows work really cool, great job!
I have a couple of suggestions though (regarding using mingw to compile the code):
1. Replace `<Windows.h>` include with `<windows.h>`. MSVC compiler is case-insensitive in this regard, while all other compilers are not.
2. Consider using regular `WinMain` instead of `wWinMain` or at least use the [_tWinMain](https://stackoverflow.com/a/4681487/1336774) trick.

dumblob:
`_tWinMain` sounds least problematic to me.

Ohjurot:
@lockie 
Thank you! 
1) `Windows.h` should be written as it is. At least with the windows sdk version that i have installed (10.0.22567.0) the file is named with a capitalized W. I'm not aware if mingw or any other compiler on windows use any special (or older) version of `Windows.h` that is lowercase. However, since MSVC doesn't really care and all the other nuklear samples are using `windows.h` with lowercase I'm going to change this for consistency. 
2) Since I'm explicitly using Unicode through the whole demo (inside `window.h`) I think using `wWinMain` is ok. You can adopt this to your entry point of choice when using the example.

dumblob:
> 2. Since I'm explicitly using Unicode through the whole demo (inside `window.h`) I think using `wWinMain` is ok. You can adopt this to your entry point of choice when using the example.

Sure, that makes sense. Could you add a comment to the `wWinMain` line hinting at the `_tWinMain` trick to ease adoption by others in the future?

I will then want to merge this in the upcoming days/weeks.

Thanks a lot @Ohjurot for the hard work and @lockie for the test!

dumblob:
Wow, this took me long. I am sorry. But it is finally merged.

Thanks a lot for waiting and all the hard work!

PR #426, opened 2022-03-14
Fix Metal rendering in sdl_renderer

kbolino:
This commit addresses an issue where the Metal render driver (default on macOS) could not handle the clipping rectangle `{-8192, -8192, 16384, 16384}` and would simply draw nothing to the screen. The solution is to ensure the cliprect coordinates are not negative, which also doesn't seem to break the other render drivers (tested OpenGL).

### DIFF
--- a/demo/sdl_renderer/nuklear_sdl_renderer.h
+++ b/demo/sdl_renderer/nuklear_sdl_renderer.h
@@ -19,6 +19,16 @@ NK_API int                  nk_sdl_handle_event(SDL_Event *evt);
 NK_API void                 nk_sdl_render(enum nk_anti_aliasing);
 NK_API void                 nk_sdl_shutdown(void);
 
+#if SDL_COMPILEDVERSION <= SDL_VERSIONNUM(2, 0, 22)
+/* Metal API does not support cliprects with negative coordinates or large
+ * dimensions. The issue is fixed in SDL2 with version 2.0.22 but until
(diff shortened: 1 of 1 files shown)


kbolino:
This is related to [this discussion](https://discourse.libsdl.org/t/rendergeometryraw-producing-different-results-in-metal-vs-opengl/34953/4) where it's suggested that SDL2 might actually need to be fixed. So maybe hold off on merging this PR in favor of that change?

dumblob:
If this happens with a release version of SDL then we want this fix - at least as comment in the source code with the reference to the discussion you linked above :wink:.

dumblob:
Speaking of which, we could detect the SDL version we're compiling against in compile time and just use ifdef :wink:. Thoughts?

kbolino:
There's only two versions affected, but that's because there's only two versions new enough to support `SDL_RenderGeometryRaw`. Could `#ifdef` the check or could call `SDL_GetVersion` to handle cases where the dylib/so/DLL is updated without recompiling.

dumblob:
>which would handle cases where the dylib/so/DLL is updated without recompiling.

It's a hot path so this doesn't sound like a good solution. Maybe `#ifdef` but additionally also `SDL_GetVersion()` during init (i.e. **outside** of the loop) checking only for the two affected versions and aborting the runtime with a message that these two affected SDL versions are only supported by recompilation due to a bug and link to the bug.

What do you think?

PR #430, opened 2022-03-16
Delete extra copy of stb_image.h

crazyBaboon:
This file was left behind and is not used by any source code in this repo.


dumblob:
Thanks!

Issue #433, opened 2022-03-20
Build fail in C++ project [`error: invalid conversion from ‘void*’ to ‘nk_glfw*’`]


tree786isback:
I'm including nuklear in my C++ project (uses glfw and opengl3), however it fails to compile giving this:
```
nuklear_glfw_gl3.h:353:32: error: invalid conversion from ‘void*’ to ‘nk_glfw*’ [-fpermissive]
  353 |     struct nk_glfw* glfw = usr.ptr;
      |                            ~~~~^~~
      |                                |
      |                                void*

```
I tried statically casting to a `nk_glfw*`, like this:
```
struct nk_glfw* glfw = (nk_glfw*)usr.ptr;
```
and although it did compile, it segfaults when I try and use the copy/paste function in text edit (see issue #431 ). I've tried setting the C standard to 99 (can't go below that), but it didn't help. 
I assume the compiler is treating the nuklear's C code as C++, hence it hits these errors.
Any ideas how I can solve this?

tree786isback:
Here is the test files if you want to give it  a try:
[issue.zip](https://github.com/Immediate-Mode-UI/Nuklear/files/8310424/issue.zip)

tree786isback:
I'll close as #434 has prob fixed this

PR #434, opened 2022-03-20
glfw_gl3: Cast the void* to a nk_glfw*

RobLoach:
While it doesn't fix the issue @tree786isback, it is a good thing to cast the void*'s correctly.

References https://github.com/Immediate-Mode-UI/Nuklear/issues/433


tree786isback:
Quite strange how it segfaults though....I've run separate tests on `glfwGetClipboardString` (and also set clipboard) in isolation (i.e. excluding nuklear), and they work fine independently. Not sure what is causing them to crash when used via nuklear?

PR #435, opened 2022-03-23
Simplify the pre-processing ladder for boolean definition

crazyBaboon:


dumblob:
Hm, actually the original way is there on purpose - to not bury the typedef into ifdefs. I think we should not change this.

PR #436, opened 2022-03-26
Further improve file_browser.c

crazyBaboon:
-Adds visible boundary to file icon box;
-Adds a 'cancel' button that closes the file dialog;
-Adds a 'save' button that should be used for the save() routine defined by the user. 
Of course, the user can also easily change the 'save' to 'open'.

### DIFF
--- a/demo/common/file_browser.c
+++ b/demo/common/file_browser.c
@@ -1,3 +1,5 @@
+nk_bool file_browser_is_open;
+
 struct icons {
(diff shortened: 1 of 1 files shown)


crazyBaboon:
Before:
FileBrowser1

After:
FileBrowser2
:

RobLoach:
Thanks so much for these updates. The Filebrowser is starting to actually be usable haha.

PR #445, opened 2022-04-13
Ignore case when sorting strings alphabetically

crazyBaboon:


dumblob:
Thanks!

It almost seems you're using the filebrowser demo for some serious work if such (important) details come up :wink:.

PR #446, opened 2022-04-13
file_browser.c - remove #include <string.h>

crazyBaboon:
It is no longer necessary, since string comparison is now carried out by Nuklear API.

### DIFF
--- a/demo/common/file_browser.c
+++ b/demo/common/file_browser.c
@@ -1,3 +1,12 @@
+#ifdef __unix__
+#include <dirent.h>
+#include <unistd.h>
+#endif
+
+#ifndef _WIN32
+# include <pwd.h>
+#endif
+
 struct icons {
     struct nk_image desktop;
     struct nk_image home;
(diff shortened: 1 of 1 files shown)


crazyBaboon:
This is ready to merge IMHO

RobLoach:
I see use of `strcpy()` and `strlen()` which both are defined in `string.h`. Are these defined elsewhere?

crazyBaboon:
@RobLoach 

String.h was already being called in stb_image.h

demo/common/file_browser.c
```diff
@@ -1,3 +1,12 @@
+#ifdef __unix__
+#include <dirent.h>
+#include <unistd.h>
+#endif
+
+#ifndef _WIN32
+# include <pwd.h>
+#endif
+
```
RobLoach, line 9:
``` suggestion

#include <string.h>

```

demo/common/file_browser.c
```diff
@@ -1,3 +1,12 @@
+#ifdef __unix__
+#include <dirent.h>
+#include <unistd.h>
+#endif
+
+#ifndef _WIN32
+# include <pwd.h>
+#endif
```
RobLoach, line 8:
``` suggestion
#endif

#include <string.h>
```

RobLoach:
It's likely best to keep string.h in there if it's making use of the functions. We could add comments to indicate which functions are used to be clear.

PR #447, opened 2022-04-15
Horizontal rule

FrostKiwi:
grafik
This widget draws a horizontal rule line, like the classic html rule below this sentence.

---
This allows one to subdivide spaces a bit nicer. The last parameter of the widget is whether or not to apply rounding. If set to true, it will apply rounding, so the line always appears to be rounded off, the "pill shape".
The height is automatically determined by the current row height.
I didn't want to create a rule.c just for this, so I included it at the end of nuklear_window.c and the definition+documentation in Window section of nuklear.h.


crazyBaboon:
Thanks @FrostKiwi ! This is looking good

RobLoach (approved):
I like this idea! Can provide some nice separation across widgets. Think we should make use of it in any of the example projects?

RobLoach:
Actually, could we make use of the horizontal bar in one of the demos?

yukyduky:
Could we get this merged? I can add an example to the demo if nobody wants to do it.

FrostKiwi:
> Actually, could we make use of the horizontal bar in one of the demos?

Adding example to `OVERVIEW`

FrostKiwi:
@RobLoach @yukyduky 
Added a showcase to the `Overview` kitchen-sink demo.
image

RobLoach:
This is fantastic. Thanks so much, all.

PR #448, opened 2022-04-18
Button behavior `NK_BUTTON_TRIGGER_ON_RELEASE`

### LINKED ISSUE #357: Incorrect button behavior with '#define NK_BUTTON_TRIGGER_ON_RELEASE'?
When the *NK_BUTTON_TRIGGER_ON_RELEASE* compile time flag is set, buttons get triggered (return true) when the user presses the left mouse button (*LMB*) **over** the button's rectangle/area (*RA*) and then releases it **over** the button's *RA*.

However buttons also get triggered when the user presses the *LMB* **outside** of the button's *RA* and releases it **over** the button's *RA*. This behavior seems incorrect.

tcdude:
This modifies the trigger behavior of buttons to only activate when the
mouse was also over the same button on the mouse button down event.

This fixes #357 to the best of my knowledge

Side note:
I also modified my version of Nuklear I use to apply the same logic for
hover, because I use Nuklear for touch input devices where traditional
hover behavior tends to be misleading. I left this out for now because
I can imagine that this could be seen as wrong/faulty behavior by some.

### DIFF
--- a/nuklear.h
+++ b/nuklear.h
@@ -4654,6 +4654,9 @@ struct nk_mouse_button {
 struct nk_mouse {
     struct nk_mouse_button buttons[NK_BUTTON_MAX];
     struct nk_vec2 pos;
+#ifdef NK_BUTTON_TRIGGER_ON_RELEASE
+    struct nk_vec2 down_pos;
+#endif
     struct nk_vec2 prev;
     struct nk_vec2 delta;
     struct nk_vec2 scroll_delta;
@@ -17881,6 +17884,13 @@ nk_input_button(struct nk_context *ctx, enum nk_buttons id, int x, int y, nk_boo
     btn->clicked_pos.y = (float)y;
     btn->down = down;
     btn->clicked++;
+#ifdef NK_BUTTON_TRIGGER_ON_RELEASE
+    if (down == 1 && id == NK_BUTTON_LEFT)
+    {
+        in->mouse.down_pos.x = btn->clicked_pos.x;
+        in->mouse.down_pos.y = btn->clicked_pos.y;
+    }
+#endif
 }
 NK_API void
 nk_input_scroll(struct nk_context *ctx, struct nk_vec2 val)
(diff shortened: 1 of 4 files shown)


RobLoach (approved):
Confirmed! Thanks a lot for the fix.

dumblob:
Thanks for the fix! Just to clarify, the desired behavior is as follows:

1. LMB down over button B0, then move pointer **inside** of B0, then LMB release results in a click on B0

2. LMB down over button B0, then move pointer **outside** of B0 to the inside of another button (B1), then LMB release results in a click on B0

3. LMB down outside of any button, then move pointer **inside** of some button (B0), then LMB release results in doing nothing

Does it behave like this (I didn't test it - I'm not at a PC now)?

tcdude:
I'm sorry, it only does 1 and 3 of the cases listed. Personally I fail to see how 2 would be desired behavior though. Were those three desired behaviors by any chance documented somewhere? (asking for future reference to not create more work and/or unwanted behavior) 

At least from the user feedback I gathered where I use Nuklear for UI, behavior number 2 is actually undesired behavior.

I would be open to make another PR to fix this, if it is indeed wanted behavior.

dumblob:
The point (2) is nonsense - I'm sorry for that. What I had in mind was this:

2. LMB down over button B0, then move pointer **outside** of B0 to the inside of another button (B1) and then move pointer back to the **inside** of B0, then LMB release results in a click on B0

Does that work?

tcdude:
Yes, that works. LMB down over B0, move anywhere (including inside another button) and finally move inside B0 again and release results in click on B0

dumblob:
Perfect. Thanks and sorry for the fuss once again!

lockie:
Looks like this introduces regression in window moving behaviour, since `nk_input_has_mouse_click_in_rect` is used in a lot of other places. The problem is that when moving a window using its header, at some point (seemingly when the cursor lefts the bounds of initial position of window header) window movement stops, while the cursor is still being moved with left mouse button still held. Here's a video to illustrate: https://imgur.com/a/AhmHeiy

tcdude:
I indeed failed to follow all indirect calls that end up in `nk_input_has_mouse_click_in_rect` and I created a separate [PR](https://github.com/Immediate-Mode-UI/Nuklear/pull/463) to fix this and restore the expected behavior.

Issue #146, opened 2020-06-05
Window rounding


Janrupf:
Currently, settings `style->window.rounding` does not seem to have any effect, and I also couldn't find any use of this variable in the source code. Am I missing something here is or is window rounding not implemented?

riri:
Did you try
```c
nk_style_push_float(ctx, &ctx->style.window.rounding, window_rounding_flow_value);
```

Janrupf:
I basicially did `context->style.window.rounding = 50f`

riri:
Actually you are right, the panel creation does not use the rounding setting, but in any case, I think changing style setting implies using `nk_style_push_***()` functions so that changed are stacked.

Janrupf:
Might make a PR for that then, sounds like someone put it there and then forgot about it

riri:
I had a look in the panel code to find out if it was applied. Then I realized it's not as simple as it seems because of clipping I guess. The way a window is drawn might need to change because of the presence of the header (for now, the window "content" is drawn below the header, so that rounding would not work as expected). I leave that to nuklear expert ;-)

Janrupf:
I think rounding the content would be fine. I also looked at the panel code, but currently I don't think there is a way to just round one corner of a rectangle, so I totally agree that rounding would look weird. It would be cool to still have rounding for just the content nevertheless, I actually don't need the titlebar and border for the rounded window

riri:
The problem is that the PR would need to handle the header.
For your exact case, I suppose using the rounding style at `nuklear_panel.c(297)` would be enough:
```c
        else nk_fill_rect(out, body, style->window.rounding, style->window.fixed_background.data.color);
```

PR #450, opened 2022-04-20
New internal rounding handling, window rounding, nk_vec4, group_rounding style setting

FrostKiwi:
grafik

This PR changes how rounding is handled internally to allow per-corner rounding to be specified in the api. As the branch name may imply, this is in preperation for quite a chunky PR introducing a bunch new elements, which depend on this feature. I decided to split it here to keep changes a bit more organized.

Now rounding is specified by nk_vec4, which is essentially the same as an nk_rect. That's the reason this PR touches so many files. The user-facing styling remains a single rounding float, so on that front nothing changes.

This unlocks window and group rounding to properly applied in regards to Titlebar and borders. Group rounding received it's own style property. Though some extra tweaks may be required to address how group borders ar not the same thickness by default on all sides. Not every combination of Window features is clean though, the scroll bar may need to be patched to inherit one corner's window rounding, as the scrollbar pokes through in certain feature combinations like: No titlebar + scroll bar + no rounding on scrollbar.
I also applied rounding 

https://github.com/Immediate-Mode-UI/Nuklear/issues/146 should be properly addressed with this.

Finally, I did try making a rounded scaler corner, but it was really just a dirt hack. This will be an extra PR later on.
grafik

### DIFF
--- a/nuklear.h
+++ b/nuklear.h
@@ -476,6 +476,8 @@ struct nk_color {nk_byte r,g,b,a;};
 struct nk_colorf {float r,g,b,a;};
 struct nk_vec2 {float x,y;};
 struct nk_vec2i {short x, y;};
+struct nk_vec4 {float a,b,c,d;};
+struct nk_vec4i {short a,b,c,d;};
 struct nk_rect {float x,y,w,h;};
 struct nk_recti {short x,y,w,h;};
 typedef char nk_glyph[NK_UTF_SIZE];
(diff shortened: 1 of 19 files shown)


RobLoach:
This makes my life a lot easier, thanks for taking it on!

crazyBaboon:
This feat looks good - we need more of these! Thanks!

FrostKiwi:
If rounding is set too strong by the user, it breaks rendering:
grafik
Now there is the nuklear_internal.h function `nk_clamp_rounding()`, which ensures this doesn't happen by clamping the rounding of each corner to the maximum permitted by the widget's dimensions.
grafik

FrostKiwi:
> Conflicts. Like the look of this though.

I'll take a look at it now.

Should be uncomplicated, it was just allowing per corner rounding and some clamping operations

PR #453, opened 2022-04-23
New folder structure -  Eliminates duplicate files

crazyBaboon:
No more duplicate files - easier maintenance and smaller repository size


crazyBaboon:
This PR fails the checks because the folder structure is different so the CI cannot find the files to compile.

image

RobLoach (changes requested):
This is a good move, though much of the demo file structure has moved around already. Mind fixing the conflicts?

PR #454, opened 2022-04-25
Display the path of the selected file in CL

crazyBaboon:
For convenience only, display the path in the terminal of the file clicked by the user. 
At least this helps me remembering which variable holds the file path


crazyBaboon:
This change also makes the demo more interactive

PR #459, opened 2022-05-12
Added working filled and outlined arc drawing to GDI binding

bmuessig:
This PR is based on #385 and correctly implements (filled) arc drawing, as well as circle stroke drawing in GDI.
It also adds the select all keyboard shortcut.

### DIFF
--- a/demo/gdi/nuklear_gdi.h
+++ b/demo/gdi/nuklear_gdi.h
@@ -15,6 +15,7 @@
 
 #define WIN32_LEAN_AND_MEAN
 #include <windows.h>
+#include <wingdi.h>
 
 typedef struct GdiFont GdiFont;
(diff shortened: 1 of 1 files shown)


bmuessig:
grafik

bmuessig:
The issue in the original PR was that the drawing commands use angles in radians, not degrees. The conversion caused incorrect coordinates to be fed to GDI.

bmuessig:
I've just realized that there is one more glitch - probably introduced when adding the circle fix:

grafik

The corner dragging triangle is no longer filled black, when the circle is visible.

grafik

bmuessig:
Filling now works correctly:

grafik

RobLoach:
Thanks so much for fixing it up!

bmuessig:
Thanks for merging!

PR #460, opened 2022-05-12
Some improvements to alpha blending in GDI

bmuessig:
This does still not work, but it corrects a few bugs that would additionally hinder it from working.
When moving the cursor over the gradient, it does now flicker and "appear" to display a gradient (probably just a redraw effect).
At this point, I am not sure which component is not properly supporting the alpha channel and causing the gradient to not display properly. From what I gathered searching, it is probably 37-year-old GDI itself.

### DIFF
--- a/demo/gdi/nuklear_gdi.h
+++ b/demo/gdi/nuklear_gdi.h
@@ -221,7 +221,7 @@ nk_gdi_set_vertexColor(PTRIVERTEX tri, struct nk_color col)
     tri->Red   = col.r * (0xffff / 0xff);
     tri->Green = col.g * (0xffff / 0xff);
     tri->Blue  = col.b * (0xffff / 0xff);
-    tri->Alpha = 0xffff;
+    tri->Alpha = col.a * (0xffff / 0xff);
 }
 
 static void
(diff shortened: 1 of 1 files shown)


demo/gdi/nuklear_gdi.h
```diff
@@ -218,27 +218,26 @@ nk_gdi_fill_rect(HDC dc, short x, short y, unsigned short w,
...
+    tri->Blue  = col.b * (0xffff / 0xff);
+    tri->Alpha = col.a * (0xffff / 0xff);
 }
 
 static void
 nk_gdi_rect_multi_color(HDC dc, short x, short y, unsigned short w,
     unsigned short h, struct nk_color left, struct nk_color top,
     struct nk_color right, struct nk_color bottom)
 {
+    /* Note that GDI (without plus) does not appear to support alpha transparency properly in gradients. */
```
dumblob, line 232: Could you also add one or two words about how it manifests itself visually (I mean the "flickering")? Thanks!

bmuessig: Sure, when hovering and moving the mouse over the color picker matrix, the displayed gradient flickers to black.
I've prepared a short video: https://www.youtube.com/watch?v=oaEzvqKD_fo

PR #466, opened 2022-06-12
New image handling and image types

FrostKiwi:
So far Nuklear has ignored image aspect ratio and getting images to properly display has been by hacking your way around the layout system. This PR introduces `enum nk_image_type` in `struct nk_image` to remedy this and obsoletes PR https://github.com/Immediate-Mode-UI/Nuklear/pull/444. The new image types are demonstrated here:

https://user-images.githubusercontent.com/60887273/173217310-f43d82ce-a71c-4e84-b7c3-5809d599c3ed.mp4

The image type is a setting defined upon image creation and does not need to be supplied each time nk_image is called. The API for *displaying* the image stays the same. nk_image() generates the `NK_COMMAND_IMAGE` with correct sizes to honor aspect ratio based on the surround layout.
- **NK_IMAGE_STRETCH**: The default stretching, as required for styling to function properly
- **NK_IMAGE_FILL**: Honors the image's aspect ratio, fills out the entire target rect
- **NK_IMAGE_FIT**: Honors the image's aspect ratio, fits the entire image into the target rect
- **NK_IMAGE_CENTER**: 1:1 pixel-mapping, center the image
- **NK_IMAGE_TILE**: 1:1 pixel-mapping, sample image outside of UV range. It is required that the backend is set to texture tiling mode. (Which is the default OpenGL Texture behaviour anyways.). This will *not* work with nk_subimage in this PR's current form, as subimage just modifies UVs. This really has to be addressed, but it cannot happen inside of the backend agnostic `nuklear.h`. I think providing a utility function in `nuklear_util.c` to copy a the subimage rect into a new buffer, so the user may upload that buffer to create a new texture.

~~This PR is an API break: It requires the user to provide image-size upon image creation, no exceptions.~~ Also the image type has to be provided upon nk_image creation. This is required for any of this to work and is a long overdue change IMO. This PR is also essential for the interface element that I'm porting to Nuklear "cards" to work. ~~Since this API break is needed, I took the chance to rename some functions for more clarity. Eg. `struct nk_image nk_image_ptr(void*);` is now `nk_image_from_ptr(void*, nk_ushort w, nk_ushort h, enum nk_image_type);`, to make it clear, that this function is for generating a new nk_image from a pointer, not for providing the pointer of an nk_image.~~

Now there is no more new command type needed, as I did in PR https://github.com/Immediate-Mode-UI/Nuklear/pull/444, which is much nicer to maintain.

TODO:
- ~~Fix demo files to speak the new API, so the Github's build pipline does not complain~~
- ~~Make nk_subimage versions of all the new image types, except NK_IMAGE_TILE~~
- Introduce a function in 'nuklear_util.c' to solve the problem of nk_subimage being incompatible with tiled image drawing. (Unless someone has a better idea...)


nuklear.h
```diff
@@ -10687,7 +10802,7 @@ nk_convert(struct nk_context *ctx, struct nk_buffer *cmds,
                 t->string, t->length, t->height, t->foreground);
         } break;
         case NK_COMMAND_IMAGE: {
-            const struct nk_command_image *i = (const struct nk_command_image*)cmd;
+            const struct nk_command_image *i = (const struct nk_command_image*)cmd;;
```
dumblob, line 10805: probably typo

nuklear.h
```diff
@@ -23475,78 +23590,117 @@ nk_handle_id(int id)
...
 }
 NK_API struct nk_image
-nk_subimage_ptr(void *ptr, nk_ushort w, nk_ushort h, struct nk_rect r)
+nk_subimage_from_ptr(void *ptr, nk_ushort w, nk_ushort h, enum nk_image_type type, struct nk_rect r)
 {
     struct nk_image s;
     nk_zero(&s, sizeof(s));
     s.handle.ptr = ptr;
+    NK_ASSERT(w); 		/* Size may not be zero */
+    NK_ASSERT(h); 		/* Size may not be zero */
```
dumblob, line 23599: I think there are use cases for zero size(s). Right now I can think of only the following one but I bet there are more.

1. An image in JPEG, PNG, etc. formats is allowed to have size 0x0 if I am not mistaken. Loading such an image **must not** crash (on assert or otherwise) the app.

dumblob:
Generally this looks like a promising approach to me.

>It requires the user to provide image-size upon image creation, no exceptions.

This though is a bit too strict in my opinion. I can not imagine myself tracking dimensions of each image, icon, ... I use in my app.

Why actually not emulate the current "old" API (which was extremely limited but did not require specifying dimensions) using this your new API?

Actually I dare to think this new approach could be implemented in a backwards compatible way (i.e. just extending the current API) without the need to introduce 5.x.y version.

What do you all think?

FrostKiwi:
> What do you all think?

Huh... Now that you mention it, maybe that is too strict. So keep the old commands, always defaulting to NK_IMAGE_STRETCH and create a set of new ones, where you have to specify size I guess?

> I can not imagine myself tracking dimensions of each image, icon

Isn't that always automatically known when you load the image?

dumblob:
> Huh... Now that you mention it, maybe that is too strict. So keep the old commands, always defaulting to NK_IMAGE_STRETCH and create a set of new ones, where you have to specify size I guess?

Yes, right now that would be my preference. But I have a weird feeling I am missing something so please come up with other ideas if you have any :wink:.

> > I can not imagine myself tracking dimensions of each image, icon
> 
> Isn't that always automatically known when you load the image?

Yes. What I meant though is the syntactical pollution of maintaining variables (scalars, arrays, maps, whatever...) holding the dimensions for no apparent benefit. Does that make more sense now?

src/nuklear_image.c
```diff
@@ -22,78 +22,117 @@ nk_handle_id(int id)
...
     struct nk_image s;
     nk_zero(&s, sizeof(s));
     s.handle = handle;
+    NK_ASSERT(w); 		/* Size may not be zero */
+    NK_ASSERT(h); 		/* Size may not be zero */
     s.w = w; s.h = h;
+    /* Because a subimage just uses modified UVs to show a small portion of an
+       nk_image, using CENTER or TILED will result in the rest of the image
+       bleeding through. */
+    NK_ASSERT(type != NK_IMAGE_CENTER && type != NK_IMAGE_TILE);
```
FrostKiwi, line 82: Ohh, totally forgot. NK_IMAGE_CENTER actually works with subimage, just TILE doesn't

FrostKiwi:
> Yes, right now that would be my preference. But I have a weird feeling I am missing something so please come up with other ideas if you have any 😉.

Created two sets of commands, one with _type_ in the name, reverted changes to old commands, removed [5.x] tag from issue titled, since there is no more API break.

nuklear.h
FrostKiwi: Fixed

nuklear.h
FrostKiwi: Good catch, always miss stuff like this.

dumblob:
Nice, thanks. Do you still plan to work on the last TODO items in the initial post above or shall we make our hands dirty with a full-scale review already?

RobLoach:
I really like where this is going!

FrostKiwi:
Note to self: Use Nuklear's internal floor(), instead of the math.h floor(). Creates an unresolved dependancy

wuxxerno:
Is there anything I can help with to get this merged?

FrostKiwi:
> Is there anything I can help with to get this merged?

I've given you access to my repo and the branch, that this PR is connected to. Go wild, in case you have any ideas.

I'll revive my Nuklear efforts over the weekend. What it got stuck on last time I wanted to contribute, are the Krita design files + nk_card code, that I couldn't publish due to being derived from company internal design guideslines.

I'll recreate them over the weekend from FOSS material, post that and then return to this PR.

wuxxerno:
I was able to start on this today. I updated the NK_IMAGE_CENTER case to use uv coords instead as it didnt work correctly on dynamic layout. I will continue with implementing the nk subimage as per @FrostKiwi todo notes.

FrostKiwi:
> as it didnt work with dynamic layouting

@wuxxerno Ohh wow, you were right. How did I miss that. Glad to have this figured out, I really didn't want to do the math there to get it to work without scissor^^

You made changed directly to [`nuklear.h`](https://github.com/Immediate-Mode-UI/Nuklear/blob/master/nuklear.h), instead of the [`src/`](https://github.com/Immediate-Mode-UI/Nuklear/tree/master/src) soruce code of nuklear. That huge [`nuklear.h`](https://github.com/Immediate-Mode-UI/Nuklear/blob/master/nuklear.h) is auto-generated and should not be changed. You should do your changes in [`src/`](https://github.com/Immediate-Mode-UI/Nuklear/tree/master/src) and either include directly from the source or build the final [`nuklear.h`](https://github.com/Immediate-Mode-UI/Nuklear/blob/master/nuklear.h) with [`src/paq.sh`](https://github.com/Immediate-Mode-UI/Nuklear/blob/master/src/paq.sh) or [`src/paq.bat`](https://github.com/Immediate-Mode-UI/Nuklear/blob/master/src/paq.bat) on windows.
For instance I test in [`demo/glfw_opengl3`](https://github.com/Immediate-Mode-UI/Nuklear/tree/master/demo/glfw_opengl3), where I have a `test.sh`, which rebuilds the final [`nuklear.h`](https://github.com/Immediate-Mode-UI/Nuklear/blob/master/nuklear.h) and compiles the test program.
```bash
cd ../../src && sh paq.sh && cd - && make && ./bin/demo.exe
```

I moved your contributions over to with https://github.com/Immediate-Mode-UI/Nuklear/pull/466/commits/fc97bb9eaf2686a7b9178baa5263fcc18776ec8b to [`src/nuklear_vertex.c`](https://github.com/Immediate-Mode-UI/Nuklear/blob/master/src/nuklear_vertex.c)

wuxxerno:
> > as it didnt work with dynamic layouting
> 
> @wuxxerno Ohh wow, you were right. How did I miss that. Glad to have this figured out, I really didn't want to do the math there to get it to work without scissor^^
> 
> You made changes directly to [`nuklear.h`](https://github.com/Immediate-Mode-UI/Nuklear/blob/master/nuklear.h), instead of the [`src/`](https://github.com/Immediate-Mode-UI/Nuklear/tree/master/src) source code of nuklear. That huge [`nuklear.h`](https://github.com/Immediate-Mode-UI/Nuklear/blob/master/nuklear.h) is auto-generated and should not be changed. You should do your changes in [`src/`](https://github.com/Immediate-Mode-UI/Nuklear/tree/master/src) and either include directly from the source or build the final [`nuklear.h`](https://github.com/Immediate-Mode-UI/Nuklear/blob/master/nuklear.h) with [`src/paq.sh`](https://github.com/Immediate-Mode-UI/Nuklear/blob/master/src/paq.sh) or [`src/paq.bat`](https://github.com/Immediate-Mode-UI/Nuklear/blob/master/src/paq.bat) on windows. For instance I test in [`demo/glfw_opengl3`](https://github.com/Immediate-Mode-UI/Nuklear/tree/master/demo/glfw_opengl3), where I have a `test.sh`, which rebuilds the final [`nuklear.h`](https://github.com/Immediate-Mode-UI/Nuklear/blob/master/nuklear.h) and compiles the test program.
> 
> ```shell
> cd ../../src && sh paq.sh && cd - && make && ./bin/demo.exe
> ```
> 
> I moved your contributions over to with [fc97bb9](https://github.com/Immediate-Mode-UI/Nuklear/commit/fc97bb9eaf2686a7b9178baa5263fcc18776ec8b) to [`src/nuklear_vertex.c`](https://github.com/Immediate-Mode-UI/Nuklear/blob/master/src/nuklear_vertex.c)

I see! thank you for explaining and correcting that.

wuxxerno:
I just finished implementing the nk_subimage handling. (All but NK_IMAGE_TILE)
Im not really sure on how to proceed with problem of tiling subimages

> I think providing a utility function in nuklear_util.c to copy a the subimage rect into a new buffer, so the user may upload that buffer to create a new texture.

Did you mean something totally generic like a function that swaps and crops two user inputted buffers that the user then can use upload as a new texture for a new nk_image? @FrostKiwi 

@dumblob 
Im happy to work on any feedback if you would like to start reviewing.

FrostKiwi:
> I just finished implementing the nk_subimage handling. (All but NK_IMAGE_TILE)

That is HUGE! It was the quint essential challenge remaining to get this PR to an at least theoretically mergable state.

> Im not really sure on how to proceed with problem of tiling subimages

In short: there is no solution from the position of a backend agnostic library, because the tile needs tile (heh). Something only the backend can do. (Except rendering a ton of nk_images to tile window, resulting in horribly wasteful performance and blowing past all kinds of buffers on a high res screen with a small tile)

> Did you mean something totally generic like a function that swaps and crops

Yes. A function which exports the content of an nk_subimage as defined by the subregion to a buffer, specified by the user with a pointer. Similar to `snprintf()`, it returns the written size and if the user pointer is NULL, it returns just the needed size without writing anything, so the user may `malloc()` the needed amount.

The user may then setup tiling correctly. (By uploading as a separate texture, or bundle them in a texture array). It's less of a solution and more of a convenient tool to help the user setup tiling. (technically there is no reason this couldn't be done via a custom shader with `frac()` returning UVs in GLSL even without such an `nk_util.h`, but we should give at least some kind of universal solution, even if naive)

malespiaut:
Hello,

I've tried to reproduce your (@FrostKiwi) example code using the SDL2 Surface backend, and the whole context crashes when trying to display `NK_IMAGE_TILE`.

https://github.com/Immediate-Mode-UI/Nuklear/assets/76531574/a5d26ac3-2021-4e93-867d-bb7324010a7a

Here's my modified `nuklear/demo/sdl_renderer/main.c` with the needed images.

[sdl_renderer.zip](https://github.com/Immediate-Mode-UI/Nuklear/files/14896730/sdl_renderer.zip)

I hope this isn't a bug in my own implementation!

FrostKiwi:
@malespiaut 
> Hello,
> 
> I've tried to reproduce your (@FrostKiwi) example code using the SDL2 Surface backend, and the whole context crashes when trying to display `NK_IMAGE_TILE`.

This should be really working. As mentioned in the doc, `NK_IMAGE_TILE` is a special case requiring cooperation from the rendering backend:
```
    /* Just draw the image with UVs for 1:1 pixel mapping. Image is draw outside
       it's bounds. What is drawn outside the bounds is determined by the
       backend. Eg. To get tiling behaviour with OpenGL, the texture should have
       been uploaded to the GPU with GL_REPEAT:
       glTexParameteri(GL_TEXTURE_2D, GL_TEXTURE_WRAP_S, GL_REPEAT);
       glTexParameteri(GL_TEXTURE_2D, GL_TEXTURE_WRAP_T, GL_REPEAT);
       (Which is the default OpenGL Texture behaviour anyways.) */
```
I followed your example and implemented it in SDL_renderer, loaded up the texture via `SDL_Texture *texture = IMG_LoadTexture(renderer, "bin/bunny_wide.png");` and let it run in `NK_IMAGE_TILE` mode.
Same deal, the rendering disappears. As with your video, not a crash though!

Checked `SDL_GetError()` and here is the cause:
`[Error] Values of 'uv' out of bounds 0.525253 1.420455 at 320/360`
So the cause is clear. No crash, no mishandling, SDL2 (When handling the textures through its API) just refuses to draw images outside its UV bounds, which is required for the most simple implementation of tiling.

I always used GLFW for my projects and never used SDL2, so from here on out I can't say why SDL2 refuses drawing a texture outside it's bounds. Here is what ChatGPT had to say on the topic:

> SDL itself doesn't directly support repeating textures (tiling) by setting texture coordinates outside this range as you might in OpenGL or Direct3D. This is because SDL abstracts away the underlying graphics API to provide a simple, cross-platform rendering API, which doesn't expose direct control over texture wrap modes.

PS: gave you commit access to this Pull Request, in-case you want to contribute a demo or fix.

malespiaut:
> I followed your example and implemented it in SDL_renderer, loaded up the texture via `SDL_Texture *texture = IMG_LoadTexture(renderer, "bin/bunny_wide.png");` and let it run in `NK_IMAGE_TILE` mode. Same deal, the rendering disappears. Neither you video, nor my recreation resulted in a crash though, just the blue background.

There is no crash. It's just me quitting the application and starting it again to show the bug in three different way (resize, scroll, different window).

Sorry for my miscommunication.

> PS: gave you commit access to this Pull Request, in-case you want to contribute a demo or fix.

Thank you very much, I'll do that!

RobLoach:
It has been officially more than 2 years since this Pull Request was opened. Is there anything we can do to help push it forwards? What else is missing?

Reticulatas:
This PR would be improve Nuklear functionality a lot.  I pulled it locally into a fork but it did not appear to work naively.  Maybe deprecated in the time it's been sitting around?

FrostKiwi:
> This PR would be improve Nuklear functionality a lot. I pulled it locally into a fork but it did not appear to work naively. Maybe deprecated in the time it's been sitting around?

I'll re-investigate now and get a demo up and running for one backend. Then let's see if we can do a platform agnostic demo for all the backends.

FrostKiwi:
I added the demo as you [see in the video in the OP](https://github.com/Immediate-Mode-UI/Nuklear/pull/466#issue-1268444818) to the demo `glfw_opengl3`. Running the make file builds it. If you run it from the same place as the make file `./bin/demo` it loads the images correctly and displays as shown in the video. @Reticulatas You should be able to build it from this PR and run it. Have not integrated the demo into other backends yet, gotta clear up the points below first.

Please note, the demo shows how to use all the modes with *non-subimage* `nk_image`. It doesn't include the @wuxxerno contributed subimage usage. Can include if wished for, thought I keep the demo simple.

There are 3 fundamental issues with, that are not new in this PR, but are a problem in general:

## Issue 1: STB_IMAGE
- Any image loading done for demos so far is done with [stb_image](https://github.com/nothings/stb#stb_libs), of which we have two. in [`/example`](https://github.com/Immediate-Mode-UI/Nuklear/tree/master/example) and [`/demo/common/filebrowser`](https://github.com/Immediate-Mode-UI/Nuklear/tree/master/demo/common/filebrowser)
  - It ***doesn't*** build with C89, which is why I did the temporary and dirty switch to C99 here: https://github.com/Immediate-Mode-UI/Nuklear/pull/466/files#diff-cbfb75ad39cfa4ad635561f5a46741a209cf79479e404a8f31ed11f98007b42f
  - I think stb_image can build with `C89` and it's just the `//` comment style.
  - If another `/demo/common` place will use `stb_image.h` we have to add a guard that makes `#define STB_IMAGE_IMPLEMENTATION` unique
  - @RobLoach how do we want to resolve this? Combine both `stb_image.h` into one, strip out all comments and have a C89 `stb_image.h` in the repo?
## Issue 2: Image paths are execution path dependent
How the images are loaded depends on where we execute from. Not sure how we want to handle this. Could binary encode small JPEG test images, just like we do with the default font file. What's the thinking here?
## Issue 3: Not backend agnostic
The demo file is here: [`/demo/common/image.c`](https://github.com/Immediate-Mode-UI/Nuklear/blob/e72bc2f86a4dd47b4a1ae35e5da0949f5beeecc5/demo/common/image.c). It's not pretty and assumes a lot of things (Won't work as-is in Direct or as a WebApp in WASM). Somehow we gotta bend it to be backend agnostic and I don't know how. The backend needs to prepare the image and this is fundamentally not something that should be in [`/demo/common/image.c`](https://github.com/Immediate-Mode-UI/Nuklear/blob/e72bc2f86a4dd47b4a1ae35e5da0949f5beeecc5/demo/common/image.c) specifically `loadTextureOpenGL` has to be moved. For DirectX and OpenGL this means creating a texture. For tiling to work, edge-repeat has to be enabled in the backend. (true by default in OpenGL).

Any suggestions? Move the creation of the `nk_image` to abstract it aways and program `OutPut_nk_image("inputPath")` a bunch of times in `nuklear_glfw_gl4.h`, `nuklear_d3d11.h` etc?

FrostKiwi:
I addressed point Issue 1 with the following:
- Removed the double `stb_image.h`
- Inserted with the newest `v2.30` in `demo/common/libs`
- removed all C++ Style comments and added a message
  - with `sed 's,//.*,,g' stb_image.h > stb_image_c89.h`
  - `This is STB_IMAGE.h with all // style comments removed to be able compile as C89 with sed 's,//.*,,g' stb_image.h > stb_image_c89.h`

Replaced all those instances, where we broke with Nuklear's ideals of being ANSI C by replacing `-std=c89` in `/demos` using the filebrowser and all `/examples`.

Made sure all the examples compile.

<details><summary>Screenshots of examples and demos working with images, as they did previously.</summary>

image
image

</details>

rswinkle:
First this is a very cool addition and I can't wait for it to get in.  Hopefully I can contribute a bit.

> Issue 3: Not backend agnostic

I've been looking at the code and while I'm not 100% on it I think I roughly understand the issue and can see two solutions.

1. Vertices are pretty cheap.  We could modify nk_push_rect_uv() to check for out of bounds UV's and just generate vertices in a grid (yes with duplication) to keep UV's in [0, 1].  I'm not sure how I feel about this but it could definitely be done.
2. Have a util function like you said but go further.  Something like this (very rough):

```
int nk_tile_image(nk_byte* img, nk_ivec2 dims, nk_vec2 uv, nk_byte** outimg, nk_ivec2 out_dims)
```

Basically have the user give the maximum UV (assuming starting at 0...there's no reason we need to handle negative UV's when they can just shift the starting point), generate an output image for them.  We could even support mirrored as well as regular tiling without too much extra work.  Then they give that image to the regular nk_image with NK_IMAGE_CENTER or something.  I'm a bit vague on how this would interact with the rest of the API  but I'm sure it wouldn't be unbearably awkward.

I would be happy to write that function if we can come up with the right API or even work on the first solution if I had access to that branch.

On a related note I was just looking at SDL3 and they added [Tiling](https://wiki.libsdl.org/SDL3/SDL_RenderTextureTiled) among other things.

FrostKiwi:
> Issue 3: Not backend agnostic

The main issue is still, that to do an image feature demo, we need to load images. This is different on DirectX than on OpenGL and I'm not sure how to do this, except write the same demo 15 times. I have written it for just `glfw_opengl3`. Maybe that's already enough to merge? 🤔

> Basically have the user give the maximum UV (assuming starting at 0...there's no reason we need to handle negative UV's when they can just shift the starting point), generate an output image for them.

Sounds great.
"Generate an output image" is exactly the thing missing to get sub-image + tiling to work as well, the one feature combo of this PR, that categorically doesn't work (Because sub-image is just a UV modification and tiling expands the what we see, undoing sub-image). 👍

> I would be happy to write that function if we can come up with the right API or even work on the first solution if I had access to that branch.

Please go ahead, you have commit permissions to this branch / PR.

RobLoach:
Thanks for picking this up. Having better image handling would greatly benefit us here.

As for image loading, we could possibly abstract it a bit a have an optional image example in the Overview Demo that only is used when image data is passed in. That way the image handling demo would use the same code and it would only be image loading that would differ.

rswinkle:
So it turns out we have another issue that requires a decent amount of work.

What I've done is enough to allow any backend that uses NK_INCLUDE_VERTEX_BUFFER_OUTPUT/nk_convert() to work correctly.  However, in the process of trying to get vanilla X11 to work I discovered that all that work you do in `nk_draw_list_add_image()` really should be in `nk_draw_image()` since that's the function that will be called regardless of whether you're using nk_convert() or not.  So rather than calculating UVs in float it should really be integer offsets and dimensions and we might need to change/add to the structures (nk_image or nk_command_image)  to pass that information along.  This would of course greatly simplify `nk_draw_list_add_image()` since all the hard work would already be done and you'd just have to convert it to UV's from raw pixel offsets.

Basically backends that use functions like [this](https://tronche.com/gui/x/xlib/graphics/XPutImage.html) don't get the information they need to from the draw image command to draw what we want them to draw, rather than just blitting the whole texture to the bounding rectangle.

I'm happy to keep going and try my hand at it but just wanted to keep everyone in the loop.

Also here is a screenshot of sdl_renderer backend working with my tiling-by-creating-a-tiled-image method

SDL_Renderer_working

And here's what the X11 backend looks like...ironically the "fake" tiling is the only one that looks right because all it does is fill the available space 1:1 regardless.

X11_backend_not_working

Speaking of which maybe that should be another image type.  Other than tiling there's no way to tell it to just render 1:1 with 0,0 in the top left, but there's no way to tell in not to repeat and I don't think there's a way to solve that with subimages either.

rswinkle:
So I'm running into some issues and I don't want to just start making executive decisions that I'll just have to undo later.
I'm sorry for the wall of text.

My approach to the problem above was to just adapt the code in `draw_list_add_image()` for `draw_image()` and since we really need both source and dest rect information and we're already using the x,y, w, h in nk_command_image as the dest I would just use the region[] array in image to be the source rect info.

As an aside, why is there no consistency in the types?  Why are we using free floating short and unsigned shorts in nk_command_image instead of just an nk_recti?  Why are we using `nk_ushort region[4]` instead of an nk_recti in nk_image?  And we pass in an nk_rect to subimage functions and cast to shorts.  As long as we're going to have to change some things I think we should just use rects everywhere they're appropriate, it makes it so much easier to read.

The new issue I just ran into that I should have realized sooner, is that `nk_draw_image()` takes the nk_image as a const so unless that changes I can't even modify the region array.  I'm not sure why it's const since it's passed by value to `nk_image()` so it's not like we'd be changing it out from under the user.

There is another issue related to my last comment.  Some backends like X11 not only don't support vertices/texcoords, but they don't even support scaling images, as show in the screenshot above and the XPutImage() function which only takes a single width/height pair because the output is always 1:1.  That breaks 3 out of our 5 image types and tiling is already using my pseudo-tiling workaround so not sure if I should count that.  I wouldn't be surprised if several of the other platforms are similarly limited either inherently or as the backend is currently written.

At some point it might be logical to just say that this feature is not universally supported.  If we limited it to just vertex mode we'd be basically done aside from minor polish/naming issues and then documentation.

I solved the image loading issue by just using stb_image but letting them override it if they want (not sure why you'd want to).  That function just returns an nk_image holding the pixel data in RGBA and width and height.  They also need a function to create a proper nk_image for their backend from the nk_image returned by the loading function.  Since OpenGL is a pretty common denominator it makes sense to have that be included so they don't have to do anything except tell us they're using OpenGL.  It's not super pretty but it works and the demos are more about showing how you can do things not how you have to so I think it serves the purpose.

Oh and I made a 4th window of subimages to make sure I was understanding them and that I didn't break them in my ongoing experiments.  I'll push that right now but that's a lot of windows so we should probably get rid of one; I vote for the portrait one since I use that for the subimage window.

StrikerTheHedgefox:
Definitely interested in this if you're still willing to work on it.

RobLoach:
I did another test of this on SDL2, and it worked quite well. What is needed? From my understanding...

1. Ensure it works across all other larger renderers
2. Use nk_recti instead for the regions
3. Remove `const` constraints
4. Clean up the casting across nk_recti and nk_rect. It gets hairy.
5. Fix X11. Add a fallback for software renderer
6. Make sure `stb_image` is clear as an optional helper library?

PR #468, opened 2022-06-17
Feature/checkbox indicator

RobertLemmens:
This is basically https://github.com/vurtun/nuklear/pull/653 from zhouxs1023 so all credit there. Followed the discussion #222 and this is something I also need so ported the change from zhouxs1023 to here because I have some time today.

tested it out and looks good to me.
check_and_radio

### DIFF
--- a/nuklear.h
+++ b/nuklear.h
@@ -24434,7 +24434,7 @@ nk_draw_checkbox(struct nk_command_buffer *out,
 
     /* draw background and cursor */
     if (background->type == NK_STYLE_ITEM_COLOR) {
-      nk_stroke_rect(out, *selector, 2, 2, background->data.color);
+        nk_stroke_rect(out, *selector, 2, 2, background->data.color);
     } else nk_draw_image(out, *selector, &background->data.image, nk_white);
     if (active) {
         if (cursor->type == NK_STYLE_ITEM_IMAGE)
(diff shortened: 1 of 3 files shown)


dumblob:
I wonder how is it with radios? Both shall behave visually "the same".

Also thinking about this I am not sure whether the originating issue lies simply within a reversed boolean coming from the checkbox/radio widgets. If we reversed the boolean meaning then I think the current style in `master` branch would be OK, right?

RobertLemmens:
Yes radio also looks the same, here is a selected radio:
nuklear_radio

I honestly think that this is much better compared to the old styling (especially radio):
check_old


It is much more clear that it is selected because of the visible border around the inset fill,  like there is a "container", and there's something visibly inside of it (selected). The solid colors in the old one looks like empty boxes (while they are actually selected).

If we take a look at other radio implementations around the web they usually all behave in this PRs way making it extremely clear to the user what is going on:
radio_example
radio_example2
radio_example3

when it is a fill color it is usually paired with some other visual indicator, like a dot or a checkmark like here:
fill_example

Checkbox is more often done with a real checkmark but I think we should just keep it in sync with radio instead of adding a real checkmark. User can add image for this for now.

RobertLemmens:
I do think however that the background also impacts this allot!

RobLoach:
I like this change, but it looks a bit off-center on some of the other demos... Could likely be something else that's affecting that :thinking: 

Screenshot at 2022-07-09 13-30-10

riri:
I should check that one too, I really prefer the PR style of checkboxes and radios on default nk style.

PR #469, opened 2022-06-26
Rename null texture field to name friendlier to language bindings

ryuukk:
"null" is a reserved keyword in most languages, to make the usage of the library easier for other languages (and other compilers zig/D), let's rename it, it also is a better name in my opinion, null should be reserved in C too imo

### DIFF
--- a/src/nuklear.h
+++ b/src/nuklear.h
@@ -906,7 +906,7 @@ NK_API void nk_input_end(struct nk_context*);
 /// cfg.curve_segment_count = 22;
 /// cfg.arc_segment_count = 22;
 /// cfg.global_alpha = 1.0f;
-/// cfg.null = dev->null;
+/// cfg.tex_null = dev->tex_null;
 /// //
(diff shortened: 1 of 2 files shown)


dumblob:
Looks OK to me on the first glance.

Could someone else please review this? Maybe also @Hejsil as a Zig expert :wink:?

RobLoach (approved):
This makes more sense to me. `null` as a variable name is pretty problematic. Glad to get some context here.

Hejsil:
As for Zig, `@"null"` can be used for variable, field, function and declaration names, so it is not really a problem as long as Zig's auto translation tools actually does the correct thing. Idk about other languages, though, so this might still be a necessary change,

dumblob:
@ryuukk ok, please bump the version (the third number only) to not break any build pipelines expecting that each release has a corresponding unique version string and then we will merge this :wink:. Thanks!

PR #480, opened 2022-07-25
Fix Shift + End in nk_edit

opsJson:
Cursor jumps to the beginning when the last character is entered to reach the maximum size of the editbox.

As it is visually at the beginning, it is possible to select all the text with shift + end key.

However, in fact, it is just a visual effect, in the internal state of the code, it is after the last character.

Thus, selection_begin, position obtained after testing the minimum value of the beginning and end of the selection, will always be the total size of the text, which is the only value excluded from the possibilities of being the initial position of the selection, so that no glyph is compared to this value.

Therefore, select_begin_ptr is not initialized, as the if block of line 27.201 is not satisfied

print

### DIFF
--- a/nuklear.h
+++ b/nuklear.h
@@ -26163,6 +26163,7 @@ nk_textedit_text(struct nk_text_edit *state, const char *text, int total_len)
             {
                 nk_textedit_makeundo_insert(state, state->cursor, 1);
                 ++state->cursor;
+				state->cursor = NK_MIN(state->cursor, state->string.len);
                 state->has_preferred_x = 0;
             }
         }
@@ -26423,7 +26424,6 @@ retry:
          if (shift_mod) {
             nk_textedit_prep_selection_at_cursor(state);
             state->cursor = state->select_end = state->string.len;
-			if (state->select_start >= state->select_end) state->select_start = 0;
             state->has_preferred_x = 0;
          } else {
             state->cursor = state->string.len;
@@ -30007,3 +30007,4 @@ nk_tooltipfv(struct nk_context *ctx, const char *fmt, va_list args)
 /// in libraries and brought me to create some of my own. Finally Apoorva Joshi
 /// for his single header file packer.
 */
+
--- a/nuklear.h
+++ b/nuklear.h
@@ -26423,6 +26423,7 @@ retry:
          if (shift_mod) {
             nk_textedit_prep_selection_at_cursor(state);
             state->cursor = state->select_end = state->string.len;
+			if (state->select_start >= state->select_end) state->select_start = 0;
             state->has_preferred_x = 0;
          } else {
             state->cursor = state->string.len;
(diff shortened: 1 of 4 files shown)


opsJson:
I managed to fix the bug of the cursor going to the beginning of the editbox when the last character is inserted, just by using an NK_MIN with the cursor position to be set, and the string size

`else edit->cursor = NK_MIN(win->edit.cursor, *len);`

instead of `else edit->cursor = win->edit.cursor;`

at line 27516.

nuklear.h
```diff
@@ -26423,6 +26423,7 @@ nk_textedit_key(struct nk_text_edit *state, enum nk_keys key, int shift_mod,
          if (shift_mod) {
             nk_textedit_prep_selection_at_cursor(state);
             state->cursor = state->select_end = state->string.len;
+			if (state->select_start >= state->select_end) state->select_start = 0;
```
RobLoach, line 26426: Would you mind adding this to nuklear_text_editor.c too? https://github.com/Immediate-Mode-UI/Nuklear/blob/6e80e2a646f35be4afc157a932f2936392ec8f74/src/nuklear_text_editor.c#L654-L658

Otherwise, we'll lose the change next time nuklear.h is updated.

opsJson: OK

I focused on solving the cause of the problem. In this last commit, I limited the cursor position to the text size prevents the visual bug of the cursor going to the beginning of the editbox, and its consequences.

RobLoach: Thanks so much! If you'd like some help, I'd be happy to make the update.

dumblob:
When at it, could you also bump the version as mentioned in [readme](https://github.com/Immediate-Mode-UI/Nuklear#reviewers-guide )?

Thank you!

nuklear.h
```diff
@@ -26163,6 +26163,7 @@ nk_textedit_text(struct nk_text_edit *state, const char *text, int total_len)
             {
                 nk_textedit_makeundo_insert(state, state->cursor, 1);
                 ++state->cursor;
+				state->cursor = NK_MIN(state->cursor, state->string.len);
```
dumblob, line 26166: Nit: indentation :wink:

src/nuklear_text_editor.c
```diff
@@ -395,6 +395,7 @@ nk_textedit_text(struct nk_text_edit *state, const char *text, int total_len)
             {
                 nk_textedit_makeundo_insert(state, state->cursor, 1);
                 ++state->cursor;
+                state->cursor = NK_MIN(state->cursor, state->string.len);
```
dumblob, line 398: Maybe
```c
state->cursor = NK_MIN(state->cursor +1, state->string.len);
```
instead of
```c
++state->cursor;
state->cursor = NK_MIN(state->cursor, state->string.len);
```
would convey the message better? IDK

I will merge this PR after we settle on something :wink:.

RobLoach: I like that shorthand. Made the change in the PR over at https://github.com/Immediate-Mode-UI/Nuklear/pull/486

opsJson: good point

RobLoach:
Merged in the other PR :+1: .... Thanks so much, @opsJson and @dumblob.

Issue #455, opened 2022-04-29
DX12 Backend runtime failure with VS 2022


Sororfortuna:
This implementation may be utilizing an obsolete DirectX12 method,
the compilation is successful, however the window appears white and then shuts after a second.

build log:
```
compilation header save succeeded; see d3d12\nuklear_d3d12_vertex_shader.h
compilation header save succeeded; see d3d12\nuklear_d3d12_pixel_shader.h
main.c
d3d12\nuklear_d3d12.h(517): warning C4244: 'function': conversion from 'SIZE_T' to 'int', possible loss of data
d3d12\nuklear_d3d12.h(523): warning C4244: 'function': conversion from 'SIZE_T' to 'int', possible loss of data
d3d12\nuklear_d3d12.h(680): warning C4002: too many arguments for function-like macro invocation 'ID3D12DescriptorHeap_GetCPUDescriptorHandleForHeapStart'
d3d12\nuklear_d3d12.h(681): warning C4002: too many arguments for function-like macro invocation 'ID3D12DescriptorHeap_GetGPUDescriptorHandleForHeapStart'
main.c(135): warning C4002: too many arguments for function-like macro invocation 'ID3D12DescriptorHeap_GetCPUDescriptorHandleForHeapStart'
d3d12\main.c(136) : warning C4700: uninitialized local variable 'descriptor_handle' used
```
I don't know what is the cause.

Ohjurot:
@Sororfortuna `ID3D12DescriptorHeap_GetCPUDescriptorHandleForHeapStart` required 2 arguments during the development of the backend (Never inteded or documented by MSFT). Seems that this issue in the C-Api is gone by now. Still having the same issue like you described when running the application (Also around the call to `ID3D12DescriptorHeap_GetCPUDescriptorHandleForHeapStart`). Need to do more debugging on why this no longer works with the WinSDK...

Ohjurot:
@Sororfortuna can you tell me wich windows sdk version you are using? There seems to be a bug in the D3D12 binding on the SDK versions before 10.0.22000.0 - I used the new SDK on my inital development. From the errors that you get, it seems that you use the old version of the SDK... For me with the new SDK version the application works fine. Using any older SDK and I get the same compiler warning and runtime behaviour.

saidwho12:
[...]

PR #487, opened 2022-08-04
Fixing #455

Ohjurot:
Current working draft on the process of fixing issue #455 
Issues with `ID3D12DescriptorHeap_GetCPUDescriptorHandleForHeapStart` and a crashing application in D3D12 Backend.


Ohjurot:
So here we go... tried to fix a bug that was not a bug... found another one and fixed it (Changes to D3D12 regarding resource states on upload heaps)... Added a note for the first bug...
@Sororfortuna can you maybe check and vertify that the issue can be solve as discussed in #455

Ohjurot:
@dumblob can you please take a look at this

demo/d3d12/main.c
```diff
@@ -1,4 +1,17 @@
 /* nuklear - 1.32.0 - public domain */
+
+/*
+ *   === IMPORTATE NOTE FOR D3D12 ===  
+ *
+ * Due to a bug in the Windows SDK that
```
dumblob, line 6: Please define/describe shortly the bug in the comment so that it is `grep`pable. Then I will merge. Thanks a lot!

dumblob: Oh and maybe squash the commits - it is a small change we could keep together :wink:.

PR #492, opened 2022-09-02
Enclose quotations

JohnAlexCO:
Minor correction to the <meta> tag on the redirect page

### DIFF
--- a/doc/nuklear.html
+++ b/doc/nuklear.html
@@ -3,7 +3,7 @@
 -->
 <html>
     <head>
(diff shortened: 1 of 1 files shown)


dumblob:
Sorry for the delay. Actually the spec of `<meta>` (see https://developer.mozilla.org/en-US/docs/Web/HTML/Element/meta#attr-http-equiv paragraph with `refresh`) specifies the format which is being used with only a minor differences:

1) it shall be lower case
2) it shall have no space between `;` and `url=`

PR #503, opened 2022-10-03
xlib: Fix xlib text drawing background color

### LINKED ISSUE #502: xlib: 9slice patch introduced a bug
I was working from a checkout 1e4eacca2e3e4abd4d8b793c580c71f2e5315651
After an update to latest on master, all xlib based demos look broken:

overview demo

I found the wrong commit (6322f53cb569bb989873ee8a1b71ae705b230708) and where it occurs (`nk_tree_state_base()` function), but I am not sure about how to fix it.

Indeed, the issue is not visible on other backends (opengl), and the xlib function which draws text also has code to draw the background filling a rectangle, but for which the color is not initialized (hence the rainbow style when running several times).

One solution could be to remove the text background fill (do we have something to say a text has a specific background ?) or to fix the `nk_tree_state_base()` function to correctly initialize the background color.

I would think fixing the function is the solution.

riri:
fixes #502

### DIFF
--- a/demo/x11/nuklear_xlib.h
+++ b/demo/x11/nuklear_xlib.h
@@ -385,15 +385,12 @@ nk_xsurf_stroke_curve(XSurface *surf, struct nk_vec2i p1,
 }
 
 NK_INTERN void
-nk_xsurf_draw_text(XSurface *surf, short x, short y, unsigned short w, unsigned short h,
-    const char *text, int len, XFont *font, struct nk_color cbg, struct nk_color cfg)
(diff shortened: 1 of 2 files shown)


dumblob:
LGTM, thanks!

@RobLoach you have been involved I noticed - could you also look at this PR? There is not much going on but perhaps you should see it also before merging :wink:.

RobLoach:
I was able to reproduce the error. Once I checked out this PR, the error was fixed. Nicely done, @riri :+1: I'm going to merge this one.

PR #504, opened 2022-10-12
demos: set style with defines

riri:
Not a big deal, but it's annoying to have to modify the demos code to see the effect of styles, then revert to keep on the upstream commit.

It's already possible to use defines to select which demos to include, for example:
```
CFLAGS='-DINCLUDE_CANVAS' make
```

This PR adds the possibility to select the used style when `INCLUDE_STYLE` is defined:
```
CFLAGS='-DINCLUDE_OVERVIEW -DINCLUDE_STYLE -DSTYLE_WHITE` make
```

All demos using the `set_style()` function have been updated, and that does not change the default behavior.

### DIFF
--- a/demo/allegro5/main.c
+++ b/demo/allegro5/main.c
@@ -116,6 +116,7 @@ int main(void)
 
     /* style.c */
     #ifdef INCLUDE_STYLE
+    /* ease regression testing during Nuklear release process; not needed for anything else */
     #ifdef STYLE_WHITE
     set_style(ctx, THEME_WHITE);
     #elif defined(STYLE_RED)
--- a/demo/allegro5/main.c
+++ b/demo/allegro5/main.c
@@ -115,10 +115,17 @@ int main(void)
     ctx = nk_allegro5_init(font, display, WINDOW_WIDTH, WINDOW_HEIGHT);
 
     /* style.c */
-    /*set_style(ctx, THEME_WHITE);*/
-    /*set_style(ctx, THEME_RED);*/
-    /*set_style(ctx, THEME_BLUE);*/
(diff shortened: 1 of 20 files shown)


demo/allegro5/main.c
```diff
@@ -115,10 +115,17 @@ int main(void)
...
+    #ifdef STYLE_WHITE
+    set_style(ctx, THEME_WHITE);
+    #elif defined(STYLE_RED)
+    set_style(ctx, THEME_RED);
+    #elif defined(STYLE_BLUE)
+    set_style(ctx, THEME_BLUE);
+    #elif defined(STYLE_DARK)
+    set_style(ctx, THEME_DARK);
+    #endif
+    #endif
```
dumblob, line 128: Maybe add a short comment as reasoning what this ifdef is useful for to make it cleare to newcomers who will be surprised if there is any **need** to do these ifdefs to make Nuklear work properly. We shall emphasize Nuklear itself and its use does not need these ifdefs.

Perhaps something like:

`/* ease regression testing during Nuklear release process; not needed for anything else */`

(this shall be present in each demo file, of course)

dumblob:
LGTM

I think it would be nice to add a comment clarifying the purpose of those ifdefs.

riri:
done, but being lazy, I just stole your message and pasted it in all demos ^^
I don't think the comment was needed, as it follows the same pattern than other ifdef to include extra demo code, but still harmless

dumblob:
>I don't think the comment was needed, as it follows the same pattern than other ifdef to include extra demo code, but still harmless

Yeah, it just appeared to me it might look a bit pointless viewed through the eyes of someone unaware of the Nuklear development :wink:.

Thanks!

PR #506, opened 2022-10-24
Fix nk_str_{append,insert}_str_utf8 always returning 0

Francesco149:
I noticed this problem because of compiler warnings about `runes` being set but unused. I'm assuming they're supposed to return `num_runes`


RobLoach (approved):
Thanks a lot!

riri (approved):
LGTM

dumblob:
Great catch! Thanks.

And thanks to all others for review!

PR #520, opened 2022-11-30
gdi_native: Fix calling convention problem

learn-more:
The calling convention of a WNDPROC is specified as being 'CALLBACK'.


dumblob:
Ok, good catch. Thanks!

PR #521, opened 2022-11-30
Add MSVC CI

learn-more:
Both a 'default' build, and with the define `INCLUDE_ALL` set will be built.
I did not know how to unroll the individual build steps (without making it a huge matrix, where each build will do a new checkout),
so all build steps are copy-pasted manually.


.github/workflows/ccpp.yml
```diff
@@ -19,3 +19,42 @@ jobs:
...
+      matrix:
+        arch: [x86, x64]
+        define: ['', '-DINCLUDE_ALL']
+    runs-on: windows-latest
+
+    steps:
+    - uses: actions/checkout@v3
+
+    - name: Visual Studio shell
+      uses: egor-tensin/vs-shell@v2
```
learn-more, line 34: Please do not merge it yet, I should replace this with `ilammy/msvc-dev-cmd` instead, since that is more actively used / developed

dumblob: Take your time. Let me know once we should jump on this to review :wink:. Thanks!

dumblob:
>I did not know how to unroll the individual build steps (without making it a huge matrix, where each build will do a new checkout),
so all build steps are copy-pasted manually.

I think your approach is totally OK - it is explicit and IMHO not that verbose and allows for quick commenting out things etc.

.github/workflows/ccpp.yml
learn-more: @dumblob I have updated the builder, so feel free to review it.

learn-more: `apt-get` seems to have some trouble ...

learn-more:
@dumblob is there anything I can do to get this merged?

PR #523, opened 2022-12-01
utils: Add .yml indent_size to .editorconfig

learn-more:


.editorconfig
```diff
@@ -12,3 +12,6 @@ trim_trailing_whitespace = true
 
 [clib.json]
 indent_size = 2
+
+[*.yml]
+indent_size = 2
```
RobLoach, line 17:
``` suggestion
[*.json,*.yml]
indent_size = 2
```

PR #530, opened 2022-12-17
DONE: successfull attempt to enable and test vertical scrollbar in no…

bart-at-qqdatafruits:
successful attempt to enable and test vertical scrollbar in node-editor demo, tested on allergo5 and sdl2 only
contribution respects and maintains same license as project


dumblob:
Hi @bart-at-qqdatafruits , sorry for the late reply. Our upstream is a bit slow. I have skimmed all the changes and I like them. It is only one commit but I think there is no need for nit picking. Except for version bump which is required.

I see you closed this PR - are you underway of some additional changes? If not, could you please bump version (take a look at how we review PRs) and reopen this PR?

Having both horizontal & vertical scrolling is a nice and useful feature. I remember there were some issues with scrolling in general, so we need to test this really thoroughly. And that will take some time (probably months), so do not feel unwelcome but rather the opposite - people will appreciate a new well-tested wished feature!

Issue #399, opened 2022-01-19
Some fonts have incorrect glyph indexes


user23333:
Some fonts have incorrect glyph indexes because when a font fails to parse (maybe?) the glyph index is not increase, so the index is wrong

The above is a translation, sorry for my poor English

[https://github.com/Immediate-Mode-UI/Nuklear/blob/9486833f617c6c50d15f6040f184ca9f2c98c6ad/src/nuklear_font.c#L357](https://github.com/Immediate-Mode-UI/Nuklear/blob/9486833f617c6c50d15f6040f184ca9f2c98c6ad/src/nuklear_font.c#L357)
Here glyph_count is not increased

llxiaoyuan:
偷偷写绘制是吧

RobLoach:
Could you provide an example?

robotoxico:
@RobLoach
it comes with some CJK fonts, i'm stucked in the same problem.

eg, specified the rune ranges like using `nk_font_chinese_glyph_ranges`

```c
    NK_API const nk_rune*
    nk_font_chinese_glyph_ranges(void)
    {
        NK_STORAGE const nk_rune ranges[] = {
            0x0020, 0x00FF,
            0x3000, 0x30FF,
            0x31F0, 0x31FF,
            0xFF00, 0xFFEF,
            0x4e00, 0x9FAF,
            0
        };
        return ranges;
    }
```

`nk_font_atlas_bake` will gets the total number of glyphs specified above, ignores it can be found in the font or not, **it is ok**.
when starts `stbtt_PackFontRangesGatherRects`, this

```c
    int glyph = stbtt_FindGlyphIndex(info, codepoint);
    if (glyph == 0 && (spc->skip_missing || missing_glyph_added)) {
      rects[k].w = rects[k].h = 0;
    } else {
```

indicates that the glyph not found, but the index of the rects which is `k` is still increase.

at last, in the `nk_font_bake`

```c
    /* query glyph bounds from stb_truetype */
    const stbtt_packedchar *pc = &range->chardata_for_range[char_idx];
    if (!pc->x0 && !pc->x1 && !pc->y0 && !pc->y1) continue;
    codepoint = (nk_rune)(range->first_unicode_codepoint_in_range + char_idx);
```

as @user23333 mentioned, the `continue` causes the `glyph_count` miss increasing, **that is the problem**.
next codes:

```c
    /* fill own glyph type with data */
[...]

PR #531, opened 2022-12-22
Fix incorrect glyph index in nk_font_bake

dulingzhi:
In stbtt_PackFontRangesGatherRects, it skips packing when glyphs are missing. Resulting in incorrect glyphs when nk_font_bake reading these ranges rects data. Referring to the way IMGUI handles it, we skip the rect data check.

ref #399 

load font:
image

before:
image

after:
image

### DIFF
--- a/clib.json
+++ b/clib.json
@@ -1,6 +1,6 @@
 {
   "name": "nuklear",
-  "version": "4.10.5",
+  "version": "4.10.6",
   "repo": "Immediate-Mode-UI/Nuklear",
   "description": "A small ANSI C gui toolkit",
   "keywords": ["gl", "ui", "toolkit"],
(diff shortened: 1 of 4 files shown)


dumblob:
Thanks for looking into this. Did you test this with various demos (and ideally also backends)?

Also, could you please also [update the source files and version](https://github.com/Immediate-Mode-UI/Nuklear#reviewers-guide )?

dulingzhi:
> 
source files and version updated.
I tried the opengl backends and it works fine. And I'm still trying to compile the example program on windows.

dulingzhi:
截屏2022-12-24 02 20 13
截屏2022-12-24 02 19 43
截屏2022-12-24 02 19 21
截屏2022-12-24 02 18 33


Works fine in all examples

FrostKiwi:
OMG Yes. Thx so much for finding the cause!
This has been plaguing me from time to time and I resorted to excluding the affected glyphs. Great to have a proper fix for this.

dumblob:
I think we could land this. Anyone any objections?

RobLoach (approved):
:rocket:

dulingzhi:
> 🚀

Is it possible to merge this PR?

PR #538, opened 2023-01-13
Remove link to no longer existing tool

LonerDan:
The link to the online single header packer (https://apoorvaj.io/single-header-packer.html) no longer works (returns 404), so remove it.
If you happen to know that it still exists but at a different URL, feel free to re-add it.


dumblob:
Thanks!

PR #550, opened 2023-03-01
demo/allegro5: minor string-related tweaks

lockie:
Hello! This tiny PR consists of a few string handling tweaks in allegro5 backend:
* local variable `strcpy` renamed to `str`, as per SEI CERT [DCL01-C](https://wiki.sei.cmu.edu/confluence/display/c/DCL01-C.+Do+not+reuse+variable+names+in+subscopes).
* `malloc` and manual null-terminator addition replaced with `calloc`.
* added missing `NULL` from allocator check.
* `strcpy` replaced with `strncpy` for better security.

### DIFF
--- a/demo/allegro5/nuklear_allegro5.h
+++ b/demo/allegro5/nuklear_allegro5.h
@@ -101,7 +101,7 @@ static float
 nk_allegro5_font_get_text_width(nk_handle handle, float height, const char *text, int len)
 {
     float width;
-    char *strcpy;
+    char *str;
     NkAllegro5Font *font = (NkAllegro5Font*)handle.ptr;
     NK_UNUSED(height);
     if (!font || !text) {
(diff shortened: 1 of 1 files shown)


RobLoach:
Worked locally. Thanks!

demo/allegro5/nuklear_allegro5.h
```diff
@@ -111,11 +111,11 @@ nk_allegro5_font_get_text_width(nk_handle handle, float height, const char *text
        as nuklear uses variable size buffers and al_get_text_width doesn't
        accept a length, it infers length from null-termination
        (which is unsafe API design by allegro devs!) */
-    strcpy = malloc(len + 1);
-    strncpy(strcpy, text, len);
-    strcpy[len] = '\0';
-    width = al_get_text_width(font->font, strcpy);
-    free(strcpy);
+    str = calloc((size_t)len + 1, 1);
```
dumblob, line 114: When at it, shouldn't the last `1` be actually `sizeof( *str )`?

RobLoach: Or `sizeof(char)` ?

lockie: I believe `sizeof(char) != 1` is true only on weird architectures that went extinct decades ago. I'll change that if you insist though :slightly_smiling_face:

dumblob: Just leave things as they are. I had yet another nit pick day, sorry for that :wink:.

lockie: No problem. Thanks for your dedication :blush:

PR #551, opened 2023-03-06
Fix nk_draw_button_text_* to use nk_flags align passed from nk_menu_item_image_text etc

wuxxerno:
nk_flags align were lost in transit and not passed to the final draw functions, where a NK_TEXT_CENTERED was in its place.


RobLoach:
Looks like it's been picked up over at https://github.com/Immediate-Mode-UI/Nuklear/pull/674 .

Issue #494, opened 2022-09-06
Many trackpad mouse down events are lost (MacOS+SDL2+OpenGL2)


cesss:
I'm using SDL2 with OpenGL2, on Macs with a Magic Trackpad configured as "tap to click" (ie: mouse-down events are generated by just tapping the trackpad, without having to fully press it).

When I tap, most of the times the mouse-down event is lost, so togglebuttons are not toggled, buttons are not pressed, etc, etc...

A few times it works, other times doesn't. Strange. I tried to debug it, but I don't know the Nuklear internals, so I don't know where to look for.

The only things I can say for the moment are these:

- **nk_sdl_handle_event()** in `nuklear_sdl_gl2.h` is getting an event for **all the taps**. No events are lost. So, the problem is not in SDL2, because the events are arriving correctly to Nuklear.
- For every tap, I'm getting both a mouse down and a mouse up event. So, the problem is not about mouse clicks being unterminated or unstarted.
- In every mouse up/down event, `ctx->input.mouse.buttons[NK_BUTTON_LEFT].down` is different from `evt->type == SDL_MOUSEBUTTONDOWN`. So, function **nk_input_button()** doesn't return prematurely at the check for both being equal.

As I said, I don't know the Nuklear internals, so I wasn't able to dig deeper. Do you have any idea on what might be happening?

cesss:
I was able to find the problem. The reason is that for these Trackpad taps, MacOS generates two consecutive events. File `main.c` of the SDL+OpenGL2 demo has this loop, whose purpose I guess is to combine consecutive events and thus be more efficient and improve performance:

```
while (SDL_PollEvent(&evt)) {
            if (evt.type == SDL_QUIT) goto cleanup;
            nk_sdl_handle_event(&evt);
        }
```

However, some events should **never** be combined. I've modified the loop as follows, and now all Trackpad taps are handled correctly, and no events are lost:

```
while (SDL_PollEvent(&evt)) {
            if (evt.type == SDL_QUIT) goto cleanup;
            nk_sdl_handle_event(&evt);
            if((evt.type==SDL_MOUSEBUTTONUP)||(evt.type==SDL_MOUSEBUTTONDOWN)||(evt.type==SDL_KEYUP)||(evt.type==SDL_KEYDOWN)) break;
        }
```
As you can see, I also included key up and key down events as not-combinable. I didn't have a problem with key up/down events, but my mind says is not a good idea to try to combine them.

[...]

PR #553, opened 2023-04-09
MacOS Trackpad fixes in SDL demos

cblc:
Fixes https://github.com/Immediate-Mode-UI/Nuklear/issues/494

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,4 @@ docs/src
 *.swo
 *.swp
 /private/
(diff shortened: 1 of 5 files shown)


RobLoach:
This seems like an improper way to "fix" the issue here. SDL_PollEvent() is expected to loop through every event, and when we are breaking from the loop, it's possible we'll lose pushed event triggers in the event queue.

PR #560, opened 2023-05-30
[New feature] Add ability to change text color/background color via (optional) inline code; should be backward compatible

xzn:
Originally requested here: https://github.com/vurtun/nuklear/issues/575

I happen to need this feature in my app so this is an implementation of it.

### How it works
A new property `draw_config` is stored in `nk_context` and passed to `nk_command_buffer` when windows are drawn.
`nk_draw_text` then use the settings in the command buffer to draw text as required.

There are two ways to encode inline color information:

1.
```c
nk_draw_set_color_inline(ctx, NK_COLOR_INLINE_TAG);
/* then pass text to widget like follows: */
nk_label(ctx, "default [color=#ff0000]red[/color] color", NK_TEXT_LEFT)
```

2.
```c
nk_draw_set_color_inline(ctx, NK_COLOR_INLINE_ESCAPE_TAG);
/* then pass text to widget like follows: */
nk_label(ctx, "default \e[color=#ff0000]red\e[/color] color", NK_TEXT_LEFT)
```

The difference is that in the first option, `'\e'` can be used to escape and display the opening tag without changing the color of subsequent text, e.g:
```c
"displaying color tag \e[color=#ff00ff]code"
```
In the second option `'\e'` must be used before each opening and closing tag for them to take effect.

You can also give _names_ to colors like so:

```c
const char *color_name = "my_color";
struct nk_color color= nk_rgb_hex("bb00bb");
struct nk_map_name_color color_name_map;
nk_map_name_color_init_colors(&color_name_map, &ctx->memory.pool, &color_name, &color, 1);

nk_draw_push_map_name_color(ctx, &color_name_map);

nk_draw_set_color_inline(ctx, NK_COLOR_INLINE_TAG);
nk_label(ctx, "default [color=\"my_color\"]some[/color] color", NK_TEXT_LEFT)
```

### Notes on possible improvement in code

Most of the changes in this [file](https://github.com/xzn/Nuklear/commit/527c85a40df981c51920bb103749959dd1b6f9eb#diff-18bc2a7408fe6d3751f4078220d46c8ff243928c6ca61c17f652f9b9bb637743) should not be needed, most of it can be cut down if we say, include [`stb_ds.h`](https://github.com/nothings/stb/blob/master/stb_ds.h) and use the hash map implementation for user-defined color name to color map.
Alternatively we can choose to not support user-define color names (although it's already implemented in this PR).

### Testing done so far

This PR has only been moderately tested. There should not be any buffer overrun or off by one error. If you've found any bugs please let me know so I can fix them.

Only the built-in `stb_truetype` font renderer has been tested. No idea right now how this feature will fare against different font rendering backend.

## Example usages:

Will be added later

## Screenshots

Will be added later.

### DIFF
--- a/nuklear.h
+++ b/nuklear.h
@@ -48,6 +48,7 @@
 /// - No global or hidden state
 /// - Customizable library modules (you can compile and use only what you need)
 /// - Optional font baker and vertex buffer output
+/// - [Code available on github](https://github.com/Immediate-Mode-UI/Nuklear/)
 ///
 /// ## Features
 /// - Absolutely no platform dependent code
@@ -1406,7 +1407,7 @@ NK_API const struct nk_draw_command* nk__draw_next(const struct nk_draw_command*
 /// Function                            | Description
 /// ------------------------------------|----------------------------------------
 /// nk_begin                            | Starts a new window; needs to be called every frame for every window (unless hidden) or otherwise the window gets removed
-/// nk_begin_titled                     | Extended window start with separated title and identifier to allow multiple windows with same name but not title
+/// nk_begin_titled                     | Extended window start with separated title and identifier to allow multiple windows with same title but not name
 /// nk_end                              | Needs to be called at the end of the window building process to process scaling, scrollbars and general cleanup
 //
 /// nk_window_find                      | Finds and returns the window with give name
@@ -4614,6 +4615,7 @@ struct nk_command_buffer {
     int use_clipping;
     nk_handle userdata;
     nk_size begin, end, last;
+    struct nk_draw_config *draw_config;
 };
 
 /* shape outlines */
@@ -4638,6 +4640,7 @@ NK_API void nk_fill_polygon(struct nk_command_buffer*, float*, int point_count, 
 NK_API void nk_draw_image(struct nk_command_buffer*, struct nk_rect, const struct nk_image*, struct nk_color);
 NK_API void nk_draw_nine_slice(struct nk_command_buffer*, struct nk_rect, const struct nk_nine_slice*, struct nk_color);
 NK_API void nk_draw_text(struct nk_command_buffer*, struct nk_rect, const char *text, int len, const struct nk_user_font*, struct nk_color, struct nk_color);
+NK_API int nk_draw_raw_text(struct nk_command_buffer*, struct nk_rect, const char *text, int len, const struct nk_user_font*, struct nk_color, struct nk_color, float *w);
 NK_API void nk_push_scissor(struct nk_command_buffer*, struct nk_rect);
 NK_API void nk_push_custom(struct nk_command_buffer*, struct nk_rect, nk_command_custom_callback, nk_handle usr);
 
@@ -4856,6 +4859,10 @@ NK_API void nk_draw_list_push_userdata(struct nk_draw_list*, nk_handle userdata)
  *                          GUI
  *
  * ===============================================================*/
+#ifndef NK_NAME_COLOR_MAX_NAME
+#define NK_NAME_COLOR_MAX_NAME 32
+#endif
+
 enum nk_style_item_type {
     NK_STYLE_ITEM_COLOR,
     NK_STYLE_ITEM_IMAGE,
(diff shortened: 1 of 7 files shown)


YgorVasilenko:
Really great feature! I also would like to see it merged. For now, I will have to integrate this into my own, already modified version of Nuklear. Having this merged into master would make life easier!

xzn:
Hi @YgorVasilenko thanks for the kind words!

Please let me know of any issues you may have while trying to use this feature.

I'm looking for feedback on potential issues in the implementation if you could:

- [ ] How easy it is to use the new functions correctly.
- [ ] The string parsing code [here](https://github.com/Immediate-Mode-UI/Nuklear/pull/560/files#diff-fb50decbe365b536ab75ec94939d6032a35932d0570bcbe31a7beebb9a3b802dR743) is fairly ad-hoc and I'm not sure if it's actually doing the correct thing (aside from the limited testing I did).
- [ ] Whether the ways memory management is used are correct for custom color names.
- [ ] Whether texts are aligned properly when rendered in ranges of different colors.

There are likely other concerns with the PR but these are of my own for now.

YgorVasilenko:
Hi, first issue I faced: MSVC compiler turns '\e' into same ASCII as 'e' (101). This resulted in 'e' being stripped out from any colored text completely. I commented out branches of code, that do something, if symbol is '\e', and that helped.<br>
Now I see that coloring of single string seems to be working<br>
image

So far no memory issues.<br>
Btw, this is where I integrated your feature, so this could be used as testing platform (it's UI editor)<br>
https://github.com/YgorVasilenko/IndieGoUI/tree/elven_city_simulator

Ryder17z:
As '\' is a special character, you might have to escape it like this: '\\e'

xzn:
Sorry my bad for not realizing that `\e` is not portable C and thus not working on MSVC. Maybe change it to `\033`?

nyaruku:
Seems working well, i like this and solved my problem of having multiple texts in different colors in a horizontal row,

nyaruku:
> \e
yep this should be also fixed, currently trying to outcomment all /e stuff too

xzn:
There's a bug that wrapped text aren't currently handled. Might take a while

xzn:
Updated first post to reflect the changes. A macro `NK_ESC` is now defined to be `"\033"`, with `NK_ESC_CHAR` pointing to the same char. Can be used in string literal for concatenation.

xzn:
Currently edit box can be styled this way as well, leading to cursor being in the wrong place.

Should it be changed so that edit boxes are always wrapped in between

```
nk_draw_push_color_inline(ctx, NK_COLOR_INLINE_NONE);
...
nk_draw_pop_color_inline(ctx);
```

Or should edit box supports this way of styling? The latter is more complicated..

YgorVasilenko:
Thanks a lot for wrapping text fix, we'll use that!

nyaruku:
image
C++ Function to create a gradient colored text

nyaruku:
I use c++ to get a gradient color output: 


```C
// You have to decleare an extra variable for gradientText()
// If u nest it into custom_strcat() u will still get memory leaked by gradientText()

const char* grText = gradientText("Nyaruku", "#7C7CFF", "#FF016F");
const char* labelText = custom_strcat(2, "My Discord: @", grText);

nk_label(ctx, labelText, 17);

// Prevent Memory Leak
free((void*)grText);
free((void*)labelText);

```

```cpp
struct Color {
	int r, g, b;

	Color(int red, int green, int blue) : r(red), g(green), b(blue) {}

	// Linear interpolation between two colors
	static Color interpolate(const Color& start, const Color& end, double t) {
		int r = static_cast<int>((1 - t) * start.r + t * end.r);
		int g = static_cast<int>((1 - t) * start.g + t * end.g);
		int b = static_cast<int>((1 - t) * start.b + t * end.b);
		
		return Color(r, g, b);
	}
};

Color getColorForStep(const Color& start, const Color& end, int steps, int step) {
	if (step < 0) step = 0;
	if (step > steps) step = steps;

	double t = static_cast<double>(step) / steps;
	return Color::interpolate(start, end, t);
}

Color HexColorToRGB(const char* hexColor) {
	Color color{255,255,255};
	if (hexColor[0] == '#' && strlen(hexColor) == 7) {
		std::stringstream ss;
		ss << std::hex << hexColor + 1; // Skip the '#' character

		int colorValue;
		ss >> colorValue;
		color.r = (colorValue >> 16) & 0xFF;
		color.g = (colorValue >> 8) & 0xFF;
		color.b = colorValue & 0xFF;
		ss.clear();
	}
	else {
		// Invalid hex color format, return the original color
	}
	
	return color;
}
std::string RGBColorToHex(Color color) {
	std::stringstream ss;
	ss << "#" << std::setfill('0') << std::setw(2) << std::hex << color.r
		<< std::setw(2) << std::hex << color.g << std::setw(2) << std::hex << color.b;
	std::string temp = ss.str();
	ss.clear();
	return temp;
}

const char* gradientText(const char* text, const char* color_start, const char* color_end) {
	std::stringstream result;
	Color color1 = HexColorToRGB(color_start);
	Color color2 = HexColorToRGB(color_end);

	int totalSteps = strlen(text);
	for (int i = 0; i < totalSteps; i++) {
		result << "[color=" << RGBColorToHex(getColorForStep(color1, color2, totalSteps, i)) << "]" << text[i] << "[/color]";
	}

	std::string resultStr = result.str();
	char* cString = new char[resultStr.size() + 1];
	strcpy(cString, resultStr.c_str());

	return cString;
}```

yukyduky:
image
I had to cast the memory in 5 places before I got it to compile since I'm coding in C++, might be something worth adding before pulling this in. Issue is on line 8380, 9964, 8192, 8220, 8335. I'm going to play around with the code now, but it looks promising!

xzn:
@yukyduky I've added the casts to the commits to help C++ compat.

For the wrapping text, did you update the container size for the wrapped text when you resize the window? Does it not work properly even when you do that?

yukyduky:
@xzn Nice! Thanks for the quick update :) 

One of the benefits is and should be to be able to resize the windows on the fly while the app is running, imo. It's possible to do so in the current version so can't be too hard to get it working again.

yukyduky:
@xzn Ahh this is my fault, I've pretty much only used the nk_layout_dynamic_row and didn't think about that I was using your window with the static layout to test your code. In that case I have no further issues, sorry for the confusion^^

Sororfortuna:
Why is this not merged?

mhcerri:
Any plans to merge this feature?

zoogies:
Would be great to see this merged

lockie:
@xzn hey, any chance to make this work with `nk_edit`s?

xzn:
nk_edit are excluded since I'm not sure what is the expected behavior.

Do you want the color command to be visible in nk_edit? If not, there's extra work to get the cursor to appear in the correct place.

PR #561, opened 2023-06-02
Node editor demo expanded

hwps:
This PR expands on the node editor demo to make it a functional example of a node-based UI.

Features: 
- multiple node types
- multiple connector types
- linking between connectors of the same type
- detaching and reattaching links
- getting value from linked node if slot is connected

node_editor

Some notes:
- only tested with the glfw_opengl3 backend
- uses system malloc, not sure if nk_malloc should be preferred.
- uses the standard panel close button , then checks for NK_WINDOW_HIDDEN whether to delete a node or display it. Not sure if there's some pitfall here.

### DIFF
--- a/demo/common/node_editor.c
+++ b/demo/common/node_editor.c
@@ -66,6 +66,7 @@ struct node_editor {
     int initialized;
     struct node *node_buf[32];
     struct node_link links[64];
+    struct node *outputNode;
     struct node *begin;
     struct node *end;
     int node_count;
@@ -196,6 +197,9 @@ node_editor_link(struct node_editor *editor, struct node *in_node, int in_slot,
     link = &editor->links[editor->link_count++];
     out_node->inputs[out_slot].isConnected = nk_true;
     in_node->outputs[in_slot].isConnected = nk_true;
+    out_node->inputs[out_slot].connectedNode = in_node;
+    out_node->inputs[out_slot].connectedSlot = in_slot;
+    
     link->input_node = in_node;
     link->input_slot = in_slot;
     link->output_node = out_node;
@@ -209,8 +213,10 @@ node_editor_init(struct node_editor *editor)
     memset(editor, 0, sizeof(*editor));
     editor->begin = NULL;
     editor->end = NULL;
+
+    editor->outputNode = node_output_create(editor, (struct nk_vec2){600, 400});
+
     node_color_create(editor, (struct nk_vec2){40, 10});
-    node_output_create(editor, (struct nk_vec2){200, 200});
     editor->show_grid = nk_true;
 }
 
--- a/demo/common/node_editor.c
+++ b/demo/common/node_editor.c
@@ -13,6 +13,8 @@
 
 
 enum node_connector_type {fValue, fColor};
+enum node_type {color, mix};
+
 
 struct node_connector {
     enum node_connector_type type;
@@ -23,6 +25,7 @@ struct node {
     int ID;
     char name[32];
     struct nk_rect bounds;
+    enum node_type type;
     float value; /* unused */
     struct nk_color color;
     int input_count;
@@ -37,9 +40,9 @@ struct node {
 };
 
 struct node_link {
-    int input_id;
+    struct node* input_node;
     int input_slot;
-    int output_id;
+    struct node* output_node;
     int output_slot;
     /* struct nk_vec2 in;
     struct nk_vec2 out; */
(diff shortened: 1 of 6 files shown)


leobenaducci:
Nice stuff

RobLoach:
I like where this is going. I'm not sure what's wrong with the demo though? This is the sdl_renderer....
Screenshot at 2023-09-12 03-25-01

hwps:
@RobLoach Can you click the connectors to draw a link or right click to add nodes?

Maybe there should be some more nodes or at least links in the initial state. I realize now that for someone not used to a node-based UI it might be non-obvious what to do.

hwps:
Ok, finally did this slight update to the init state. Screenshot attached.

Screenshot 2023-12-11 at 19 44 53

riri:
Nice work! I ran it on `glfw_opengl3` and `x11_xft` demos (screenshot for the later below)

x11_xft

riri (approved):

RobLoach (approved):
Finally had another look at this. Looks great!

efa:
is this PR added to the master branch?
Because I cannot find any of this files in path:
demo/common/nodeeditor/*

RobLoach:
Thanks for the ping, @efa. Could we get a more impressive initial state for the node editor? While I don't think that should be a blocker for merging, keep it as clear as the original may help demonstrate the node editor better. Perhaps recreating the original state with the new update could even be good.

### Before
Screenshot from 2024-09-04 11-52-29

### After
Screenshot from 2024-09-04 11-53-31

efa:
undoubtedly an aesthetic with colors attracts a lot of attention, so to a non-programmer it seems much nicer than before.

In my application I only needed numbers (the graph payload) so no colors, so I changed the node editor a bit.
But above all I needed to be able to delete links and nodes (deleting connected links too), so I changed the struct allocation a bit making it dynamic, and I included my data payload. I discovered this PR after my modification, that maybe it is done even better than how I did it. I plan to do a merge and provide a giveback once I'm satisfied. Anyway that part is already now very stable.

efa:
PS. Congratulations for this great modular and compact library

efa:
sure this August will try to synch my fork (as now included in https://github.com/efa/PowerBudget) with upstream and propose a PR if still miss some features in node editor engine.

PS I'm not sure about right licensing. Currently I'm using GPL for my code, but Nuklear is MIT/public domain. I'd like to cause as few problems as possible given the excellent basic work of the library, and my additions are only small ones. Do I need to declare the "node_editor.c" file I modified under the MIT license to be compatible?

awschult002:
> PS I'm not sure about right licensing. Currently I'm using GPL for my code, but Nuklear is MIT/public domain. I'd like to cause as few problems as possible given the excellent basic work of the library, and my additions are only small ones. Do I need to declare the "node_editor.c" file I modified under the MIT license to be compatible?

Although i am not an authority on licensing, nor am i an authority on this project, I would say "yes" there would be conflict. The `node_editor.c` file (if you want it updated into this project) will be combined into a single header file with all of the other code. 

There is nothing that says your entire project needs to be 1 license. each file could technically have a difference license. So you could state the license in comments at the top of the file. but then that confuses things if its different for Nuklear; as someone would really need to consider different licenses within the header and then which GPL it is and whether they can use multiples in their own project and .. and .. and..

If I were you, and you don't care about the licensing too much. I would submit the code for `node_editor.c`  as MIT for this project. Then, if you really want to, submit the code to your project under GPL. You are the author of the code and you are allowed to change the license. However, if you are not the sole author, and part of `node_editor.c` was MIT by someone else; then I am pretty sure you are not allowed to change the license without the express authorization of the previous author. 

hopefully someone else can help me out with this.

efa:
> If I were you, and you don't care about the licensing too much. I would submit the code for `node_editor.c` as MIT for this project. Then, if you really want to, submit the code to your project under GPL

OK, when will be the time, I will ask details in the PR to avoid pollute here. Thank you for now!

aichingm:
> If I were you, and you don't care about the licensing too much. I would submit the code for node_editor.c as MIT for this project.

@awschult002 I don't see it that way. Code contributed to this project has to be licensed under the projects license. The projects license is not only MIT but also Public Domain. 

What this meas to me (IANAL) is that the contributor has be allow the users of the project to use the contributed code under the projects license (in this case all of the used license options (MIT and PD) and not "any" or only a subset). 

@efa This would mean, that you would pass the code to Nuklear under a dual license MIT&PD this is only possible if

* you are the sole creator of the code
* or all of the creators a fine with the dual licensing (MIT&PD)

If you have our own project where you use the code you can grant that project any licensing terms you want (be it GPL or even proprietary).

RobLoach:
Built upon this over at https://github.com/Immediate-Mode-UI/Nuklear/pull/685 .

PR #564, opened 2023-07-10
[Documentation] Remove the "doc" binary

RobLoach:
This change removes the doc/doc binary, and .gitignores it so that it's no longer added to the repository.


dumblob:
Good catch! Thanks.

PR #573, opened 2023-08-18
fix compiling error when using source mode

walkthetalk:


cmake/nuklear-config.cmake
```diff
@@ -0,0 +1,15 @@
...
+get_filename_component(nuklear_INSTALL_PREFIX "${CMAKE_CURRENT_LIST_DIR}/../" ABSOLUTE)
+
+set(nuklear_INCLUDE_DIRS ${nuklear_INSTALL_PREFIX})
+
+add_library(nuklear INTERFACE IMPORTED)
+set_target_properties(nuklear PROPERTIES
+	INTERFACE_INCLUDE_DIRECTORIES "${nuklear_INCLUDE_DIRS}"
+)
+
+find_package_handle_standard_args(nuklear DEFAULT_MSG nuklear_INCLUDE_DIRS)
```
RobLoach, line 15: I see we're adding a Nuklear config here. Should this be in the root?

RobLoach, line 15:
``` suggestion
```

PR #577, opened 2023-09-12
Use __PPC64__ macro for GCC detection

fitzsim:
PowerPC 64-bit platforms use the lowercase `__ppc64__` on Apple systems (based on comments I've seen).  However lowercase `__ppc64__` is not defined by GCC; uppercase `__PPC64__` is.  This pull request adds a check for the uppercase macro so that compiling Nuklear.h works on systems that use GCC, like Debian GNU/Linux ppc64le.


RobLoach (approved):
Seems good with me! I'll let someone else give another :eyes: though

PR #580, opened 2023-10-11
Added feature to disable widgets

yukyduky:
I've added functions to disable widgets and gray them out. You use it like this:

nk_widget_disable_begin(ctx);
nk_widget(...);
nk_widget(...);
nk_widget(...);
nk_widget_disable_end(ctx);

There's a variable in each style struct in case you want to change the gray out factor per widget type. So far I've only tested it on the widgets in overview.c. Let me know if you have any issues :)


src/nuklear_color.c
```diff
@@ -23,6 +23,16 @@ nk_parse_hex(const char *p, int length)
...
 }
 NK_API struct nk_color
+nk_rgb_factor(const struct nk_color col, const float factor)
+{
+    struct nk_color ret;
+    ret.r = col.r * factor;
+    ret.g = col.g * factor;
+    ret.b = col.b * factor;
+    ret.a = col.a;
+    return ret;
```
RobLoach, line 33: Could have micro performance tweak here :wink:  .... Also, not impacting the alpha is intentional, correct?
``` suggestion
    if (factor == 1.0f)
        return col;
    col.r *= factor;
    col.g *= factor;
    col.b *= factor;
    return col;
```

src/nuklear_style.c
```diff
@@ -193,6 +201,8 @@ nk_style_from_table(struct nk_context *ctx, const struct nk_color *table)
     toggle->border_color    = nk_rgba(0,0,0,0);
     toggle->border          = 0.0f;
     toggle->spacing         = 4;
+    toggle->color_factor    = 1.0f;
+    toggle->disabled_factor = 0.5f;
```
RobLoach, line 205: Lots of setting of `0.5f`. Perhaps might benefit from adding a `NK_WIDGET_DISABLED_FACTOR` or something so that it's easy to change globally?

src/nuklear_color.c
yukyduky: Ahh yes! Good catch!

src/nuklear_style.c
yukyduky: Very true, I will fix that :)

RobLoach:
Looks great! Thanks so much. Will let others review :+1:

kotvkvante:
That's great. 

I think it would be cool to add another option to disable all widgets on window. For example add flag like  `NK_WINDOW_DISABLED`. We already have `NK_WINDOW_NO_INPUT`, but it's useless (i think).
I understand that this flag handles different purpose. But I think no one need just disable input without any visual effect of this.

yukyduky:
Neat idea! I personally don't need that atm, but I already have an idea of how to add it so I'll look into it.

kotvkvante:
By the way, is it good to change style parameters for every widget and restore them back every time? If I'm using  only buttons in my GUI, can I change only button style? I don't know lot about how  compilers works (does he enough clever to remove unnecessary assignments??), but.. I go crazy when I think about the speed of programs... If there is no other options to do this or this just such things works, then ok.. I'm just beginner programmer.

riri:
I'll check this asap

src/nuklear_edit.c
```diff
@@ -89,6 +89,9 @@ nk_edit_draw_text(struct nk_command_buffer *out,
     float line_offset = 0;
     int line_count = 0;
 
+    foreground = nk_rgb_factor(foreground, style->color_factor);
+    background = nk_rgb_factor(background, style->color_factor);
+
     struct nk_text txt;
```
riri, line 95: Move this up with other variable declarations to be C90 compliant.

riri (changes requested):
Aside the C90 compliance change I have asked, I suggest making the disabling functionality more visible in the demo, adding a `nk_bool disable_widgets = nk_false` variable in the overview, adding a checkbox to change that variable in the Window tree content
```c
            nk_checkbox_flags_label(ctx, "Minimizable", &window_flags, NK_WINDOW_MINIMIZABLE);
            nk_checkbox_flags_label(ctx, "Scale Left", &window_flags, NK_WINDOW_SCALE_LEFT);
            nk_checkbox_label(ctx, "Disable widgets", &disable_widgets);
            nk_tree_pop(ctx);
```

 and globally disabling the Widgets tree content.
```c
        if (nk_tree_push(ctx, NK_TREE_TAB, "Widgets", NK_MINIMIZED))
        {
            enum options {A,B,C};
            static int checkbox;
            static int option;
            if (disable_widgets) nk_widget_disable_begin(ctx);

// ...

            if (nk_tree_push(ctx, NK_TREE_NODE, "Horizontal Rule", NK_MINIMIZED))
            {
                nk_layout_row_dynamic(ctx, 12, 1);
                nk_label(ctx, "Use this to subdivide spaces visually", NK_TEXT_LEFT);
                nk_layout_row_dynamic(ctx, 4, 1);
                nk_rule_horizontal(ctx, nk_white, nk_true);
                nk_layout_row_dynamic(ctx, 75, 1);
                nk_label_wrap(ctx, "Best used in 'Card'-like layouts, with a bigger title font on top. Takes on the size of the previous layout definition. Rounding optional.");
                nk_tree_pop(ctx);
            }
            if (disable_widgets) nk_widget_disable_end(ctx);

            nk_tree_pop(ctx);
```
I've tried locally and it works great :)

yukyduky:
I will try and fix this ASAP :)

riri (approved):
LGTM

riri:
@yukyduky  Thanks for the functionality and efforts after the review ;-)

PR #583, opened 2023-10-22
add vulkan demo

m0ppers:
finally I found the time to create a PR.

This is mostly based on https://github.com/m0ppers/nuklear-glfw-vulkan but it now uses an independent pipeline to make it more compatible and not assume too much global state. The original reason I created this was this: https://www.youtube.com/watch?v=ySk0Vo70ox0 we were using nuklear to draw the UI :) not sure if there are other nuklear projects that are running on android?

this is quite a heavy PR due to the nature of vulkan unfortunately. I have tried to make it c89 compatible.

I am using lots of memsets. I am not sure if I should be using NK_MEMSET?

the demo application uses bools for errors. however the nuklear_glfw_vulkan.h does not as the other integrations also just use void. not sure if this is ok.

this is not super clean but it should work I think.

I am happy to change things based on feedback.


RobLoach (approved):
I'm unable to test it on my machine, but this looks great. Thanks a lot. Happy to have more implementations available.

m0ppers:
so what are the next steps? :thinking:

RobLoach:
I'm happy to merge, but it would be awesome to get another person to review. Let's leave it up for a bit to see if someone else can try it out.

RobLoach (approved):

riri:
I cloned the project, it compiles but I get an error at execution. I'm not familiar with vulkan so I don't know if it's a normal error or not, but would like to be able to see the render before I give my opinion:
```
Available vulkan layers:
  VK_LAYER_VALVE_steam_fossilize_32
  VK_LAYER_VALVE_steam_fossilize_64
  VK_LAYER_VALVE_steam_overlay_32
  VK_LAYER_VALVE_steam_overlay_64
  VK_LAYER_MANGOHUD_overlay_x86
  VK_LAYER_MANGOAPP_overlay
  VK_LAYER_MANGOHUD_overlay_x86_64
  VK_LAYER_MANGOAPP_overlay
  VK_LAYER_NV_optimus
Couldn't find validation layer VK_LAYER_KHRONOS_validation
failed to create vulkan demo!
```

m0ppers:
@riri  the vulkan validation layers are an integral part when doing vulkan development. they check api calls for validity and verify things against what the driver/GPU supports. for example when you try to allocate a large buffer and the buffer is larger than what the driver supports you will get a nice message that you did something wrong. more details here: https://github.com/KhronosGroup/Vulkan-ValidationLayers#introduction

In a release version you would disable them as they impose quite some overhead but they are absolutely crucial when doing development. As the demo is likely being used by people wanting to integrate nuklear into their project (so these are people having them installed anyway) I made them non-optional because they are "needed" for the development of the demo itself anyway.

It is of course possible to make them optional (note that this ONLY applies to the vulkan demo - the nuklear vulkan implementation does NOT depend on them). Should I do it? As we can see here it is not completely unlikely that people would start the demo without having them installed :joy_cat: so maybe that is a good idea even if they are not really the target audience.

However they can easily be installed. for arch linux the package is "vulkan-validation-layers". for ubuntu: vulkan-validationlayers-dev. On windows you would have to install the vulkan sdk (https://vulkan.lunarg.com/). for mac there is an emulation layer on top of metal that allows mac to run vulkan applications however as far as I know the demo application needs some changes to make it work on mac (I don't have a mac test device). So the demo is currently win/linux only.

I think I answered my own question meanwhile. I will make them optional :joy:

riri:
> I think I answered my own question meanwhile. I will make them optional 😂

Yes you did lol

Thanks for the clear information. I think then it's better to make it optional either with a define or a boolean value. People knowing how to develop with vulkan will know what to do, and others (like me) just curious and reading a vulkan implementation wont be blocked.

I'll add the arch package in the meanwhile.

riri:
Perfect, I merge

RobLoach:
Thanks @m0ppers and @riri!

Issue #513, opened 2022-11-16
warning: potential null pointer dereference


rjopek:
Hi, 

I did the tests on linux.

```
../../../../../../include/nuklear_xlib/../nuklear.h:8838:14: warning: potential null pointer dereference [-Wnull-dereference]
 8838 |     cb->last = b->allocated;
      |     ~~~~~~~~~^~~~~~~~~~~~~~
../../../../../../include/nuklear_xlib/../nuklear.h:8837:13: warning: potential null pointer dereference [-Wnull-dereference]
 8837 |     cb->end = b->allocated;
      |     ~~~~~~~~^~~~~~~~~~~~~~
../../../../../../include/nuklear_xlib/../nuklear.h:8836:15: warning: potential null pointer dereference [-Wnull-dereference]
 8836 |     cb->begin = b->allocated;
      |     ~~~~~~~~~~^~~~~~~~~~~~~~
../../../../../../include/nuklear_xlib/../nuklear.h:8835:22: warning: potential null pointer dereference [-Wnull-dereference]
 8835 |     cb->use_clipping = (int)clip;
      |     ~~~~~~~~~~~~~~~~~^~~~~~~~~~~
../../../../../../include/nuklear_xlib/../nuklear.h:8834:14: warning: potential null pointer dereference [-Wnull-dereference]
 8834 |     cb->base = b;
      |     ~~~~~~~~~^~~
In file included from ../../../../../../include/nuklear_xlib/hb_nuklear_xlib.h:24,
                 from ../../../../../../src/nuklear.c:34:
../../../../../../include/nuklear_xlib/../nuklear.h:20749:23: warning: potential null pointer dereference [-Wnull-dereference]
20749 |         popup->parent = win;
      |         ~~~~~~~~~~~~~~^~~~~
../../../../../../include/nuklear_xlib/../nuklear.h: In function ‘nk_popup_begin’:
../../../../../../include/nuklear_xlib/../nuklear.h:20657:23: warning: potential null pointer dereference [-Wnull-dereference]
20657 |         popup->parent = win;
      |         ~~~~~~~~~~~~~~^~~~~
../../../../../../include/nuklear_xlib/../nuklear.h: In function ‘nk_group_set_scroll’:
../../../../../../include/nuklear_xlib/../nuklear.h:22759:19: warning: potential null pointer dereference [-Wnull-dereference]
22759 |     *y_offset_ptr = y_offset;
[...]

PR #584, opened 2023-10-30
Fix null deref with y_offset in nk_group and nk_listview

mtijanic:
We hit a rare null deref on `y_offset` in `nk_group_scrolled_offset_begin()`, that I think happens like this:
```C
    // snippet from nk_group_begin_titled()

    x_offset = nk_find_value(win, id_hash);
    if (!x_offset) {
        x_offset = nk_add_value(ctx, win, id_hash, 0);
        y_offset = nk_add_value(ctx, win, id_hash+1, 0);
        NK_ASSERT(x_offset);
        NK_ASSERT(y_offset);
        if (!x_offset || !y_offset) return 0;
        *x_offset = *y_offset = 0;
    } else y_offset = nk_find_value(win, id_hash+1);
    return nk_group_scrolled_offset_begin(ctx, x_offset, y_offset, title, flags);
```
First, we don't find the `x_offset`, so we go into the if() branch. There, we manage to add `x_offset` but not `y_offset`. This causes it to bail early. Then, next frame, it will find `x_offset` and go into the else branch. There, it fails to find `y_offset`, and eventually calls into `nk_group_scrolled_offset_begin()` with `y_offset = NULL`.

Never got a local repro so can't say for sure if the existing `NK_ASSERT(y_offset)` was firing, but end user reports that this patch fixes it.

I think this is one of the issues that was reported in #513

### DIFF
--- a/nuklear.h
+++ b/nuklear.h
@@ -22877,7 +22877,15 @@ nk_group_get_scroll(struct nk_context *ctx, const char *id, nk_uint *x_offset, n
         NK_ASSERT(y_offset_ptr);
         if (!x_offset_ptr || !y_offset_ptr) return;
         *x_offset_ptr = *y_offset_ptr = 0;
-    } else y_offset_ptr = nk_find_value(win, id_hash+1);
+    } else {
+        y_offset_ptr = nk_find_value(win, id_hash+1);
+        if (!y_offset_ptr) {
+            y_offset_ptr = nk_add_value(ctx, win, id_hash+1, 0);
+            NK_ASSERT(y_offset_ptr);
+            if (!y_offset_ptr) return;
+            *y_offset_ptr = 0;
+        }
+    }
     if (x_offset)
       *x_offset = *x_offset_ptr;
     if (y_offset)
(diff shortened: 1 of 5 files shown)


nuklear.h
```diff
@@ -22827,7 +22827,15 @@ nk_group_begin_titled(struct nk_context *ctx, const char *id,
...
-    } else y_offset = nk_find_value(win, id_hash+1);
+    } else {
+        y_offset = nk_find_value(win, id_hash+1);
+        if (!y_offset) {
+            y_offset = nk_add_value(ctx, win, id_hash+1, 0);
+            NK_ASSERT(y_offset);
+            if (!y_offset) return 0;
+            *y_offset = 0;
+        }
+    }
```
RobLoach, line 22838: Good find, thanks a lot. Two things...

1. Make sure to update the change in the `src` directory too. Otherwise the next time the paq.sh file is run, we'll lose this change.
2. Also, what do you think about updating this pattern in both `nuklear_group.c` and `nuklear_list_view.c` too? Could probably run into the same issue there, as it's using a similar pattern.

mtijanic: 1. Ah, right, sorry! I just ported the change from the vendored file in my project, I didn't read up on how development actually goes here. Sorry. Do I need to _also_ edit `nuklear.h` or will that be handled automatically when you pack it up?
2. Yep, seems similar. Do you want the same change there as well, or do you have some nicer way to generalize it in mind?

Thanks!

RobLoach: 1. No need to appologize. Can be confusing. You don't absolutely need to edit nuklear.h. Anyone can run the script and it'll get updated eventually. Whichever is easier for you
3. Same way is fine :shrug:

mtijanic: Done. I left it as smaller commits, but let me know if you prefer a squash. Or just squash when merging.

PR #586, opened 2023-11-14
Fix button text alignment error.

mackron:
When calculating the width and height of the content rect of a button, only the border of one side is being taken into account. Instead the border width needs to be multiplied by two before subtracting from the bounds. This is being done for padding and rounding, but not border, which is resulting in text being misaligned. The screenshots below demonstrate the issue (I've exaggerated the borders to make it easier to see the effect).

Before:
image

After:
image

### DIFF
--- a/clib.json
+++ b/clib.json
@@ -1,6 +1,6 @@
 {
   "name": "nuklear",
-  "version": "4.10.6",
+  "version": "4.10.7",
   "repo": "Immediate-Mode-UI/Nuklear",
   "description": "A small ANSI C gui toolkit",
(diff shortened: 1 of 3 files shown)


riri (changes requested):
LGTM

Be sure to rebase from master to get the latest version then bump the patch version in clib.json.

mackron:
Cheers. clib.json updated. Was already up to date with master.

riri (approved):
Ok for me, I let another maintainer reviewing for merging.

RobLoach:
Good catch.

PR #592, opened 2023-11-26
Add Text Alignment Options to Checkboxes and Radios

yukyduky:
This PR adds alignment to check boxes and radio buttons. Originally I was annoyed at not being able to set what side the text appears on and how much space there was between the widget and the text without making a separate label widget and set out to add alignment to all widgets. I then quickly realized that most widgets take up all of their allocated space so it doesn't apply to them in the same way. If there is any widget that could use this that I have missed, please let me know :)

Anyway, here is how it looks:

image

### DIFF
--- a/nuklear.h
+++ b/nuklear.h
@@ -3205,23 +3205,23 @@ NK_API nk_bool nk_button_pop_behavior(struct nk_context*);
  *                                  CHECKBOX
  *
  * ============================================================================= */
-NK_API nk_bool nk_check_label(struct nk_context*, const char*, nk_bool active, nk_flags alignment);
-NK_API nk_bool nk_check_text(struct nk_context*, const char*, int, nk_bool active, nk_flags alignment);
-NK_API unsigned nk_check_flags_label(struct nk_context*, const char*, unsigned int flags, unsigned int value, nk_flags alignment);
-NK_API unsigned nk_check_flags_text(struct nk_context*, const char*, int, unsigned int flags, unsigned int value, nk_flags alignment);
-NK_API nk_bool nk_checkbox_label(struct nk_context*, const char*, nk_bool *active, nk_flags alignment);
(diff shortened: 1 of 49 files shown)


yukyduky:
Should be good to go :)

riri:
Instead of doing an API breaking change, I would implement functions with extra parameters, and make the original ones using the new functions with the default alignment (left) as argument. This avoiding a major version change and keeping existing code base working.

yukyduky:
Alright, I was a bit skeptical if it would be accepted as is^^ I can fix that tomorrow :)

yukyduky:
I was also going to ask if I should've consolidated the alignment enums into one for both text and widgets but I'll take that as a no since that would also become a breaking change?

RobLoach:
Oh, this is a great addition. Haven't tested yet, or read through the code, but it's looking great. Anything else you'd like to get in for it?

RobLoach:
Anything else you'd like to add to this Pull Request. Seems feature-complete, and other widgets could be in follow ups. Your call.

yukyduky:
I think this will do for this PR. I would add more widgets, but I am unsure how they should act when aligned because all widgets I can think of take up all of the allocated space. If I come up with a way to do it then I will submit another PR :)

riri:
> Fixed the API break.

I recheck asap.

@RobLoach I second voice to keep it only with actual features in this PR, to enable a faster review and merge, unless you see that enabling to other widgets in the future would break the new API for alignment?

yukyduky:
Fixed the issue and also set my IDE to use spaces for tabs from now on :)

nuklear.h
riri, line 1: You didn't rebuild nuklear.h from sources

yukyduky:
Whoops! Fixed.

riri (approved):
It seems good for merging. Needing another reviewer before merging.

Thanks for that contribution :)

PR #595, opened 2023-12-19
Fix undefined behavior in `strncpy()` call

LocalSpook:
`strncpy()`'s pointer parameters are marked `restrict`, but when `file_browser_reload_directory_content()` was being called to reload the directory that was currently open:
```c
file_browser_reload_directory_content(browser, browser->directory)
```
it would try to copy its path buffer to itself, causing undefined behavior. The solution uses `memmove()`, which can handle overlapping pointers.

Found with Address Sanitizer.

### DIFF
--- a/example/file_browser.c
+++ b/example/file_browser.c
@@ -346,7 +346,7 @@ static void
 file_browser_reload_directory_content(struct file_browser *browser, const char *path)
 {
     const size_t path_len = strlen(path) + 1;
-    memmove(browser->directory, path, MIN(path_len, MAX_PATH_LEN));
+    NK_MEMCPY(browser->directory, path, MIN(path_len, MAX_PATH_LEN));
     browser->directory[MAX_PATH_LEN - 1] = 0;
     dir_free_list(browser->files, browser->file_count);
     dir_free_list(browser->directories, browser->dir_count);
--- a/example/file_browser.c
+++ b/example/file_browser.c
@@ -345,7 +345,8 @@ media_init(struct media *media)
 static void
 file_browser_reload_directory_content(struct file_browser *browser, const char *path)
 {
-    strncpy(browser->directory, path, MAX_PATH_LEN);
+    const size_t path_len = strlen(path) + 1;
+    memmove(browser->directory, path, MIN(path_len, MAX_PATH_LEN));
     browser->directory[MAX_PATH_LEN - 1] = 0;
     dir_free_list(browser->files, browser->file_count);
     dir_free_list(browser->directories, browser->dir_count);
--- a/example/file_browser.c
+++ b/example/file_browser.c
@@ -345,7 +345,7 @@ media_init(struct media *media)
 static void
 file_browser_reload_directory_content(struct file_browser *browser, const char *path)
 {
-    const size_t path_len = strlen(path) + 1;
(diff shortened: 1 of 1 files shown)


example/file_browser.c
```diff
@@ -345,7 +345,8 @@ media_init(struct media *media)
 static void
 file_browser_reload_directory_content(struct file_browser *browser, const char *path)
 {
-    strncpy(browser->directory, path, MAX_PATH_LEN);
+    const size_t path_len = strlen(path) + 1;
+    memmove(browser->directory, path, MIN(path_len, MAX_PATH_LEN));
```
RobLoach, line 349: Would a `strcpy()` be better here? Or explicitly calling it like... `strncpy(browser->directory, path, strlen(path));`?

LocalSpook: Both `strcpy()` (https://en.cppreference.com/w/c/string/byte/strcpy) and `strncpy()` (https://en.cppreference.com/w/c/string/byte/strncpy) have `restrict` pointer parameters, so both would trigger undefined behavior.

RobLoach (approved):

example/file_browser.c
RobLoach: Thanks for the explaination. Had a look through the rest of Nuklear, and it looks like rather than using memmove directly, uses `NK_MEMCPY`, which could make sense here...
``` c
        /* memmove */
        char *dst = nk_ptr_add(char, s->buffer.memory.ptr, pos);
        char *src = nk_ptr_add(char, s->buffer.memory.ptr, pos + len);
        NK_MEMCPY(dst, src, s->buffer.allocated - (nk_size)(pos + len));
```

https://cplusplus.com/reference/cstring/memcpy/

RobLoach, line 349:
``` suggestion
    NK_MEMCPY(browser->directory, path, MIN(path_len, MAX_PATH_LEN));
```

RobLoach (approved):
With using nk_memcpy, I think this should be okay.

example/file_browser.c
```diff
@@ -345,7 +345,8 @@ media_init(struct media *media)
 static void
 file_browser_reload_directory_content(struct file_browser *browser, const char *path)
 {
-    strncpy(browser->directory, path, MAX_PATH_LEN);
+    const size_t path_len = strlen(path) + 1;
```
RobLoach, line 348:
``` suggestion
    const size_t path_len = nk_strlen(path) + 1;
```

otonoton:
> With using nk_memcpy, I think this should be okay.

@RobLoach If you are using the system memcpy() then it does not handle overlapping pointers... why wasn't memmove kept here?

There's also other places I have found that need memmove() for the same reason, such as:

https://github.com/Immediate-Mode-UI/Nuklear/blob/master/src/nuklear_buffer.c#L128

https://github.com/Immediate-Mode-UI/Nuklear/blob/master/src/nuklear_string.c#L144

https://github.com/Immediate-Mode-UI/Nuklear/blob/master/src/nuklear_string.c#L283

I don't think the user is going to know they shouldn't define NK_MEMCPY to the system memcpy().

I propose either replacing these instances with the real memmove(), or renaming the NK_MEMCPY define to NK_MEMMOVE and updating the documentation to not suggest using the system memcpy() for it.

PR #597, opened 2023-12-20
Fix spelling mistakes

waywardmonkeys:


RobLoach (approved):
Looks like some minor updates to spelling, no code changes.

stb_rect_pack.h is a third-party file. May want to submit those changes over at https://github.com/nothings/stb/blob/master/stb_rect_pack.h .

PR #602, opened 2024-01-05
Add a style option to disable point markers on charts

### LINKED ISSUE #594: Line chart performance: boxes on points make line charts much slower than without
By default line charts include a box on each point as can be seen in the image below. I've found these boxes drastically reduce performance compared to if I remove them - on my machine the frametime is 74ms with the boxes and locked to 16ms/vsync without (sorry, I don't have an easy way to disable vsync to get better resolution data).
image

I've debugged a bit in renderdoc - there's a single draw call for the charts with or without the boxes. I can see there's 111612 elements in the vkCmdDrawIndexed call with boxes vs only 19542 elements without boxes - that's expected given the extra geometry for the boxes. I'll run this through RGP to see if I can find any hints on exactly what's going on.

I've fixed this for myself by simply removing the boxes [in my own fork](https://github.com/brianwatling/Nuklear/tree/4.10.7-no-boxes-on-line-charts), but wanted to see if there's interest to make the boxes optional.

Some more context: 
GPU: RX6650 XT
Renderer: Vulkan (logic to convert NK commands to draw commands based on demo/sdl_opengl3/nuklear_sdl_gl3.h)
Number of points in the chart: 3072 (3 lines, 1024 each)

brianwatling:
With point markers enabled (default):
with_markers
With point markers disabled:
without_markers

Fixes #594

### DIFF
--- a/demo/common/overview.c
+++ b/demo/common/overview.c
@@ -621,6 +621,7 @@ overview(struct nk_context *ctx)
             float id = 0;
             static int col_index = -1;
             static int line_index = -1;
+            static int hide_markers = nk_false;
             float step = (2*3.141592654f) / 32;
 
             int i;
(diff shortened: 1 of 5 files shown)


RobLoach (approved):
Seems good to me! Nice work.

dumblob:
Looks good to me in general. My personal preference would be to use a non-negated form (`nk_bool show_markers;` instead of `nk_bool hide_markers;`) but I would be fine merging if others will not agree :wink:.

RobLoach:
That's a good point. Would it be possible to default `nk_bool show_markers` to `nk_true` when initializing the struct?

brianwatling:
> That's a good point. Would it be possible to default `nk_bool show_markers` to `nk_true` when initializing the struct?

Done - renamed to show_markers and defaults to nk_true to preserve existing default behaviour (and tested that)

RobLoach (approved):
Looks solid. Thanks so much.

PR #604, opened 2024-01-15
Fixed declaration after statement in overview.c .

b-aaz:


riri:
No need for more reviews here, I thought of doing it myself but hadn't time.

Thank you for the fix @b-aaz

PR #605, opened 2024-01-20
Added missing includes to the platform headers .

b-aaz:
Added missing include decelerations and in one case removed an unnecessary one form the platform headers in the `demo/` directory .


riri (approved):
LGTM

Issue #608, opened 2024-02-05
Inconsistent grabbing behavior in demos (click and drag property)


paccerdk:
There seems to be an issue with the implementation of grabbing behavior across several demos, particularly when this behavior is combined with relative movement (dragging an input value).

The problem primarily arises due to how the grab state (`nk_context` `input.mouse.grab/ungrab`) is always reset, even if no event happens, leading to grab/ungrab potentially being missed in event handling, especially noticeable with frameworks like SDL.

#### Issue Details

The issue occurs with the following sequence of events:
  1. **Main Loop, Cycle 1:** A mouse down event is registered, invoking `nk_input_button` which in turn causes `nk_property_*` to set the grab state to `1` (`nk_true`) later in main loop.
  2. **Main Loop, Cycle 2:** No new event is registered (mouse remains pressed), causing the event loop to be skipped, then `nk_input_end` is called, which resets the grab state to `0` without being handled.

Similarly, The same problem happens with the ungrab state

#### Impact Details
- This is a problem for frameworks that do not automatically generate an event following a mouse down/up action. For instance, the SDL implementation requires two consecutive events for proper grabbing behavior. A scenario where the mouse is pressed down and then moved with more than 1 cycle in-between fails to maintain the grab/ungrab state, unlike when multiple events happen, such as a mouse move together with mouse down event.

- Its an even bigger problem when a grab event gets handled, but the ungrab event doesn't, causing the mouse cursor to disappear permanently and staying in relative mode with the SDL demos

#### Solution
I wouldn't mind providing a pull request, but i'm not sure about the ideal way to implement the solution, 
maybe someone more familiar with the framework might be able to suggest a more integrated solution than mine?

A simple way of solving it would be something like the following:
```
bool grab_handled = false;
while (running) {
	/* Input */
	SDL_Event evt;
	nk_input_begin(ctx);
	while (SDL_PollEvent(&evt)) {
		if (evt.type == SDL_QUIT)
			goto cleanup;
		nk_sdl_handle_event(&evt);
		grab_handled = true;
	}

	if (grab_handled) {
		nk_input_end(ctx);
		grab_handled = false;
	}
	...
```
This approach only resets grab / ungrab when its been handled.

I believe the following issues are related:
These two has @vurtun mentioning that here is a problem:
https://github.com/vurtun/nuklear/issues/203
https://github.com/vurtun/nuklear/issues/631

[...]

PR #609, opened 2024-02-10
Fix missed grab state changes in SDL demos

paccerdk:
Moved grab handling out of `nk_sdl_handle_event` to `nk_sdl_handle_grab`, which is now called outside of the event loop.

This change makes the logic similar to the GLFW demos and fixes issues with missed grab state changes and mouse cursor disappearing.

Discussed in https://github.com/Immediate-Mode-UI/Nuklear/issues/608

### DIFF
--- a/demo/sdl_opengl2/main.c
+++ b/demo/sdl_opengl2/main.c
@@ -139,6 +139,7 @@ main(int argc, char *argv[])
             if (evt.type == SDL_QUIT) goto cleanup;
             nk_sdl_handle_event(&evt);
         }
+        nk_sdl_handle_grab(); /* optional grabbing behavior */
         nk_input_end(ctx);
 
         /* GUI */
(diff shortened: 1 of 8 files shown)


paccerdk:
Reverted the  order of SDL_SetRelativeMouseMode and SDL_WarpMouseInWindow since it was causing the cursor to warp to center of screen on older SDL versions (tested with SDL 2.0.22)

dumblob:
I like the PR, thanks @paccerdk for spending the time with all the investigations across backends!

> Reverted the order of SDL_SetRelativeMouseMode and SDL_WarpMouseInWindow since it was causing the cursor to warp to center of screen on older SDL versions (tested with SDL 2.0.22)

Could you please add a comment to the code about this finding? I find it important to have it documented. Thanks!

RobLoach:
I tested `sdl_opengl3` and mouse drag is still looking good :+1:

PR #613, opened 2024-02-21
Add example widget using an nk_image to the sdl_renderer demo

zoogies:
This PR is intended to help illustrate the creation of `nk_image` from `SDL_Surface` and `SDL_Texture`.

A `nk_button_image_label` has been added to the sdl_renderer demo which looks like this:

Screenshot_20240221_143348

The icon is loaded from small bmp image that has been added to the `example/icon` directory, since SDL can only load bmp images without linking against something like SDL_Image.

Some things potentially worth considering:

- should the relative pathing to the bmp in `example/icon` be handled differently?
- would it be better to eliminate the added image by instead using an existing png and linking against SDL_Image?

When I first started using Nuklear, an example like this would have been of great assistance to me. 

Cheers! 😄


RobLoach:
Thanks, I'm sure this will be helpful for others looking to bring SDL images in. While we will likely want to keep all the examples the same, I'm sure people would appreciate if you added some notes about SDL images to the wiki:
https://github.com/Immediate-Mode-UI/Nuklear/wiki/SDL-%28OpenGL-3%29

RobLoach:
I've pushed up your code over at https://github.com/Immediate-Mode-UI/Nuklear/wiki/SDL . Thanks for the demonstration! Feel free to edit it if you think it could do with some tweaks.

PR #618, opened 2024-03-07
Fix bitwise operations warnings in C++20

### LINKED ISSUE #170: Compile error: C++20 strict
The following is deemed as deprecated by the C++20 standard;
Having an enum that's used for initialization in another enum by bitwise operation.
This:
NK_EDIT_SIMPLE | (result from nk_edit_flags or)
aka nk_edit_types | nk_edit_flags.
Is now not legal anymore in W3 Wall. 
I'd say that the solution is for nk_edit_types and nk_edit_flags to be into one enum or explicitly cast.
In my own branch I opted for the first approach.

dav64:
This PR is intended to fix multiple warnings 
`warning: bitwise operation between different enumeration types XXX and YYYY is deprecated `
raised when compiling in C++20

Each time I updated Nuklear on my side I did those changes, today I decided to make them once and for all with a PR :smile: 

Fixes #170

### DIFF
--- a/nuklear.h
+++ b/nuklear.h
@@ -30123,7 +30123,7 @@ nk_tooltipfv(struct nk_context *ctx, const char *fmt, va_list args)
 ///   - [y]: Minor version with non-breaking API and library changes
 ///   - [z]: Patch version with no direct changes to the API
 ///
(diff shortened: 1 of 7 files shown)


RobLoach (approved):
While it does look like some bloat, I think this is a fine addition. Thanks a lot. Anyone else care to review?

PR #636, opened 2024-04-27
Proposal: Update nuklear_sdl_renderer.h to use recommended path for SDL.h

Ryder17z:
change SDL include path to the recommended path

### DIFF
--- a/demo/sdl_renderer/nuklear_sdl_renderer.h
+++ b/demo/sdl_renderer/nuklear_sdl_renderer.h
@@ -11,7 +11,7 @@
 #ifndef NK_SDL_RENDERER_H_
 #define NK_SDL_RENDERER_H_
 
-#include <SDL2/SDL.h>
+#include <SDL.h>
 NK_API struct nk_context*   nk_sdl_init(SDL_Window *win, SDL_Renderer *renderer);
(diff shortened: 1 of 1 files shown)


RobLoach (approved):
I'm okay with this change, though would like a review from others. The other thing we could consider is something like...

``` c
#ifndef NK_SDL_RENDERER_SDL_H
#define NK_SDL_RENDERER_SDL_H <SDL.h>
#endif
#include NK_SDL_RENDERER_SDL_H
```

Ryder17z:
Sounds good to me.

demo/sdl_renderer/nuklear_sdl_renderer.h
```diff
@@ -11,7 +11,7 @@
 #ifndef NK_SDL_RENDERER_H_
 #define NK_SDL_RENDERER_H_
 
-#include <SDL2/SDL.h>
+#include <SDL.h>
```
RobLoach, line 14:
``` suggestion
#ifndef NK_SDL_RENDERER_SDL_H
#define NK_SDL_RENDERER_SDL_H <SDL.h>
#endif
#include NK_SDL_RENDERER_SDL_H
```

PR #628, opened 2024-04-02
Add delta_time_seconds to a few renderers

RobLoach:
This change adds `delta_time_seconds` to a few of the renderers. Would love some help on supporting some of the others.

- [x] allegro5
- [ ] d3d11
- [ ] d3d12
- [ ] d3d9
- [ ] gdi
- [ ] gdi_native_nuklear
- [ ] gdip
- [x] glfw_opengl2
- [x] glfw_opengl3
- [x] glfw_opengl4
- [x] glfw_vulkan
- [x] sdl2surface_rawfb
- [x] sdl_opengl2
- [x] sdl_opengl3
- [x] sdl_opengles2
- [x] sdl_renderer
- [x] sfml_opengl2
- [x] sfml_opengl3
- [ ] wayland_rawfb
- [x] x11
- [x] x11_opengl2
- [x] x11_opengl3
- [ ] x11_rawfb
- [ ] wayland rawfb
- [ ] sdl rawfb
- [x] x11_xft

Fixes #627

PR #639, opened 2024-05-09
Key repeat for nk_keys

PROP65:
Adds key repeat for nk_keys (Arrows, Home, End, Backspace, Delete, etc).

The default delay and interval can be changed, as well as updated at runtime.
```c
/* send 20 inputs per second after a key has been held for 1/3 of a second */
#define NK_INPUT_REPEATER_DELAY 0.33f
#define NK_INPUT_REPEATER_INTERVAL 0.05f
#define NK_KEY_REPEAT
#include "nuklear.h"

...

struct nk_context ctx;
nk_init_default(&ctx, 0);
/* update at runtime, 10 inputs per second after 0.4 seconds */
ctx.input.keyboard.repeater_delay    = 0.40f;
ctx.input.keyboard.repeater_interval = 0.10f;
```
Tested and working with the following render backends.

- `glfw_opengl2`
- `glfw_opengl3`
- `glfw_vulkan`


RobLoach:
Nice work! Looking good... Adding delta time to the renders will help too https://github.com/Immediate-Mode-UI/Nuklear/pull/628

src/nuklear.h
```diff
@@ -4516,12 +4516,26 @@ struct nk_mouse {
...
+#endif
+
 struct nk_keyboard {
     struct nk_key keys[NK_KEY_MAX];
     char text[NK_INPUT_MAX];
     int text_len;
+    float repeater_delay;
+    float repeater_interval;
+    /* copied from nk_context.delta_time_seconds on nk_end_input call */
+    float delta;
```
RobLoach, line 4538: Rather than copying the data each frame, and requiring the `delta_time_seconds` to be set within `nk_begin/end`, perhaps we have a `struct nk_context* ctx` instead?

PROP65:
I've reworked this patch to make the key repeat functionality more generic/reusable.
Most checks/calculations have been moved to `nuklear_input.c` and a new function is provided (`nk_input_is_key_fired`) to check the new boolean.

After submitting this pull request I realized some backends have key repeat via their abstraction layer (SDL, X11, etc) so this functionality is also now opt-in and can be enabled by defining NK_KEY_REPEAT before including Nuklear
```c
...
#define NK_KEY_REPEAT
#include "nuklear.h"
...
```

PR #642, opened 2024-05-21
Add `const` qualifiers to `nk_allocator` in more places.

SnootierMoon:
The `nk_allocator` struct consists of 3 fields, `userdata` (application-specified ptr), `alloc` (allocate function), and `free` (deallocate function). None of these are modified once provided by the library user as far as I can tell, so it should be safe to make this struct `const` in all API parameters.

Since stb-truetype uses a `void*` for allocation data, a cast is necessary. But, since the `nk_stbtt_malloc`/`nk_stbtt_free` implementations don't modify the `nk_allocator` (as the potentially mutable allocator is stored indirectly in `nk_allocator.userdata`), this should be fine as well.


RobLoach (approved):
Looks good to me.

PR #651, opened 2024-06-17
Fix rounding for window and border 

SchelZ:
wrong branch


RobLoach:
Looks liket his is the right PR https://github.com/Immediate-Mode-UI/Nuklear/pull/652

PR #654, opened 2024-06-27
Add triangle outline symbols

RobLoach:
This change adds triangle outline symbols.

Screenshot from 2024-06-27 14-16-35

### DIFF
--- a/demo/sdl_renderer/main.c
+++ b/demo/sdl_renderer/main.c
@@ -39,7 +39,7 @@
 /*#define INCLUDE_CANVAS */
 /*#define INCLUDE_OVERVIEW */
 /*#define INCLUDE_NODE_EDITOR */
-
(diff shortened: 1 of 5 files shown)


demo/sdl_renderer/main.c
```diff
@@ -39,7 +39,6 @@
 /*#define INCLUDE_CANVAS */
 /*#define INCLUDE_OVERVIEW */
 /*#define INCLUDE_NODE_EDITOR */
-
```
RobLoach, line 42:
``` suggestion

```

src/nuklear_button.c
```diff
@@ -6,6 +6,7 @@
  *                          BUTTON
  *
  * ===============================================================*/
+#include <stdio.h>
```
RobLoach, line 9:
``` suggestion
```

PR #661, opened 2024-07-11
Quickfix glfw shift-key handling

bolt-blue:
All of the glfw demo headers unset shift keypresses when the ctrl key is not also down.
I can't think of a reason this would be desired and it appears no other demo code does this.
Unless I'm mistaken, simply removing these lines leaves us with the desired behaviour.


RobLoach (approved):
While I don't often use the glfw runtime, this makes sense to me. Anyone else able to review? Thanks!

PR #662, opened 2024-07-18
Improve one poor case of indentation

bolt-blue:
Not at all a critical change request, but I'll leave it here for consideration when the time arrises.

The original code was distinctly unclear. I've tried to keep to the surrounding code style while making the code flow more comprehensible.

I've not looked elsewhere for any other similar occurrences; this became apparent while investigating a specific line of thought.


bolt-blue:
Maybe if the change is deemed meaningful, it would first be worth going through correcting any other similar cases.

PR #664, opened 2024-07-28
add RGFW examples

ColleagueRiley:
This adds RGFW examples to nuklear including,

rgfw_opengl2, rgfw_opengl3, and rgfw_opengl4 as well as an rgfw example for rawfb.

(I think it would be better if Nuklear separated the platform API from the graphics API, similar to Dear ImGUI)

### DIFF
--- a/.github/workflows/ccpp.yml
+++ b/.github/workflows/ccpp.yml
@@ -23,6 +23,14 @@ jobs:
       run: make -C demo/glfw_opengl4
 #    - name: build glfw_vulkan
 #      run: make -C demo/glfw_vulkan
+    - name: build rgfw_opengl2
+      run: make -C demo/rgfw_opengl2
+    - name: build rgfw_opengl3
+      run: make -C demo/rgfw_opengl3
+    - name: build rgfw_opengl4
+      run: make -C demo/rgfw_opengl4
+    - name: build rgfw_rawfb
+      run: make -C demo/rawfb/rgfw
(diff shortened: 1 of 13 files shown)


ColleagueRiley:
@RobLoach 
This is a problem for each platform as far as I can tell. 

Dear ImGUI handles it with 
imgui_opengl.h and imgui_<platform>.h 

Where the graphics API platform can work with any windowing API platform. Assuming the platform has support for that graphics API.

ColleagueRiley:
@RobLoach I do not and never intended on implementing that change and I think Nuklear ought to have this change implemented as it is better for maintainablity and makes it easier to understand and create backends. This is exactly what libraries like Dear Imgui does and it would be very easy to implement with a fairly high reward.

RobLoach:
Had missed getting this going. Thanks for the ping. We got some consolidation around some of the examples, but I don't think it should block this change.

Merging forwards 👍

Edit: Ahh, conflicts.

ColleagueRiley:
This version of RGFW is so out of date that it's not even relevant anymore. I host a newer version of the nuclear code on a separate repo. 

I don't know if there is any particular reason for me to put that here, I guess I can leave that up to someone else.

RobLoach:
Nice work. I can see it's linked in the [Integrations](https://github.com/Immediate-Mode-UI/Nuklear/wiki/Integrations) page too :+1: 

Was able to run all of the examples without issue.

PR #666, opened 2024-08-14
RFC: Inline text instructions

PROP65:
An alternative method to achieve something similar to #560

The idea is to use the UTF-8 [Private Use Area](https://en.wikipedia.org/wiki/Private_Use_Areas) ( U+E000 -> U+F8FF ) to pass instructions to the text renderer in a manner similar to ANSI escape codes.

I've implemented hex RGB (`\uE000`), hex RGBA (`\uE001`) and color reset (`\uE002`) instructions.

Each instruction has a hardcoded value(payload) length.
RGB expects 6 ascii bytes containing a color in hex format
while color reset expects no payload

Backends using `nk_convert` will only require minimal, if any, modification to take advantage of this feature while any other backends will need to implement the instructions for them to function.  A function is provided to filter  out the inlined instructions along with their payloads (`nk_utf_filter_instructions`) for use in copy/cut/paste operations. Selecting text in `nk_edit_string` (including `NK_EDIT_BOX`) has been adjusted and copying text works as intended. Cut and paste have not been tested on text containing these instructions.

```c
nk_label(ctx,
	NK_INSTRUCT_SET_RGB  "FF0000"   "N"
	NK_INSTRUCT_SET_RGB  "00FF00"   "u"
	NK_INSTRUCT_SET_RGB  "0000FF"   "k"
	NK_INSTRUCT_SET_RGB  "00FFFF"   "l"
	NK_INSTRUCT_SET_RGB  "FF00FF"   "e"
	NK_INSTRUCT_SET_RGB  "FFFF00"   "a"
	NK_INSTRUCT_SET_RGBA "FFFFFF88" "r"
	NK_INSTRUCT_RESET_COLOR, NK_TEXT_LEFT);
```
looks like

glfw_opengl3

The following backends have been tested

- [x] glfw_opengl3

### DIFF
--- a/src/nuklear_utf8.c
+++ b/src/nuklear_utf8.c
@@ -19,6 +19,7 @@ nk_utf_validate(nk_rune *u, int i)
     if (!NK_BETWEEN(*u, nk_utfmin[i], nk_utfmax[i]) ||
          NK_BETWEEN(*u, 0xD800, 0xDFFF))
             *u = NK_UTF_INVALID;
+
     for (i = 1; *u > nk_utfmax[i]; ++i);
     return i;
 }
@@ -58,6 +59,7 @@ nk_utf_decode(const char *c, nk_rune *u, int clen)
     if (j < len)
         return 0;
     *u = udecoded;
+
     nk_utf_validate(u, len);
     return len;
 }
@@ -96,12 +98,21 @@ nk_utf_len(const char *str, int len)
 
     text = str;
     text_len = len;
-    glyph_len = nk_utf_decode(text, &unicode, text_len);
-    while (glyph_len && src_len < len) {
-        glyphs++;
-        src_len = src_len + glyph_len;
+    do {
         glyph_len = nk_utf_decode(text + src_len, &unicode, text_len - src_len);
-    }
+
+        if (NK_UTF_IS_INSTRUCTION(unicode)) {
+            int payload_size = nk_utf_instruction_payload_size(unicode);
+
+            /* invalid payload */
+            if(payload_size == -1) break;
+
+            glyph_len += payload_size;
+        }
+        glyphs++;
+
+        src_len = src_len + glyph_len;
+    } while (glyph_len && src_len < len);
     return glyphs;
 }
 NK_API const char*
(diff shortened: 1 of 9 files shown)


RobLoach:
This is really cool, thanks for putting it together. Would you provide a small example in the overview demo for this? It would be great way to showcase it.

PROP65:
glfw_vulkan

src/nuklear_utf8.c
```diff
@@ -137,8 +148,66 @@ nk_utf_at(const char *buffer, int length, int index,
...
+        /* check for inline instructions */
+        if (NK_UTF_IS_INSTRUCTION(unicode)) {
+            int payload_size = nk_utf_instruction_payload_size(unicode);
+
+            /* invalid payload */
+            if(payload_size == -1) break;
+
+            glyph_len += payload_size;
+        } else {
+            NK_MEMCPY(output, input + text_len, glyph_len);
```
RobLoach, line 205: There's a warning about the type conversion here. Casting to a size type should fix that.

``` suggestion
            NK_MEMCPY(output, input + text_len, (NK_SIZE_TYPE)glyph_len);
```

src/nuklear_utf8.c
```diff
@@ -137,8 +148,66 @@ nk_utf_at(const char *buffer, int length, int index,
...
+            output += glyph_len;
+        }
+
+        text_len += glyph_len;
+    } while(text_len <= len && glyph_len);
+
+    return output - start;
```
RobLoach, line 212: The function expects to return an `int`, but we're returning a char*-char*. May need a cast.

awschult002:
I decided to take a look at the last two X11 tasks for this PR and I am not sure how to go about solving it. I am learning UTF8 for the first time, and I don't have much experience with X11, but I still tried to give it my best shot. 

At the moment, X11 is not rendering the UTF8 encoded string at all. Or, at least, it renders up until the first character with a special color encoding; then it chokes. 

I have discovered that both of the final X11 demos do not render UTF8 at all, so I wouldn't really expect this to work even without the special color codes, yet I have not tested this. There is an `Xutf8DrawText()` function, but that one also chokes on our new UTF strings. 

But aside from that, setting the color for X seems to be "global" kind of thing; or at least, its expecting the entire string to be one color. So the only solution that I see is to process the UTF colors glyph by glyph; then set the FG color, then call XDraw, then move our X, Y, by the glyph width amount of the particular font, then get/set the color of the next glyph. 

not super efficient, but it would get us the color. In my attempts at getting that to work, I realized that I would need to turn on VERTEX_BUFFER_OUTPUT in order to get glyph width information and font->querying. But the comments say that is really for hardware rendering like OpenGL (makes sense to me); so i chose to abandon that attempt and just get something working to pass this PR.

So can we just print the string without color and anyone using the X11 basic render is just out of luck? Well, sort of. Using the new `nk_filter_instructions` function, I was able to get the XDraw function to print out the entire string. This was a great success. The only issue is that XDraw doesn't have the rectangle bounds of the window; so it doesn't line wrap the text at the boundary. I have no idea how to figure that out without knowing the glyph width and calculating what is needed to get to the window rect; then adding newlines or a second draw call. 

hopefully someone less ignorant that myself can come to swift solution with the above information.

RobLoach:
Tried this out on a few other renderers too, and it wasn't quite working correctly. Will have to investigate further.

PROP65:
Converted to draft. Text rendering state gets reset on line break and some backends have issue. This still needs quite a bit of work.

makefiler:
May I somehow use nk_font_config with x11 frontend? As I can see font rendering done using native lib... I tried set font using "-*-helvetica-bold-r-normal-*-*-120-*-*-*-*-koi8-r" mask, but still do not have Cyrillic letters. May ranges be set in x11 directly?

lockie:
@PROP65 would you need a hand with this PR? I'll be happy to offer help here.

PROP65:
> @PROP65 would you need a hand with this PR? I'll be happy to offer help here.

If you would like to try to implement this feature yourself feel free to use my code but it might be better to start from scratch, at least when it comes to the rendering code. I believe this needs to be implemented earlier in the text rendering process (before text wrapping). Some internal structs might need to be modified to pass extra data.

My original goal was to support "vertex buffer output" backends without modification. If a backend can't support this I would prefer that Nuklear functions as if the feature was absent.

I may come back to this PR in the future, but if you want to try to implement this I'll try to contribute when I can. I never got around to it, but I would also like to see Nuklear support blinking text (like ANSI "\e[5m"), and gradient colored text at some point.

PR #685, opened 2024-09-05
Add node editor nodes

RobLoach:
- **Renamed some things + added comments**
- **Added node_connector struct**
- **Node links now take node pointers instead of IDs**
- **Allow only one input per input slot**
- **Detachable/movable links**
- **Drawing of node as a separate function**
- **Color node creation added**
- **Node connector positioning**
- **Color coded slots + only allow link to same type**
- **Color node creation and display, base Node now generic**
- **Float and Output nodes, all mallocs into node_editor_add**
- **Node eval functions**
- **Comment clean up + add node_editor_eval_connected & node_output_get**
- **Node deletion**
- **node_editor_delete_link + C89 compliance**
- **Clean up warnings + C89 compliance**
- **Link buffer reuse / "memory management"**
- **Bug fix**
- **Made float node min/max 0.0/1.0 for simplicity**
- **Blend nodes added**
- **camelCase -> snake_case, as in the rest of Nuklear**
- **Renamed node_editor_main back to node_editor**
- **Minor cleanups**
- **Restore main.c from master**
- **Fix blend node out color inverted vs inputs**
- **Fix blend node out color inverted vs inputs**
- **Added link between Color and Output nodes in initial state of node editor demo**
- **demo: Add node editor example**

### DIFF
--- a/demo/common/nodeeditor/node_type_output.c
+++ b/demo/common/nodeeditor/node_type_output.c
@@ -5,8 +5,10 @@ struct node_type_output {
 
 struct nk_colorf *node_output_get(struct node* node) {
     struct node_type_output *outputnode = (struct node_type_output*)node;
-    if (!node->inputs[0].isConnected)
-        outputnode->inputVal = (struct nk_colorf){0.0f, 0.0f, 0.0f, 0.0f};
+    if (!node->inputs[0].isConnected) {
+        struct nk_colorf black = {0.0f, 0.0f, 0.0f, 0.0f};
+        outputnode->inputVal = black;
+    }
     return &outputnode->inputVal;
 }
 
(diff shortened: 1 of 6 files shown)


demo/common/nodeeditor/node_type_output.c
```diff
@@ -0,0 +1,31 @@
+struct node_type_output {
+    struct node node;
+    struct nk_colorf input_val;
```
RobLoach, line 3: `node_type_output` is more of a color output. Should we consider renaming it to `node_type_color` instead?

awschult002:
> @awschult002 How about something like this as an example starter...
> 
> Screenshot from 2024-09-05 11-26-53

This is way better than what I had in mind

Issue #59, opened 2020-02-03
How to get TAB to navigate between edit widgets?


ronaaron:
I've got a layout with labels on the left and edit controls on the right. Everything works nicely, but I want to be able to hit TAB (for instance) to move from one 'field' to the next.

What is the proper way to accomplish this?

ronaaron:
OK, so just to say I do have a way to do it, but it seems like much more work than makes sense.

Check for KEY_TAB pressed in the render loop, and increment a counter; if the edit being created matches the counter, set the edit focus.  Some other mechanism would be much more friendly.

dumblob:
Thanks for asking and glad you already found the "recommended" solution.

>Some other mechanism would be much more friendly.

I'm not sure we would want something like that. There would need to be a **transparent** way how to determine which widget has widget in the current frame and which widget will get it after receiving Shift+Tab (backward) or just Tab (forward). Such thing is not easy and we would probably end up with just saying "the last rendered" widget for Shift+Tab and "the next to be rendered widget" for Tab. But that's still far from being transparent/unambiguous - imagine all the if-else logic for widget creation. Furthermore there are popups and stuff (messing with the logical "previous widget" and "next widget" transparency). This would get messy very early.

So I think the solution with an explicit variable (e.g. counter) is the best to my knowledge :wink:. Thoughts?

ronaaron:
The real difficulty is that because of immediate-mode, you don't really have knowledge of all the widgets which might exist during a frame, as you would with traditional GUIs.

That said, if there were a 'tabbing order' field, say, in each widget, where the programmer assigned that order (or it could alternatively be auto-assigned in widget-creation order), then it would be possible to handle Tab/Shift+Tab inside Nuklear and just give the focus to the next widget in the assigned order.

Possible?

dumblob:
>That said, if there were a 'tabbing order' field, say, in each widget, where the programmer assigned that order (or it could alternatively be auto-assigned in widget-creation order), then it would be possible to handle Tab/Shift+Tab inside Nuklear and just give the focus to the next widget in the assigned order.

That's what I described above as a problematic approach. But how would you handle (in a more readable and easier way than your current solution) all the cases where the programmer wouldn't want neither the default ordering and maybe not even the at-compile-time-known ordering (because that changes in accordance to which widgets will get drawn and which not in that particular frame based on user input or other criteria)?

[...]

PR #688, opened 2024-09-11
Fix nk_property not focusing when clicking mouse when the clicked property box is before the currently focused property box in the current window

xzn:
As titled.

Demonstration:

Current (notice when clicking the first box after the second one is focused, it immediately loses focus)
https://github.com/user-attachments/assets/a04e6fb8-e638-4b7f-b0d6-fee5c297a657

Fixed (this PR):
https://github.com/user-attachments/assets/e482c328-1bf3-4751-979c-6f7e63292b7a

Also ran paq.sh and added casts to match the changes in a previous PR (#682) to get rid of the type warnings introduced in that previous PR.

### DIFF
--- a/nuklear.h
+++ b/nuklear.h
@@ -6773,7 +6773,7 @@ nk_strtoi(const char *str, char **endptr)
         p++;
     }
     if (endptr)
-        *endptr = p;
+        *endptr = (char *)p;
     return neg*value;
 }
 NK_API double
@@ -6830,7 +6830,7 @@ nk_strtod(const char *str, char **endptr)
     }
     number = value * neg;
     if (endptr)
-        *endptr = p;
+        *endptr = (char *)p;
     return number;
 }
 NK_API float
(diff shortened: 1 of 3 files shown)


RobLoach:
Thanks for updating this. Should we bring this in and tackle the follow up afterwards? Or wait until the other case is covered.

xzn:
I can certainly make a follow up PR. To cover the other case I'll have to add some fields to the context struct so it remembers the previous edit variant. It's a bit more involved and you may want to review that separately. (If you can wait that's fine too, I can get to this later tonight)

xzn:
Fixed the second case as well. If you want I can add an example to one of the examples though it seems to work fine for now.

rswinkle:
@RobLoach @xzn

Hey I just rebased this on master, I really want to get this in as I also want to get something like #59 working. I can try force pushing it to xzn's branch (can I even do that? I didn't even add his as a remote, just grabbed the PR directly with gh). Or I could just submit a separate PR.

Would be really cool if we had a set of macros/functions specifically for tabbing through elements but I'd have to think about how that would work across backends, and before that this has to be in.

RobLoach:
Submitting a different PR may be easiest. Once that merged, this one will automatically close.

xzn:
Good to close I guess?

PR #690, opened 2024-09-17
allegro5: take image region into account for NK_COMMAND_IMAGE

lockie:
Hey. This pull request makes allegro5 backend take the image region into account for `NK_COMMAND_IMAGE` command. Without that, it is impossible to e.g. correctly render 9-slice, consider:

before:
2024-09-17-205052_277x244_scrot

after:
2024-09-17-205031_266x252_scrot

I suspect similar problem is present in gdi, gdi_native, gdip, rawfb, x11 and x11_xft backends, but unfortunately I lack the knowledge nor time to fix those.


lockie:
Took me a while to get it totally right, but everything is correct this time, I swear 😅

RobLoach:
That's one nice looking button :+1:

lockie:
Thanks ❤️

PR #693, opened 2024-09-19
Fix incompatible pointer types warning

RobLoach:


RobLoach:
@PROP65 Is this correct?

PROP65:
LGTM

PR #697, opened 2024-09-23
Update proggyfont urls

ryuukk:
Original website was compromised and is serving malware

Changed it for the author's github page instead


RobLoach (approved):
Thanks +1

PR #709, opened 2024-09-28
Round layout widget position to nearest 0

### LINKED ISSUE #708: Layout: Rounding issue
In the overview demo, in the "Simple" layout section, try to resize the window and notice sometimes the button gets clipped, sometimes not

image

image

ryuukk:
Fix #708


RobLoach:
Could certainly use this function elsewhere too. Great addition. Thanks.

PR #712, opened 2024-09-30
demo: Disable compiler optimization

RobLoach:
Using `-O2` across the demos enables some optimization, which is great, but results in slow compilation. This can be particularly slow when testing out some new components during development. I recommend diabling this to speed up compilation when testing.


awschult002:
if you havent heard of the "Tiny C Compiler" I cannot recommend it enough. Its the fastest compiler out there and its absolutely perfect for super quick development testing. I would not recommend it for release because its non-optimizing; but I use it all of the time for testing and its lightning fast.

https://github.com/TinyCC/tinycc

RobLoach:
Glad switching the `CC` environment variable worked well, and happy to hear that tinycc compiles it fine :+1: 

Updated the PR.

riri:
I just made a quick clone and try on sdl and x11 demos, I think it's trivial enough to merge without further tests

PR #714, opened 2024-09-30
internal: Allow changing the roundf() method

RobLoach:
This allows you to define `NK_ROUNDF()` to change the `nk_roundf()` implementation.

```c
#include <math.h>
#define NK_ROUNDF roundf
```

### DIFF
--- a/nuklear.h
+++ b/nuklear.h
@@ -5993,6 +5993,9 @@ NK_LIB float nk_atan(float x);
 #ifndef NK_ATAN2
 NK_LIB float nk_atan2(float y, float x);
 #endif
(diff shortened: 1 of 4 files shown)


nuklear.h
```diff
@@ -5993,6 +5993,9 @@ NK_LIB float nk_atan(float x);
 #ifndef NK_ATAN2
 NK_LIB float nk_atan2(float y, float x);
 #endif
+#ifndef NK_ROUNDF
+NK_LIB float nk_roundf(float x);
+#endif
```
RobLoach, line 5998: I think we should have the declaration, but change its usage in the definition.

PR #725, opened 2024-10-27
added support for combo from an array of struct

gooosedev:
This PR adds a new combo function ``nk_combo_from_struct_array``
the objective is to be able to make combos from an array of structs (the only requirement is for the struct to have a char* ptr to hold the name)

parameters are the same as for the ``nk_combo`` function, with the addition of the size of the struct and a stride to the 
member to use for the combo entry.

example:
```
struct mystruct{
  float f;
  char* name;
  int i;
};
struct mystruct items [3]={
  {1.0, "foo", 1},
  {1.0, "foobar", 1},
  {1.0, "bar", 1},
};
int selected;
/* ... */
selected = nk_combo_from_struct_array(ctx, &items, 3, sizeof(struct mystruct), offsetof(struct mystruct, name), selected, 15, nk_vec2(x,y));
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -42,7 +42,7 @@ DEMO_LIST = $(shell find $(DEMO_PATH) -type f -name Makefile -printf "%h ")
 usage:
 	echo "make docs		to create documentation"
 	echo "make nuke		to rebuild the single header nuklear.h from source"
-	echo "make demos	to build all of the demos
+	echo "make demos	to build all of the demos"
 	echo "make all 		to re-pack the header and create documentation"
 
 all: docs nuke demos 


RobLoach:
Thanks a lot. Would you mind adding a small example in the overview demo? It serves as a good space to test the code and make sure it's all still functional.

demo/common/overview.c
```diff
@@ -584,6 +584,36 @@ overview(struct nk_context *ctx)
...
                 }
+                {
+                    /* Combobox from array of struct */
+                    static int sel=0;
+                    struct student {
+                        int id;
+                        char* name;
+                        int age;
+                        char* major;
+                    };
```
RobLoach, line 595: May want to declare the struct outside of the code block.

demo/common/overview.c
```diff
@@ -584,6 +584,36 @@ overview(struct nk_context *ctx)
...
+
+                    static struct student students[4] = {
+                        {0, "----", 0, "----"},
+                        {1, "Mike", 20, "CS"},
+                        {2, "Jim", 19, "Maths"},
+                        {3, "Julia", 20, "Biology"}
+                    };
```
RobLoach, line 602: I believe for static definitions, those may want to be outside as well. In addition, using this format may make it easier to read...

```c
static struct student students[4] = {
  {
    .id = 0,
    .name = "---"
    .age = 0,
    .major = "---"
  },
  ...
```

In addition, what is the first `----` entry for?

nuklear.h
```diff
@@ -30504,6 +30505,41 @@ nk_combo_callback(struct nk_context *ctx, void(*item_getter)(void*, int, const c
         nk_combo_end(ctx);
     } return selected;
 }
+
+
+NK_API int nk_combo_from_struct_array(struct nk_context *ctx, const void *items, int count, int item_sz, int stride, int selected, int item_height, struct nk_vec2 size)
```
RobLoach, line 30510: Could you add some API documentation? Define each parameter, and give examples where possible. Thanks.

nuklear.h
```diff
@@ -30504,6 +30505,41 @@ nk_combo_callback(struct nk_context *ctx, void(*item_getter)(void*, int, const c
...
+    NK_ASSERT(ctx);
+    NK_ASSERT(items);
+    NK_ASSERT(ctx->current);
+    if (!ctx || !items ||!count)
+        return selected;
+
+    item_spacing = ctx->style.window.spacing;
+    window_padding = nk_panel_get_padding(&ctx->style, ctx->current->layout->type);
+    max_height = count * item_height + count * (int)item_spacing.y;
+    max_height += (int)item_spacing.y * 2 + (int)window_padding.y * 2;
```
RobLoach, line 30527: What is the `2`? Could this be grabbed from the nk style object?

src/nuklear_combo.c
```diff
@@ -819,6 +819,41 @@ nk_combo_callback(struct nk_context *ctx, void(*item_getter)(void*, int, const c
         nk_combo_end(ctx);
     } return selected;
 }
+
+
+NK_API int nk_combo_from_struct_array(struct nk_context *ctx, const void *items, int count, int item_sz, int stride, int selected, int item_height, struct nk_vec2 size)
```
RobLoach, line 824: Could you add some API documentation? Define each parameter, and give examples where possible. Thanks.

RobLoach (changes requested):
Not entirely sure about the function name. Do we have examples of this pattern elsewhere through Nuklear that we could copy the naming convention from?

demo/common/overview.c
awschult002: Declare or define? What would be the reasoning g for this? I have definitely seen structure definitions with local scope. 

But then again. I don't think that is a c89 capability.

awschult002:
Definitely in the NK combo section. Preferably in the nuclear.H file, as the packer puts that first. when reading the single header, the public api prototypes are listed and documented near the top, then private library comes second; then definitions later. So placing docs above the definition will not be seen by future readers unless they explicitly scroll through 30k lines to find it. 

Also, overview.c is not sourced for documentation

awschult002:
Unfortunately, the documentation format you have chosen has been recently deprecated. Take a look at the other header blocks in the most recent `src/nuklear.h`

PR #732, opened 2024-11-15
Makefile Upgrade

awschult002:
The top level makefile has been upgraded to handle the building of everything in the project. Demos are recursively built, documentation is created, and Nuklear.h gets paq'd.  This does not break the previous build system, only adds a new path.

In order to make this happen, a small modification to `build.py` needed to happen to allow space delimited lists of files. Note, this change is backwards compatible.

Using the makefile at the top level for the building is not only familiar to C developers; but also provides better control and build capabilities to any automated systems. Multi-threaded building and processing, plus inherited compiler variables can help to simplify more advanced build procedures while providing a clean and common way for developers to make future adjustments.

After this update, I hope to modify the github workflow to reduce build times.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,36 +1,90 @@
 
-docs_path:=./doc
-doxyfile:=$(docs_path)/Doxyfile
+######################################################################################
+##  								 SETTINGS                                       ##
+######################################################################################
+
+## path stuff
+DOCS_PATH:=./doc
+DEMO_PATH=demo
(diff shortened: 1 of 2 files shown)


RobLoach:
Oh this is nice. Thanks so much. Before merging, is there anything else you'd like to add to it?

awschult002:
I would prefer this get merged before I continue working on it

Issue #730, opened 2024-11-08
Documentation link no longer works


csstup:
The doc link on the main README of the repo goes to https://immediate-mode-ui.github.io/Nuklear/doc/nuklear.html.   This is a 404 now.

adrianboyko:
Captured by archive.org on Feb 8, 2024:
https://web.archive.org/web/20240208041722/https://immediate-mode-ui.github.io/Nuklear/doc/index.html

riri:
It seems it is because of PR #722 merged, but no CI/DI workflow has been created (because doc is in ignore list now), hence the absence on github.io. @awschult002 could you have a look ?

riri:
Fixed in merged PR #734

PR #734, opened 2024-11-18
Restore documentation on github pages with Doxygen generated doc

riri:
This is an attempt to fix issue #730 since the original `gh-pages` branch is not fed anymore (with `.gitignore` changes).

I'm not an expert at all with Doxygen and github actions, so that might need more work (help is welcome).

With this, a few hints:
* I have removed README.md as start page, because it was not giving much and the _Documentation_ link was conflicting, instead the start page is made from `src/HEADER.md`
* The build generates a lot of warnings and errors. I suppose warnings can be fixed in the future from source files. Errors seem to come from the missing `dot` command (an additional package might be added to the required docker image before running doxygen)
* Once merged, `Pages` settings of the repository need to be changed because the generated doc is done in the root of the `gh-pages` branch (and not `/docs` anymore). This is how I have setup my fork :

Change in the folder to be used as root for pages

The result can be [seen on my fork](https://riri.github.io/Nuklear/index.html) deployed pages :)

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -1,0 +1,26 @@
+name: Documentation on github.io
+
+on:
+  push:
+    branches: [ github.io ]
+
+jobs:
+  build-documentation:
+    runs-on: ubuntu-latest
+    steps:
+    - name: checkout
+      uses: actions/checkout@v4
+    - name: apt-update
+      run: sudo apt-get update -qq
+    - name: apt-get doxygen
+      run: sudo apt-get install -y doxygen
+    - name: build doc
+      run: make docs
+    - name: deploy
+      uses: peaceiris/actions-gh-pages@v3
+      with:
+        github_token: ${{ secrets.GITHUB_TOKEN }}
+        publish_dir: ./docs/html/
+        enable_jekyll: false
+        allow_empty_commit: false
+        force_orphan: true
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -2,7 +2,7 @@ name: Documentation on github.io
 
 on:
   push:
-    branches: [ github.io ]
+    branches: [ master ]
 
 jobs:
   build-documentation:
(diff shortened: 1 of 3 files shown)


RobLoach (approved):
This is a much better approach. I have access to https://github.com/Immediate-Mode-UI/Nuklear/settings/pages if you're missing it to make the change afterwards.

awschult002:
i know that this is already merge, but i decided to give a read through since my name was mentioned in the issue #730.

First thing. Thank you for handling the issue. This definitely fixed the issue. but i do have a couple questions/notes

I am not super familiar with github pages, so i am going to have to do some reading; but i really do like the feature and the automated workflow. My only concern with this PR is it appears that the published documents are not on master; but instead published from the `gh-pages` branch. and the `gh-branch` has all of the docs in the root folder. I am not sure how you managed that; but at least it is working for now.

I would recommend that we try to move our publishing to be based directly off master and change its root directory to `/docs/html` . That way, we have minimal changes required when pushing a new update. I will do some testing on my end to see what I can learn about github pages.

Thank you again for the quick fix.

riri:
In fact it's simple once the github action is made. It's a two parts process:

1. automatic generation is done through github action which exactly processes like testing builds (on CI) in a temporary docker image created for this job (it's not retained). Then automatic deployment of generated content to github page is done via another step in the job. Historically it has been a commit in the `gh-pages` branch
2. Project pages is configured to check the content from one branch (`gh-pages`) in a specific directory (`/ (root)` or `/docs`)

The interest of having generated content in a specific branch is.... it does not pollute the master branch, otherwise we would need an extra commit each time something is pushed in master (or rely on people re-generate the docs before PR, but we can't rely on that).

And it's all automated with CI: each push on master triggers documentation generation and update of github.io.

awschult002:
Thank you for the insight. I did some research into the actions and pages. I just posted a PR that continues to use actions, but removes the need for HTML to be committed at all. The actions handle the generation and the artifact uploading and deployment automatically. PR #737

PR #742, opened 2024-11-20
Fix `nk_roundf()` int type casting

RobLoach:
Was getting a conversion warning on `nk_roundf`...

> include/nuklear.h: In function ‘nk_roundf’:
include/nuklear.h:6630:52: warning: conversion from ‘int’ to ‘float’ may change value [-Wconversion]

### DIFF
--- a/nuklear.h
+++ b/nuklear.h
@@ -30698,6 +30698,7 @@ nk_tooltipfv(struct nk_context *ctx, const char *fmt, va_list args)
 ///   - [y]: Minor version with non-breaking API and library changes
 ///   - [z]: Patch version with no direct changes to the API
 ///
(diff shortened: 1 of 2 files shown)


riri:
You don't increment de patch version?

RobLoach:
I should likely do that, good call. Would be nice to automate the changelog, eventually.

riri:
I thought about it, but it's tricky. How to determine which of patch or minor to increment (major version can't be automated)

Sha2561986:
Hi its possible we can talk, through phone not Chatting,09682053072/or we can talk in any platform in social media,/ I don't know if you can reach me  entire device are controlled, jJomar B. Taboco, ,

PR #748, opened 2024-11-26
add conditional style configurator to all demos

riri:
I have added the new style-configurator demo to all platform demos, and tested what I could test (on linux).

There are still a few things that could be improved, that why I suggest we keep that branch for further updates:
* add a style section for knobs
* find a way to keep the color table inside the demo function

Other points:
* during testing, I couldn't make `sfml_opengl2` and `sfml_opengl3` demos working, because they rely on hard coded paths.
* the `glfw_opengl4` is segfaulting, even without the addition.


riri:
> Outside the scope of this changeset, but I think it would be really cool to be able to click an Export button to send the contents of the style changes to a `.h` file in your clipboard.

Working on it, I really want to easily provide custom styles to my projects, thanks to @rswinkle feature.

PR #749, opened 2024-11-30
sets xcb window's title

beatwise:
sets xcb demo window's title


riri:
I've added it to my uptream copy to test.

Thanks for the small code ;-)

PR #755, opened 2024-12-07
Fix: property border uses same color as its background

Grisshink:
Got across this small bug in code where the border color of the property widget matches the color of its background rendering it invisible. This PR should fix this bug.

### Expected behavior
Property draws its border as i defined it in styles
Property with borders

### Actual behavior
Property draws its border with the same color making it not visible
Property with bugged borders

This becomes more apparent with wider borders
Property with large bugged borders


RobLoach:
:beans:

PR #767, opened 2025-02-02
Support numpad Enter with SDL backends

rswinkle:
If you tell a user they can submit with Enter and then it doesn't work, well... they don't care that the numpad Enter is technically a different key.  They have a point, they both are labeled "Enter" on most keyboards.

I can go through the other backends adding support but before I do I want to get feedback on my approach.  I can't think of any reason we'd want/need to add NK_KEY_KP_ENTER to nuklear itself because I can't think of a case we'd actually want distinct behavior for regular Enter/Return vs Numpad Enter.  If anyone can think of a convincing case I'll make the internal changes instead.

In either case I would need help testing several of them, specifically the windows backends, but possibly some of the more esoteric ones as well.


RobLoach:
Thanks!

Issue #760, opened 2025-01-04
SDL3 support


trexxet:
Shouldn't be hard, just a few patches to nuklear_sdl_gl3.h

Here's my working solution using CMake:
```cmake
file(READ ${NUKLEAR_PREFIX_INCLUDE}/nuklear_sdl_gl3.h SDL_FILE_CONTENTS)
string(REPLACE "SDL2" "SDL3" SDL_FILE_CONTENTS "${SDL_FILE_CONTENTS}")
string(REPLACE "SDL_GetTicks64" "SDL_GetTicks" SDL_FILE_CONTENTS "${SDL_FILE_CONTENTS}")
string(REPLACE "SDL_GL_GetDrawableSize" "SDL_GetWindowSizeInPixels" SDL_FILE_CONTENTS "${SDL_FILE_CONTENTS}")
string(REPLACE "SDL_TRUE" "true" SDL_FILE_CONTENTS "${SDL_FILE_CONTENTS}")
string(REPLACE "SDL_FALSE" "false" SDL_FILE_CONTENTS "${SDL_FILE_CONTENTS}")
string(REPLACE "SDL_KEYUP" "SDL_EVENT_KEY_UP" SDL_FILE_CONTENTS "${SDL_FILE_CONTENTS}")
string(REPLACE "SDL_KEYDOWN" "SDL_EVENT_KEY_DOWN" SDL_FILE_CONTENTS "${SDL_FILE_CONTENTS}")
string(REPLACE "SDL_MOUSEBUTTONUP" "SDL_EVENT_MOUSE_BUTTON_UP" SDL_FILE_CONTENTS "${SDL_FILE_CONTENTS}")
string(REPLACE "SDL_MOUSEBUTTONDOWN" "SDL_EVENT_MOUSE_BUTTON_DOWN" SDL_FILE_CONTENTS "${SDL_FILE_CONTENTS}")
string(REPLACE "SDL_MOUSEMOTION" "SDL_EVENT_MOUSE_MOTION" SDL_FILE_CONTENTS "${SDL_FILE_CONTENTS}")
string(REPLACE "SDL_MOUSEWHEEL" "SDL_EVENT_MOUSE_WHEEL" SDL_FILE_CONTENTS "${SDL_FILE_CONTENTS}")
string(REPLACE "SDL_TEXTINPUT" "SDL_EVENT_TEXT_INPUT" SDL_FILE_CONTENTS "${SDL_FILE_CONTENTS}")
[...]

PR #719, opened 2024-10-06
Added demo updated for SDL3

lessthen3:
added new demo headers that are updated to SDL3 and have been confirmed to compile and run appropriately on my machine.

PR #772, opened 2025-02-06
[DRAFT] SDL3_renderer and SDL_gpu support

zoogies:
I plan on starting SDL_gpu soon, but I've fully implemented the SDL3_renderer backend.

Main changes:
- `SDL_Vertex` now uses `SDL_FColor` (float) instead of `SDL_Color` (uint8)
- removed grabbing, because as far as I can tell it's been broken for a long time (I have it disabled in my own personal projects)
   - #608

Considerations:
- I've had to `#include <stddef.h>` for `offsetof`, but the sdl2_renderer backend doesn't do this. I'm assuming that @RobLoach knows where these Nuklear backend definitions are 

Relevant:
- #760 
- #719


zoogies:
It's worth mentioning that the default text input event behavior has changed from SDL2 to SDL3.

In SDL2 the events were fired on keypresses by default on desktop, but not mobile. In SDL3, these events are never fired until `SDL_StartTextInput()` is called. This function will also force showing the on-screen keyboard on devices that support it, so if this demo is adapted for other platforms the call to `SDL_StartTextInput()` should be moved out of `SDL_AppInit()`

RobLoach:
Looking great! Both demos worked well here. Couple questions...
1. Do you think we should move the demos to an `sdl3` folder or something to help consolidate them? Could save on CMake definitions then.
2. > I've had to #include <stddef.h> for offsetof, but the sdl2_renderer backend doesn't do this. I'm assuming that @RobLoach knows where these Nuklear backend definitions are

    That seems fine with me. If an NK_OFFSETOF doesn't exist, we could introduce one.
    
3. > In SDL3, these events are never fired until SDL_StartTextInput() is called

    Think that's worth avoiding a merge? Or just do it?

zoogies:
Currently, the `SDL_gpu` backend is just using sdl_renderer, but I'm going to switch it over (hopefully tomorrow) and I'll convert this PR from a draft into a regular PR when I finish 😄 

> Do you think we should move the demos to an sdl3 folder or something to help consolidate them? Could save on CMake definitions then.

Personally, I like the current setup of each backend having its own folder in `demo/`

> That seems fine with me. If an NK_OFFSETOF doesn't exist, we could introduce one.

For some reason, `offsetof` is properly defined in the existing SDL2 renderer backend without `stddef.h`. Maybe there already is a Nuklear version floating around? 

> Think that's worth avoiding a merge? Or just do it?

What do you mean by merge? In this case, I think the only change worth making is adding a big comment above the `SDL_StartTextInput()` call to explain it, since it's relevant for any developers using the demo as a reference

Also, I had to link both demos against `libm` due to a linker error about a missing `floor` implementation. Not sure if you have any ideas? (I've put `TODO` comments around anything like this that I think deserves review)

RobLoach:
> I like the current setup of each backend having its own folder in demo/

We do adopt that pattern for the rawfb ones. It's kind of nice having a consolidated one, could possibly save some code:
https://github.com/Immediate-Mode-UI/Nuklear/tree/master/demo/rawfb

We could in theory have a `nuklear_sdl3.h` that contains the input handling, and then calls either the GPU or renderer functions depending on which is picked.

> What do you mean by merge?

Apologies. I meant merge the pull request as is. But sounds like we should wait off.

zoogies:
I've gotten pretty busy lately, but this PR is still on my mind. Pretty soon I hope to get back around to implementing the `SDL_gpu` backend, but it will likely take me a while as I've never worked with it before.

> We could in theory have a nuklear_sdl3.h that contains the input handling, and then calls either the GPU or renderer functions depending on which is picked.

This is a good point. Personally I think it makes the most sense to keep them separate, but we can see what the overlap looks like once the `SDL_gpu` backend is ready.

> Apologies. I meant merge the pull request as is. But sounds like we should wait off.

@RobLoach if you would like, I can split the unfinished `SDL_gpu` work into a separate PR, and clean this one up for merge. Let me know what you'd prefer 😄

RobLoach:
No worries. I can clean this up and get things in place. Thanks so much for the push forwards here 👀

RobLoach:
https://github.com/Immediate-Mode-UI/Nuklear/pull/779

paccerdk:
I'd just like to add a quick note

@zoogies commented:

>* removed grabbing, because as far as I can tell it's been broken for a long time (I have it disabled in my own personal projects)
>     * [Inconsistent grabbing behavior in demos (click and drag property) #608](https://github.com/Immediate-Mode-UI/Nuklear/issues/608)

If you check the [Feb 10, 2024](https://github.com/Immediate-Mode-UI/Nuklear/issues/608#issuecomment-1936840879) post in the issue you're linking to, the grabbing should be fixed and working correctly in the **SDL demos** through the mentioned [PR](https://github.com/Immediate-Mode-UI/Nuklear/pull/609)

The issue remains open because grabbing behavior is still inconsistent in **other** demos.
However, grabbing in SDL demos should be working correctly (at least they were at that time).

zoogies:
Thanks for pointing that out, I wasn't aware since I'm not using the latest Nuklear version in my personal projects. It's definitely worth re-introducing the grabbing in #779

RobLoach:
> The issue remains open because grabbing behavior is still inconsistent in **other** demos. However, grabbing in SDL demos should be working correctly (at least they were at that time).

Thanks so much for fixing that up. We will have to do another check on the SDL3 one, which isn't quite merged yet. I haven't been able to test fully yet.

Issue #739, opened 2024-11-18
Add a clang-format format file


theoparis:
I was looking to make a improved version of https://github.com/Immediate-Mode-UI/Nuklear/pull/50, which I needed to have implemented along with the knob widget change for a mini DAW implementation. I was also considering making a SDL3 GPU api PR in the future if no one else does it. However, I noticed that there is no .clang-format file and I am assuming the defaults for clang-format would mess up any PR that it is applied to.

This would probably introduce some formatting changes, but it would make it easier to contribute if formatting on saving is enabled (a feature in most IDEs).  This also means there would be a standard for styling code, instead of just saying it has to be "consistent".

RobLoach:
Feel free to submit a `.clang-format` that makes closely to what's currently there.

PR #774, opened 2025-02-10
clang-format config

PROP65:
clang-format does not have the most granular configuration so some things (like different pointer placement on unnamed arguments in function prototypes) are not currently achievable. 

From the tinkering I've done with it so far it seems that formatting before building yields better results.  If formatting is done on the final header file all preprocessor directives will be further indented.

Major changes compared to current styling:
 * preprocessor directive indentation
 * max line length of 160

I'm open to changing the current config if needed, I'll admit it's far from perfect.

If needed I can format the src folder and rebuild once the style config is agreed on.

Here's `nuklear.h` after formatting and rebuilding with this change
https://gist.github.com/PROP65/bf35c5a22a210584d5057fb2b6a1323d


PROP65:
#739

RobLoach:
I think this is worth adopting. The syntax within nuclear.h right now is all over the place. Having adopt this format across the src files sequentially could get us there.

PR #779, opened 2025-02-20
Add SDL3

RobLoach:


zoogies:
Sure! I'll take a look in a few hours. Thanks

RobLoach:
Also made it thread-safe. There are a couple things that I think could improve. I'm not 100% sure it's cleaning up the memory correctly tho

demo/sdl3/nuklear_sdl3_renderer.h
```diff
@@ -0,0 +1,351 @@
...
+    (void)usr;
+}
+
+static void
+nk_sdl_clipboard_copy(nk_handle usr, const char *text, int len)
+{
+    char *str = 0;
+    (void)usr;
+    if (!len) return;
+    str = (char*)SDL_malloc((size_t)len+1);
```
klei1984, line 195: In case of multi-byte text, like SDL TextInput which is unicode, `len` is count grapheme clusters or glyphs, not bytes.

demo/sdl3/nuklear_sdl3_renderer.c
```diff
@@ -0,0 +1,251 @@
...
+    if (appContext == NULL) {
+        return SDL_Fail();
+    }
+
+    if (!SDL_CreateWindowAndRenderer("Nuklear: SDL3 Renderer", WINDOW_WIDTH, WINDOW_HEIGHT, SDL_WINDOW_RESIZABLE, &appContext->window, &appContext->renderer)) {
+        SDL_free(appContext);
+        return SDL_Fail();
+    }
+
+    SDL_StartTextInput(appContext->window);
```
klei1984, line 106: On desktop platforms, SDL_StartTextInput() is implicitly called on SDL window creation. On some, typically non desktop, platforms this call implicitly enables and shows virtual keyboard.

demo/sdl3/nuklear_sdl3_renderer.h
```diff
@@ -0,0 +1,351 @@
...
+    str[len] = '\0';
+    SDL_SetClipboardText(str);
+    SDL_free((void*)str);
+}
+
+NK_API struct nk_context*
+nk_sdl_init(SDL_Window *win, SDL_Renderer *renderer)
+{
+    struct nk_sdl* sdl = SDL_malloc(sizeof(struct nk_sdl));
+    NK_ASSERT(sdl);
```
klei1984, line 207: Probably it is a good idea to sanitize the inputs too. Assert on win/renderer = null.

demo/sdl3/nuklear_sdl3_renderer.h
```diff
@@ -0,0 +1,351 @@
...
+                        nk_input_button(ctx, NK_BUTTON_LEFT, x, y, down); break;
+                    case SDL_BUTTON_MIDDLE: nk_input_button(ctx, NK_BUTTON_MIDDLE, x, y, down); break;
+                    case SDL_BUTTON_RIGHT:  nk_input_button(ctx, NK_BUTTON_RIGHT, x, y, down); break;
+                }
+            }
+            return 1;
+
+        case SDL_EVENT_MOUSE_MOTION:
+            if (ctx->input.mouse.grabbed) {
+                int x = (int)ctx->input.mouse.prev.x, y = (int)ctx->input.mouse.prev.y;
```
klei1984, line 311: SDL3 (mouse) movement events were changed from integer to float. Nuklear uses float internally too. Is the integer cast truncation intentional?

demo/sdl3/nuklear_sdl3_renderer.h
```diff
@@ -0,0 +1,351 @@
...
+                nk_input_motion(ctx, x + evt->motion.xrel, y + evt->motion.yrel);
+            }
+            else nk_input_motion(ctx, evt->motion.x, evt->motion.y);
+            return 1;
+
+        case SDL_EVENT_TEXT_INPUT:
+            {
+                nk_glyph glyph;
+                SDL_memcpy(glyph, evt->text.text, NK_UTF_SIZE);
```
klei1984, line 320: `evt->text.text` is UTF-8 encoded const char* nul terminated string. Based on limited research it is not guaranteed that a single glyph is reported by this event. Could it happen that text input gets cropped as the UTF-8 input is passed per glyph to nk_input _glyph in a loop?

demo/sdl3/nuklear_sdl3_renderer.h
```diff
@@ -0,0 +1,351 @@
...
+                nk_input_glyph(ctx, glyph);
+            }
+            return 1;
+
+        case SDL_EVENT_MOUSE_WHEEL:
+            nk_input_scroll(ctx,nk_vec2((float)evt->wheel.x,(float)evt->wheel.y));
```
klei1984, line 326: SDL3 input device movements are float by definition, casts are redundant.

demo/sdl3/nuklear_sdl3_renderer.h
```diff
@@ -0,0 +1,351 @@
...
+
+    if (sdl->ogl.font_tex != NULL) {
+        SDL_DestroyTexture(sdl->ogl.font_tex);
+        sdl->ogl.font_tex = NULL;
+    }
+
+    ctx->userdata = nk_handle_ptr(0);
+    SDL_free(sdl);
+    nk_free(ctx);
+    /* TODO: Need to call SDL_free(ctx)? */
```
klei1984, line 348: This can be removed as NK_MFREE is overridden now.

RobLoach: `nk_free()` doesn't call free the memory with `NK_MFREE` though. It just zeroes everything out :thinking:

klei1984: Oh! Sorry, i made wrong assumption on NK_MFREE macro and confused nk_free with nk_mfree which is not even an API but LIB function.

RobLoach: It is confusing. When SDL_free() is in there, there's a double free conflict. Not entirely sure what's wrong.

klei1984: So if I understand correctly, then it is safe to clear the Nuklear context by nk_free as in the given integration `nk_sdl_shutdown()` is only called when Nuklear APIs are not used anymore. And as the allocator of the nk_context `ctx` was the integrator (application), it should be responsible for the deallocation. E.g. an integrator may statically allocate a `ctx` variable and then calling on it `SDL_free()` is wrong.

But looking into the integration actually there are several issues with the current concept as there are several exit paths via the SDL3 callbacks that could trip on null pointers or could attempt to access uninitialized resources. SDL_AppQuit() is called by most callbacks like `SDL_AppInit()`, `SDL_AppIterate()` or even `SDL_AppEvent()` when anything goes wrong.

**Error case 1: `appstate**` is not set yet**
`SDL_AppInit()` is called, which calls `SDL_Init()` that fails and in turn `SDL_Fail()` sets SDL_APP_FAILURE. Then `SDL_AppQuit()` gets called and it dereferences a potentially invalid or null pointer by `AppContext* app = (AppContext*)appstate;`.

**Error case 2: `SDL_malloc(sizeof(AppContext));` fails**
If the given malloc fails, `*appstate` is set to null pointer, same failure mode happens as in error case 1, but this time it is dereferencing null not potentially random value.

**Error case 3: `SDL_CreateWindowAndRenderer(...)` fails**
In this case the appstate is valid, the appContext is valid too, but as it is not memset to zero, potentially random value is present in `appContext->window` and `appContext->renderer`. Then in `SDL_AppQuit()` eventually `SDL_DestroyRenderer()` and the likes are called. This API asserts or trips on null pointer input and probably crashes on random values.

**Error case 4: `nk_sdl_shutdown(app->ctx);` before successful call to nk_sdl_init()**
nk_sdl_init() is assumed not to fail due to failed memory allocations. In such cases release builds would segfault. So lets assume again that `SDL_CreateWindowAndRenderer()` fails. In that case `nk_sdl_shutdown(app->ctx);` is called with random value in ctx and `nk_font_atlas_clear` or similar will probably trip.

**Potential error case 1: `SDL_StartTextInput() without SDL_StopTextInput()`**
Documentation says `Please use this function in pair with SDL_StopTextInput`. And on desktops the start API is called implicitly too, so probably when window is destroyed it is again called implicitly. But as start is also called explicitly by the integrator or application, it should probably call the stop function too to be in pairs. I have no clue whether these are counting semaphores or similar.

demo/sdl3/nuklear_sdl3_renderer.c
zoogies: In SDL3, this needs to be explicitly called on Desktop platforms. I think adding a comment about the behavior of calling this on mobile is probably in order

demo/sdl3/nuklear_sdl3_renderer.h
zoogies: This was a hack on my part, changing to float and removing the cast should be fine

demo/sdl3/nuklear_sdl3_renderer.c
```diff
@@ -0,0 +1,251 @@
...
+    #endif
+    #ifdef INCLUDE_CONFIGURATOR
+        style_configurator(ctx, color_table);
+    #endif
+    #ifdef INCLUDE_NODE_EDITOR
+        node_editor(ctx);
+    #endif
+    /* ----------------------------------------- */
+
+    SDL_SetRenderDrawColor(app->renderer, 255, 0, 255, SDL_ALPHA_OPAQUE);
```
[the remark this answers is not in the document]
zoogies: The color can be changed, but it exists on the demo side rather than the backend to show a typical render loop end-to-end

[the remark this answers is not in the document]
zoogies: I think of the backend header as the Nuklear code, and the demo C to be just a reference example on how to use said backend

zoogies:
I don't have push access to this branch, but this patch fixes the casting klei mentioned:
```diff
index 5a8c589..e819169 100644
--- a/demo/sdl3/nuklear_sdl3_renderer.h
+++ b/demo/sdl3/nuklear_sdl3_renderer.h
@@ -308,10 +308,10 @@ nk_sdl_handle_event(struct nk_context* ctx, SDL_Event *evt)
 
         case SDL_EVENT_MOUSE_MOTION:
             if (ctx->input.mouse.grabbed) {
-                int x = (int)ctx->input.mouse.prev.x, y = (int)ctx->input.mouse.prev.y;
-                nk_input_motion(ctx, x + evt->motion.xrel, y + evt->motion.yrel);
+                float x = ctx->input.mouse.prev.x, y = ctx->input.mouse.prev.y;
+                nk_input_motion(ctx, (int)(x + evt->motion.xrel), (int)(y + evt->motion.yrel));
             }
-            else nk_input_motion(ctx, evt->motion.x, evt->motion.y);
+            else nk_input_motion(ctx, (int)evt->motion.x, (int)evt->motion.y);
             return 1;
 
         case SDL_EVENT_TEXT_INPUT:
@@ -323,7 +323,7 @@ nk_sdl_handle_event(struct nk_context* ctx, SDL_Event *evt)
             return 1;
 
         case SDL_EVENT_MOUSE_WHEEL:
-            nk_input_scroll(ctx,nk_vec2((float)evt->wheel.x,(float)evt->wheel.y));
+            nk_input_scroll(ctx, nk_vec2(evt->wheel.x, evt->wheel.y));
             return 1;
     }
     return 0;
```

The first casting change is technically unnecessary, but it allows adding the prev and rel values as floats before converting them back to ints.

demo/sdl3/nuklear_sdl3_renderer.c
```diff
@@ -0,0 +1,251 @@
+/* nuklear - public domain */
+#include <stdio.h>
+#include <stddef.h> /* offsetof TODO: sdl2_renderer didn't need this header? */
```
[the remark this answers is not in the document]
RobLoach: Right. And we can use `NK_OFFSET` here anyway. Removing this header.

demo/sdl3/nuklear_sdl3_renderer.h
RobLoach: Know the significance of `ctx->input.mouse.grabbed`? Looking at the other implementations, I'm not quite sure how they'd apply here.

klei1984: It seems as if while window is moved around by grabbing it with mouse, then absolute mouse pointer x,y cannot be trusted and thus relative x,y are being used.

The SDL documentation says in this scenario `SDL_GetGlobalMouseState()` is supposed to be used.

But actually the situation is more complex. Typically a full screen or borderless window SDL application would force relative mouse movement, and locked mouse into window by SDL_SetRelativeMouseMode. This mode grabs the mouse pointer by the window and hides the system cursor or something like that. And when the application switches to windowed mode, then relative mouse mode is disabled and mouse is only locked otionally into window when window gets focus and mouse gets unlocked when window loses focus.

RobLoach: After some testing, I think what I've put in there seems to be working with or without window grabbing.
``` c
case SDL_EVENT_MOUSE_MOTION:
    ctx->input.mouse.prev = ctx->input.mouse.pos;
    ctx->input.mouse.delta.x = evt->motion.xrel;
    ctx->input.mouse.delta.y = evt->motion.yrel;
    ctx->input.mouse.pos.x = evt->motion.x;
    ctx->input.mouse.pos.y = evt->motion.y;
    return 1;
```

I played around with SetRelativeMouseMode too, and it didn't seem to have any real affect...
``` c
case SDL_EVENT_MOUSE_MOTION:
{
    struct nk_sdl* sdl = (struct nk_sdl*)ctx->userdata.ptr;
    ctx->input.mouse.prev = ctx->input.mouse.pos;
    ctx->input.mouse.delta.x = evt->motion.xrel;
    ctx->input.mouse.delta.y = evt->motion.yrel;
    ctx->input.mouse.pos.x = evt->motion.x;
    ctx->input.mouse.pos.y = evt->motion.y;
    if (ctx->input.mouse.grabbed) {
        SDL_SetWindowRelativeMouseMode(sdl->win, true);
        ctx->input.mouse.pos.x += evt->motion.x;
        ctx->input.mouse.pos.y += evt->motion.y;
    }
    else {
        SDL_SetWindowRelativeMouseMode(sdl->win, false);
        ctx->input.mouse.pos.x = evt->motion.x;
        ctx->input.mouse.pos.y = evt->motion.y;
    }
    return 1;
}
```

Overall, I think what's in there is fine and don't want to over-complicate it :shrug: ... What do you think?

demo/sdl3/nuklear_sdl3_renderer.c
RobLoach: Hmm, is there a cleaner way we could accomplish this? Perhaps a callback when a textbox on focus/blur?

demo/sdl3/nuklear_sdl3_renderer.h
```diff
@@ -0,0 +1,351 @@
...
+{
+    struct nk_sdl* sdl = (struct nk_sdl*)ctx->userdata.ptr;
+    SDL_Texture *g_SDLFontTexture = SDL_CreateTexture(sdl->renderer, SDL_PIXELFORMAT_ARGB8888, SDL_TEXTUREACCESS_STATIC, width, height);
+    if (g_SDLFontTexture == NULL) {
+        SDL_Log("error creating texture");
+        return;
+    }
+    SDL_UpdateTexture(g_SDLFontTexture, NULL, image, 4 * width);
+    SDL_SetTextureBlendMode(g_SDLFontTexture, SDL_BLENDMODE_BLEND);
+    sdl->ogl.font_tex = g_SDLFontTexture;
```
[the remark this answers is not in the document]
RobLoach: Thanks. Added a check to unload the texture if it already exists.

RobLoach (commented):
Added a few more changes.

demo/sdl3/nuklear_sdl3_renderer.h
```diff
@@ -0,0 +1,380 @@
...
+    NK_ASSERT(ctx);
+    sdl = (struct nk_sdl*)ctx->userdata.ptr;
+    NK_ASSERT(sdl);
+    return sdl->userdata;
+}
+
+NK_API void nk_sdl_set_userdata(struct nk_context* ctx, nk_handle userdata) {
+    struct nk_sdl* sdl;
+    NK_ASSERT(ctx);
+    sdl = (struct nk_sdl*)ctx->userdata.ptr;
```
RobLoach, line 75: One worry I have about this strategy is that the SDL renderer is taking over the `nk_context` `userdata` to store information required by the renderer. If the application is making use of the `userdata`, there will be a conflicts unless they change to use `nk_sdl_set_userdata()` instead.

Perhaps we introduce another renderer-specific `userdata` property? Something like `void* renderer_data` or something?

demo/sdl3/nuklear_sdl3_renderer.h
klei1984: I agree, it is fine as it is.

`SDL_SetWindowRelativeMouseMode()` is best to use when the application is in full screen or borderless full screen modes. The mouse cursor is confined into the whole screen (so app window), the OS / HW mouse pointer is hidden (but software cursor could be used), and mouse motion events are rapidly reporting mouse motion position and delta position. This is fit for most action games where you control a player character. Or in case of emulators where cursor is also emulated, e.g. DOSBox.

The way you used that API is somewhat wrong as when relative mode is enabled, the current event does not reflect the changes yet. In relative mode we would only use `evt->motion.xrel/yrel`. In normal mode we would use only the absolute position so `evt->motion.x/y`.

demo/sdl3/nuklear_sdl3_renderer.h
klei1984: I assume the renderer backends are still examples only and not part of the UI library, which is a single header. E.g. I change nk_sdl_handle_event to add more keys, etc. As integrator, I am in full control over the userdata. E.g. I do not have nk_sdl sdl at all, I embed the stuff from it into my own context thingy.

RobLoach: So you use your own userdata elsewhere outside of the nuklear context?

klei1984: Um, I do not use `nuklear_sdl3_renderer.h` and APIs like `nk_sdl_render(struct nk_context* ctx` at all. I use the demo as an example and adapt it to my needs. E.g. my own version of `nk_sdl_render()` gets not a Nuklear context, it gets an application context that embeds the Nuklear context too. What I wanted to communicate above is that the life cycle of the Nuklear context is shorter then the application context in my case, and that I consider  `nuklear_sdl3_renderer.h` to be an example, not the official API.

So the answer to your question is actually no. I do not have a specific need for userdata enabled by `NK_INCLUDE_COMMAND_USERDATA`, it can be used by `nuklear_sdl3_renderer.h` and I will adapt that header content as I need.

- Text processing: In case of text processing I use freetype2, different texture packer, custom font atlas, and custom cached glyph encoding library that are not exclusive to `nk_user_font`, and Nuklear. So the life cycle of my text rendering sub system is typically longer than the life cycle of the Nuklear context. The font atlas is based on `SDL_surface` objects where I could do palette color manipulations, which is not a typical use case I admit, and when the Nuklear context is created and render backend is selected the font atlas layers are converted to `SDL_textrure` objects in native texture format. I actually intend to use SDL3 GPU API not the SDL3 renderer. That is important as my whole design revolves around SDL3 GPU's ability to transparently select any backend that is available. I am not deciding which backend to use so my font atlas needs to remain backend agnostic (e.g. uses `SDL_surface` not textures that have backend dependent pixel formats).
 
- GUI: my application context is not a sub feature of Nuklear, instead Nuklear's context is the sub feature of my application context. The same way as for text processing, I tried to shoot for a design which separates the GUI library and its render backend and the application core logic even if this is very difficult with the immediate mode pattern. Nuklear requires a lot of boilerplate and that boilerplate is interleaved into the core application logic which is difficult to maintain so I try to encapsulate, structure and generalize things for which I created an abstraction level above Nuklear similar to a c++ object oriented wrapper would do, but focusing on features (buttons, lists, hotkeys, sound effects, ...) instead of API coverage. I use a custom GUI layout manager that is loading parameters and asset or resource IDs from json files above and independent of Nuklear. Assets are in custom file formats using LZ4 compression and are loaded into renderer agnostic `SDL_surface` objects. Later on when the system specific render backend is known, D3D/OGL/etc, then these surfaces are converted and loaded into GPU memory into `SDL_texture` objects, using native texture formats, with assets packed as most optimal for particular screen content expected to be rendered. I only use Nuklear STATIC layout and position every top level (custom) widgets inside the application frontend or menu screens. Nuklear's render target is just an `SDL_texture` not the swapchain texture or whatever. In short my application's context is one order or abstraction level higher than Nuklear's context. The Nuklear context is only created after the render backend is determined. I am sure there is a better way to do all of this, but this is what i came up with for my particular needs.

demo/sdl3/nuklear_sdl3_renderer.c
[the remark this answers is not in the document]
RobLoach: Thanks for the heads up, fixed.

zoogies:
What is the state of this PR? I'm happy to help with any remaining items

RobLoach:
I believe all that's really missing is...

- Testing
- Figure out the best method around SDL_StartTextInput() usage
- UTF-8 support in SDL_EVENT_TEXT_INPUT

zoogies:
Cool, for `SDL_StartTextInput()`, running this once on init _should_ be okay unless we are on a non-desktop platform.

Sam told me in the SDL Discord to "Remember [that] this will bring up the on-screen keyboard on some platforms, so you should only do this when you're ready for text input".

We could macro guard it or put a comment above it so people are aware of its behavior. I treat the source files for the backends as examples, but the header files for the backends as Nuklear vendored, so IMO it's not unreasonable to just put a disclaimer since it's "just an example".

Alternatively, you could place the start and stop calls inside the focus/defocus events for text inputs or windows, but that might interfere with some edge cases.

ref: https://wiki.libsdl.org/SDL3/SDL_StartTextInput

zoogies:
Actually, Sam also said that:

> Conceptually it should be tied to an input box getting focus.

So it sounds like the latter is the best option. Any ideas on where we should hook into the input focus? I'm not too familiar on whether that should lie at a window focus or input focus boundary. Are there any cases of keyboard navigation that would make sense on desktop but not on other platforms?

RobLoach:
> looking into the SDL TextInput stuff and it looks like the most trivial solution would be to somehow insert the calls into nk_edit_focus and nk_edit_unfocus.

Agreed. Not an elegant solution currently. Are you proposing some kind of event callback system for them?

> I've fully integrated this into my own project

The other thing I'd like to see incorporated into this is a clearer implementation of detailing with the global object state. Or something like encorporating https://github.com/Immediate-Mode-UI/Nuklear/pull/799 into here.

rswinkle:
> > Are you proposing some kind of event callback system for them?
> 
> That's the first thing that comes to mind. I don't have deep knowledge of the Nuklear codebase, so I wasn't sure if that's done in other places, or what the best way to integrate something like that would be. Do you have any opinions?

Maybe a strict callback isn't necessary as that would incur a small but present overhead for all backends.  I could see it being 2 (mutually inclusive) optional macros that are called if they exist at appropriate places in Nuklear.  It wouldn't just be nk_edit_focus/nk_edit_unfocus, because that's just for the manually forcing focus/unfocus, not from the user actually clicking on an edit field or something else.  I'm not sure what the best name is but something like NK_ON_EDIT_FOCUS()/UNFOCUS() might work.  Not sure about any necessary parameters either.

Just my two cents till I actually have the chance to review/test this and have a more informed opinion.

demo/sdl3/nuklear_sdl3_renderer.h
```diff
@@ -0,0 +1,388 @@
...
+
+NK_API struct nk_context*
+nk_sdl_init(SDL_Window *win, SDL_Renderer *renderer)
+{
+    struct nk_sdl* sdl;
+    NK_ASSERT(win);
+    NK_ASSERT(renderer);
+    sdl = SDL_malloc(sizeof(struct nk_sdl));
+    NK_ASSERT(sdl);
+    nk_zero(sdl, sizeof(struct nk_sdl));
```
sleeptightAnsiC, line 230: `nk_zero()` is a private/internal symbol that comes from `src/nuklear_util.c` (defined in `src/nuklear_internal.h`). This causes compilation to fail when the amalgamated header isn't used.

You could probably replace this line with variation of SDL_zero*

EDIT: I can confirm that `SDL_zerop` fixes this issue for me:
```diff
-    nk_zero(sdl, sizeof(struct nk_sdl));
+    SDL_zerop(sdl);
```

demo/sdl3/nuklear_sdl3_renderer.h
```diff
@@ -0,0 +1,388 @@
...
+ *
+ * ===============================================================
+ */
+#ifdef NK_SDL3_RENDERER_IMPLEMENTATION
+#ifndef NK_SDL3_RENDERER_IMPLEMENTATION_ONCE
+#define NK_SDL3_RENDERER_IMPLEMENTATION_ONCE
+
+#ifndef NK_INCLUDE_COMMAND_USERDATA
+#error "nuklear_sdl3 requires the NK_INCLUDE_COMMAND_USERDATA define"
+#endif
```
sleeptightAnsiC, line 43: Not sure if this is useful but there should be more errors, for `NK_INCLUDE_VERTEX_BUFFER_OUTPUT` and `NK_INCLUDE_FONT_BAKING` (and probably more), as I was unable to compile without those dependencies. Perhaps, some functions could be defined conditionally depending on those macros.

demo/sdl3/nuklear_sdl3_renderer.c
sleeptightAnsiC: According to SDL docs, call to `SDL_StartTextInput` needs to be paired with `SDL_StopTextInput`, probably implying that TextInput should NOT be active all the time https://wiki.libsdl.org/SDL3/SDL_StartTextInput#remarks

For reference, take a look at how one of SDL tests implements this: https://github.com/libsdl-org/SDL/blob/cc9937201e421ec55b12ad3f07ff2268f15096e8/test/checkkeys.c#L372-L395

demo/sdl3/nuklear_sdl3_renderer.h
```diff
@@ -0,0 +1,388 @@
...
+
+        case SDL_EVENT_MOUSE_BUTTON_UP: /* MOUSEBUTTONUP & MOUSEBUTTONDOWN share same routine */
+        case SDL_EVENT_MOUSE_BUTTON_DOWN:
+            {
+                int down = evt->button.down;
+                const int x = evt->button.x, y = evt->button.y;
+                switch(evt->button.button)
+                {
+                    case SDL_BUTTON_LEFT:
+                        if (evt->button.clicks > 1)
```
sleeptightAnsiC, line 333:
```patch
-                        if (evt->button.clicks > 1)
+                        if (evt->button.clicks == 2)
```
This isn't perfect but should help with the issue mentioned in: https://github.com/Immediate-Mode-UI/Nuklear/pull/779#issuecomment-2859431959

sleeptightAnsiC:
How about the dependency on `NK_INCLUDE_DEFAULT_ALLOCATOR` (default malloc/free from libc) ?

This PR tries to solve this by providing `NK_MALLOC`/`NK_MFREE` macros but this does not seem like the right approach. There is already a solution to this problem: instead of using `nk_*_init_default()` functions, one could use `nk_*_init()` and pass custom `nk_allocator` to them. Adding new macros while this whole `nk_allocator` mechanism exists feels quite confusing.

EDIT: The patch bellow works with `nk_allocator` approach. I would also suggest moving `NK_MALLOC`/`NK_MFREE` to another PR (or just removing it).

<details><summary>patch</summary>
<p>

```diff
diff --git a/demo/sdl3/nuklear_sdl3_renderer.h b/demo/sdl3/nuklear_sdl3_renderer.h
index c57a423..4b67310 100644
--- a/demo/sdl3/nuklear_sdl3_renderer.h
+++ b/demo/sdl3/nuklear_sdl3_renderer.h
@@ -16,6 +16,12 @@
 #error "nuklear_sdl3_renderer requires at least SDL 3.0.0"
 #endif
 
+/* We have to redefine it because demos do not include any headers
+ * This is the same default value as the one from "src/nuklear_internal.h" */
+#ifndef NK_BUFFER_DEFAULT_INITIAL_SIZE
+    #define NK_BUFFER_DEFAULT_INITIAL_SIZE (4*1024)
+#endif
+
 NK_API struct nk_context*   nk_sdl_init(SDL_Window *win, SDL_Renderer *renderer);
 NK_API struct nk_font_atlas* nk_sdl_font_stash_begin(struct nk_context* ctx);
 NK_API void                 nk_sdl_font_stash_end(struct nk_context* ctx);
@@ -79,6 +85,27 @@ NK_API void nk_sdl_set_userdata(struct nk_context* ctx, nk_handle userdata) {
     sdl->userdata = userdata;
 }
 
+NK_INTERN void *
+nk_sdl_alloc(nk_handle user, void *old, nk_size size)
+{
+    NK_UNUSED(user);
+    NK_UNUSED(old);
+    return SDL_malloc(size);
+}
+
+NK_INTERN void
+nk_sdl_free(nk_handle user, void *old)
+{
+    NK_UNUSED(user);
+    SDL_free(old);
+}
+
+NK_GLOBAL const struct nk_allocator nk_sdl_allocator = {
+    .userdata = NULL,
+    .alloc = nk_sdl_alloc,
+    .free = nk_sdl_free,
+};
+
 NK_INTERN void
 nk_sdl_device_upload_atlas(struct nk_context* ctx, const void *image, int width, int height)
 {
@@ -146,8 +173,8 @@ nk_sdl_render(struct nk_context* ctx, enum nk_anti_aliasing AA)
         config.line_AA = AA;
 
         /* convert shapes into vertexes */
-        nk_buffer_init_default(&vbuf);
-        nk_buffer_init_default(&ebuf);
+        nk_buffer_init(&vbuf, &nk_sdl_allocator, NK_BUFFER_DEFAULT_INITIAL_SIZE);
+        nk_buffer_init(&ebuf, &nk_sdl_allocator, NK_BUFFER_DEFAULT_INITIAL_SIZE);
         nk_convert(&sdl->ctx, &sdl->ogl.cmds, &vbuf, &ebuf, &config);
 
         /* iterate over and execute each draw command */
@@ -230,12 +257,12 @@ nk_sdl_init(SDL_Window *win, SDL_Renderer *renderer)
     nk_zero(sdl, sizeof(struct nk_sdl));
     sdl->win = win;
     sdl->renderer = renderer;
-    nk_init_default(&sdl->ctx, 0);
+    nk_init(&sdl->ctx, &nk_sdl_allocator, NULL);
     sdl->ctx.userdata = nk_handle_ptr((void*)sdl);
     sdl->ctx.clip.copy = nk_sdl_clipboard_copy;
     sdl->ctx.clip.paste = nk_sdl_clipboard_paste;
     sdl->ctx.clip.userdata = nk_handle_ptr(0);
-    nk_buffer_init_default(&sdl->ogl.cmds);
+    nk_buffer_init(&sdl->ogl.cmds, &nk_sdl_allocator, NK_BUFFER_DEFAULT_INITIAL_SIZE);
     return &sdl->ctx;
 }
 
@@ -246,7 +273,7 @@ nk_sdl_font_stash_begin(struct nk_context* ctx)
     NK_ASSERT(ctx);
     sdl = (struct nk_sdl*)ctx->userdata.ptr;
     NK_ASSERT(sdl);
-    nk_font_atlas_init_default(&sdl->atlas);
+    nk_font_atlas_init(&sdl->atlas, &nk_sdl_allocator);
     nk_font_atlas_begin(&sdl->atlas);
     return &sdl->atlas;
 }
```

</p>
</details>

sleeptightAnsiC:
Small thing that I'm trying to solve: How can I get SDL's internal debug font (the one from [SDL_RenderDebugText](https://wiki.libsdl.org/SDL3/SDL_RenderDebugText)) and use it inside of Nuklear? I don't believe that PR would need this functionality but I'm trying to remove any significant dependency on Nuklear's Font handling (especially `NK_INCLUDE_FONT_BAKING` because it pulls STB libraries).

demo/sdl3/nuklear_sdl3_renderer.h
zoogies: Nice, I'll go ahead and test this. When you say it isn't perfect, are you still experiencing some undesirable behavior?

zoogies:
@sleeptightAnsiC I'm glad to see someone else thinking about this PR!

IMO the biggest unknown right now is solving how we can call `SDL_StartTextInput` and `SDL_StopTextInput` in `nk_input_focus` and `nk_input_unfocus`, respectively.

From the original thread, I believe everything else is reasonably resolved except for UTF-8 support in SDL_EVENT_TEXT_INPUT.

zoogies:
> , one could use nk_*_init() and pass custom nk_allocator to them. Adding new macros while this whole nk_allocator mechanism exists feels quite confusing

agreed, and your solution seems reasonable.

@RobLoach thoughts?

zoogies:
> Small thing that I'm trying to solve: How can I get SDL's internal debug font (the one from [SDL_RenderDebugText](https://wiki.libsdl.org/SDL3/SDL_RenderDebugText)) and use it inside of Nuklear?

Interesting, although likely not relevant to this exact PR (but it is a new feature in SDL3 iirc)

You might have luck looking at [the atlas implementation](https://github.com/libsdl-org/SDL/blob/79b881fc2e9174f1e103f6064d7c1bb937284156/src/render/SDL_render.c#L5703), but there are definitely a few downsides to using this: (from [the docs](https://wiki.libsdl.org/SDL3/SDL_RenderDebugText))

- It accepts UTF-8 strings, but will only renders ASCII characters.
- It has a single, tiny size (8x8 pixels). One can use logical presentation or scaling to adjust it, but it will be blurry.
- It uses a simple, hardcoded bitmap font. It does not allow different font selections and it does not support truetype, for proper scaling.
- It does no word-wrapping and does not treat newline characters as a line break. If the text goes out of the window, it's gone.

demo/sdl3/nuklear_sdl3_renderer.h
sleeptightAnsiC: Per our discussion on Discord, this is how SDL2 demo handles double-click and this is what works the best:

https://github.com/Immediate-Mode-UI/Nuklear/blob/171090c8e21584330207544dbde5e7c1ec581135/demo/sdl_renderer/nuklear_sdl_renderer.h#L358-L361

So better fix would be:

```diff
diff --git a/demo/sdl3/nuklear_sdl3_renderer.h b/demo/sdl3/nuklear_sdl3_renderer.h
index c57a423..b9b3a29 100644
--- a/demo/sdl3/nuklear_sdl3_renderer.h
+++ b/demo/sdl3/nuklear_sdl3_renderer.h
@@ -332,8 +332,7 @@ nk_sdl_handle_event(struct nk_context* ctx, SDL_Event *evt)
                     case SDL_BUTTON_LEFT:
                         if (evt->button.clicks > 1)
                             nk_input_button(ctx, NK_BUTTON_DOUBLE, x, y, down);
-                        else
-                            nk_input_button(ctx, NK_BUTTON_LEFT, x, y, down);
+                        nk_input_button(ctx, NK_BUTTON_LEFT, x, y, down);
                         break;
                     case SDL_BUTTON_MIDDLE: nk_input_button(ctx, NK_BUTTON_MIDDLE, x, y, down); break;
                     case SDL_BUTTON_RIGHT:  nk_input_button(ctx, NK_BUTTON_RIGHT, x, y, down); break;
```

Basically, repetitive clicks always send single-click + double-click. However, I'm unsure how widgets with special double-click interactions would work in some edge-cases.

The best approach would be to use timer/delay and custom logic, but this would be way harder to implement, especially with new SDL_app abstraction. Btw, GLFW demos do use delay in this case: https://github.com/Immediate-Mode-UI/Nuklear/blob/171090c8e21584330207544dbde5e7c1ec581135/demo/glfw_opengl2/nuklear_glfw_gl2.h#L280-L286

zoogies: Have you had the chance to test this? It wouldn't be too hard to add a similar delay tracking variable into this demo, assuming the double click does have unintended interactions with widgets

demo/sdl3/nuklear_sdl3_renderer.h
```diff
@@ -0,0 +1,388 @@
...
+};
+
+struct nk_sdl_vertex {
+    float position[2];
+    float uv[2];
+    float col[4];
+};
+
+struct nk_sdl {
+    SDL_Window *win;
```
sleeptightAnsiC, line 58: Does it make sense to keep reference to `SDL_Window`?

This demo does not use it anywhere (and even if it would, it probably shouldn't). Back in SDL2 demos, they were using it in `nk_sdl_handle_grab` for `SDL_WarpMouseInWindow` [(link)](
https://github.com/Immediate-Mode-UI/Nuklear/blob/171090c8e21584330207544dbde5e7c1ec581135/demo/sdl_renderer/nuklear_sdl_renderer.h#L288)

Unless this reference is meant to be used in the future, maybe we can simply remove it?

zoogies: If it's unused, I would remove it

demo/sdl3/nuklear_sdl3_renderer.h
sleeptightAnsiC: > Have you had the chance to test this?

I've only tested the SDL2-demo way. Haven't messed with the timer.

> It wouldn't be too hard to add a similar delay tracking variable into this demo

I think that would be too "specific" for a demo. The easiest way to implement this is by using [SDL_GetTicks](https://wiki.libsdl.org/SDL3/SDL_GetTicks). To be honest, I don't really like how GLFW-demo implements this, as it may detect two quick clicks with movement between them as double-click.

EDIT: after thinking about this, SDL already passes the timestamp with the Event, so this should be trivial to implement

demo/sdl3/nuklear_sdl3_renderer.h
sleeptightAnsiC: > If it's unused, I would remove it

Just to confirm that it's only being assigned but never really used anywhere:

<details><summary>grep</summary>
<p>

```sh
/tmp/Nuklear $ grep -n '[^a-z]win[^a-z]' ./demo/sdl3/nuklear_sdl3_renderer.h
19:NK_API struct nk_context*   nk_sdl_init(SDL_Window *win, SDL_Renderer *renderer);
58:    SDL_Window *win;
223:nk_sdl_init(SDL_Window *win, SDL_Renderer *renderer)
226:    NK_ASSERT(win);
231:    sdl->win = win;
```

</p>
</details>

Working patch:

<details><summary>diff</summary>
<p>

```diff
diff --git a/demo/sdl3/nuklear_sdl3_renderer.c b/demo/sdl3/nuklear_sdl3_renderer.c
index b407af0..bc3433c 100644
--- a/demo/sdl3/nuklear_sdl3_renderer.c
+++ b/demo/sdl3/nuklear_sdl3_renderer.c
@@ -126,7 +126,7 @@ SDL_AppResult SDL_AppInit(void** appstate, int argc, char* argv[]) {
         font_scale = scale_y;
     }
 
-    struct nk_context* ctx = nk_sdl_init(appContext->window, appContext->renderer);
+    struct nk_context* ctx = nk_sdl_init(appContext->renderer);
     appContext->ctx = ctx;
 
     {
diff --git a/demo/sdl3/nuklear_sdl3_renderer.h b/demo/sdl3/nuklear_sdl3_renderer.h
index c57a423..69cc474 100644
--- a/demo/sdl3/nuklear_sdl3_renderer.h
+++ b/demo/sdl3/nuklear_sdl3_renderer.h
@@ -16,7 +16,7 @@
 #error "nuklear_sdl3_renderer requires at least SDL 3.0.0"
 #endif
 
-NK_API struct nk_context*   nk_sdl_init(SDL_Window *win, SDL_Renderer *renderer);
+NK_API struct nk_context*   nk_sdl_init(SDL_Renderer *renderer);
 NK_API struct nk_font_atlas* nk_sdl_font_stash_begin(struct nk_context* ctx);
 NK_API void                 nk_sdl_font_stash_end(struct nk_context* ctx);
 NK_API int                  nk_sdl_handle_event(struct nk_context* ctx, SDL_Event *evt);
@@ -55,7 +55,6 @@ struct nk_sdl_vertex {
 };
 
 struct nk_sdl {
-    SDL_Window *win;
     SDL_Renderer *renderer;
     struct nk_sdl_device ogl;
     struct nk_context ctx;
@@ -220,15 +219,13 @@ nk_sdl_clipboard_copy(nk_handle usr, const char *text, int len)
 }
 
 NK_API struct nk_context*
-nk_sdl_init(SDL_Window *win, SDL_Renderer *renderer)
+nk_sdl_init(SDL_Renderer *renderer)
 {
     struct nk_sdl* sdl;
-    NK_ASSERT(win);
     NK_ASSERT(renderer);
     sdl = SDL_malloc(sizeof(struct nk_sdl));
     NK_ASSERT(sdl);
     nk_zero(sdl, sizeof(struct nk_sdl));
-    sdl->win = win;
     sdl->renderer = renderer;
     nk_init_default(&sdl->ctx, 0);
     sdl->ctx.userdata = nk_handle_ptr((void*)sdl);
```

</p>
</details>

sleeptightAnsiC: Ahhh... looking through other PRs/Issues such as https://github.com/Immediate-Mode-UI/Nuklear/pull/772#issuecomment-2691573311 https://github.com/Immediate-Mode-UI/Nuklear/issues/608 https://github.com/Immediate-Mode-UI/Nuklear/pull/609 gave me more idea about why the reference was there in the first place.

Maybe we should actually add `nk_sdl_handle_grab` back to this demo? Removing reference to `SDL_Window` still wouldn't be a problem, since we could ask for it when calling `nk_sdl_handle_grab`

zoogies: I can check when I get back to my PC, but iirc the SDL2 demo had some broken grabbing behavior, and I think I have that modified/disabled in my working copy

sleeptightAnsiC:
Something that I noticed while looking at other similar PR https://github.com/Immediate-Mode-UI/Nuklear/pull/825#discussion_r2331685368 :

> Is `sdl.ctx.delta_time_seconds` being used anywhere? I think that #779 is missing this one.
> [...]
> EDIT: ahh, now I understand, Nuklear needs this for hiding scroll bar!

see [src/nuklear_panel.c#L493](https://github.com/Immediate-Mode-UI/Nuklear/blob/171090c8e21584330207544dbde5e7c1ec581135/src/nuklear_panel.c#L493)

So we forgot to set `sdl.ctx.delta_time_seconds` (?)

sleeptightAnsiC:
> > Small thing that I'm trying to solve: How can I get SDL's internal debug font (the one from [SDL_RenderDebugText](https://wiki.libsdl.org/SDL3/SDL_RenderDebugText)) and use it inside of Nuklear? I don't believe that PR would need this functionality but I'm trying to remove any significant dependency on Nuklear's Font handling (especially NK_INCLUDE_FONT_BAKING because it pulls STB libraries).
> 
> Interesting, although likely not relevant to this exact PR (but it is a new feature in SDL3 iirc) You might have luck looking at [the atlas implementation](https://github.com/libsdl-org/SDL/blob/79b881fc2e9174f1e103f6064d7c1bb937284156/src/render/SDL_render.c#L5703), but there are definitely a few downsides to using this: (from [the docs](https://wiki.libsdl.org/SDL3/SDL_RenderDebugText))

Got it! :smiley: 

I was able to implement this by recreating the atlas texture with SDL_RenderDebugText. The code is not perfect and needs some additional work but should be enough for anyone who wants to display simple debug font.

<details><summary>screenshot</summary>
<p>

image

</p>
</details>

<details><summary>code</summary>
<p>

```diff
diff --git a/demo/sdl3/nuklear_sdl3_renderer.c b/demo/sdl3/nuklear_sdl3_renderer.c
index b407af0..1d63d7b 100644
--- a/demo/sdl3/nuklear_sdl3_renderer.c
+++ b/demo/sdl3/nuklear_sdl3_renderer.c
@@ -129,6 +129,7 @@ SDL_AppResult SDL_AppInit(void** appstate, int argc, char* argv[]) {
     struct nk_context* ctx = nk_sdl_init(appContext->window, appContext->renderer);
     appContext->ctx = ctx;
 
+#if 0
     {
         struct nk_font_atlas *atlas;
         struct nk_font_config config = nk_font_config(0);
@@ -152,6 +153,9 @@ SDL_AppResult SDL_AppInit(void** appstate, int argc, char* argv[]) {
         /*nk_style_load_all_cursors(ctx, atlas->cursors);*/
         nk_style_set_font(ctx, &font->handle);
     }
+#else
+    nk_sdl_style_set_font_debug(ctx);
+#endif
 
     return SDL_APP_CONTINUE;
 }
diff --git a/demo/sdl3/nuklear_sdl3_renderer.h b/demo/sdl3/nuklear_sdl3_renderer.h
index c57a423..dc22ded 100644
--- a/demo/sdl3/nuklear_sdl3_renderer.h
+++ b/demo/sdl3/nuklear_sdl3_renderer.h
@@ -24,6 +24,7 @@ NK_API void                 nk_sdl_render(struct nk_context* ctx, enum nk_anti_a
 NK_API void                 nk_sdl_shutdown(struct nk_context* ctx);
 NK_API nk_handle            nk_sdl_userdata(struct nk_context* ctx);
 NK_API void                 nk_sdl_set_userdata(struct nk_context* ctx, nk_handle userdata);
+NK_API void                 nk_sdl_style_set_font_debug(struct nk_context* ctx);
 
 #endif /* NK_SDL3_RENDERER_H_ */
 
@@ -371,7 +372,8 @@ void nk_sdl_shutdown(struct nk_context* ctx)
     sdl = (struct nk_sdl*)ctx->userdata.ptr;
     NK_ASSERT(sdl);
 
-    nk_font_atlas_clear(&sdl->atlas);
+    if (sdl->atlas.font_num > 0)
+        nk_font_atlas_clear(&sdl->atlas);
     nk_buffer_free(&sdl->ogl.cmds);
 
     if (sdl->ogl.font_tex != NULL) {
@@ -384,5 +386,116 @@ void nk_sdl_shutdown(struct nk_context* ctx)
     nk_free(ctx);
 }
 
+/* FIXME: width/height of the debug font atlas (integer)
+ *        Probably needs better name... */
+#define WH (10)
+
+NK_INTERN float
+nk_sdl_query_debug_font_width(nk_handle handle, float height,
+                              const char *text, int len)
+{
+    NK_UNUSED(handle);
+    NK_UNUSED(text);
+    /* FIXME: does this thing need to handle newlines and tabs ??? */
+    return height * len;
+}
+
+NK_INTERN void
+nk_sdl_query_debug_font_glypth(nk_handle handle, float font_height,
+                               struct nk_user_font_glyph *glyph,
+                               nk_rune codepoint, nk_rune next_codepoint)
+{
+    char ascii;
+    int idx, x, y;
+    NK_UNUSED(next_codepoint);
+
+    ascii = (codepoint > (nk_rune)'~' || codepoint < (nk_rune)' ')
+            ? '?' : (char)codepoint;
+    NK_ASSERT(ascii <= '~' && ascii >= ' ');
+
+    idx = (int)(ascii - ' ');
+    NK_ASSERT(idx >= 0 && idx <= (WH * WH));
+
+    x = idx / WH;
+    y = idx % WH;
+
+    glyph->height = font_height;
+    glyph->width = font_height;
+    glyph->xadvance = font_height;
+    glyph->uv[0].x = (float)(x + 0) / WH;
+    glyph->uv[0].y = (float)(y + 0) / WH;
+    glyph->uv[1].x = (float)(x + 1) / WH;
+    glyph->uv[1].y = (float)(y + 1) / WH;
+    glyph->offset.x = 0.0f;
+    glyph->offset.y = 0.0f;
+}
+
+NK_API void
+nk_sdl_style_set_font_debug(struct nk_context* ctx)
+{
+    struct nk_user_font* font;
+    struct nk_sdl* sdl;
+    SDL_Surface *surface;
+    SDL_Renderer *renderer;
+    char buf[2];
+    int x, y;
+    bool success;
+    NK_ASSERT(ctx);
+
+    /* FIXME: For now, use another software Renderer just to make sure
+     *        that we won't change any state in the main Renderer,
+     *        but it would be nice to reuse the main one */
+    surface = SDL_CreateSurface(WH * SDL_DEBUG_TEXT_FONT_CHARACTER_SIZE,
+                                WH * SDL_DEBUG_TEXT_FONT_CHARACTER_SIZE,
+                                SDL_PIXELFORMAT_RGBA32);
+    NK_ASSERT(surface);
+    renderer = SDL_CreateSoftwareRenderer(surface);
+    NK_ASSERT(renderer);
+    success = SDL_SetRenderDrawColor(renderer, 0xFF, 0xFF, 0xFF, 0xFF);
+    NK_ASSERT(success);
+
+    /* SPACE is the first printable ASCII character */
+    buf[0] = ' ';
+    buf[1] = '\0';
+    for (x = 0; x < WH; x++)
+    {
+        for (y = 0; y < WH; y++)
+        {
+            success = SDL_RenderDebugText(renderer,
+                                          (float)(x * SDL_DEBUG_TEXT_FONT_CHARACTER_SIZE),
+                                          (float)(y * SDL_DEBUG_TEXT_FONT_CHARACTER_SIZE),
+                                          buf);
+            NK_ASSERT(success);
+
+            buf[0]++;
+            /* TILDE is the last printable ASCII character */
+            if (buf[0] > '~')
+                break;
+        }
+    }
+    success = SDL_RenderPresent(renderer);
+    NK_ASSERT(success);
+
+    font = SDL_malloc(sizeof(*font));
+    NK_ASSERT(font);
+    font->userdata.ptr = NULL;
+    font->height = SDL_DEBUG_TEXT_FONT_CHARACTER_SIZE;
+    font->width = &nk_sdl_query_debug_font_width;
+    font->query = &nk_sdl_query_debug_font_glypth;
+
+    /* HACK: nk_sdl_device_upload_atlas turns pixels into SDL_Texture
+     *       and sets said Texture into sdl->ogl.font_tex
+     *       then nk_sdl_render expects same Texture at font->texture */
+    sdl = (struct nk_sdl*)ctx->userdata.ptr;
+    nk_sdl_device_upload_atlas(&sdl->ctx, surface->pixels, surface->w, surface->h);
+    font->texture.ptr = sdl->ogl.font_tex;
+
+    /* FIXME: font is passed but never free'd (memleak) */
+    nk_style_set_font(ctx, font);
+
+    SDL_DestroyRenderer(renderer);
+    SDL_DestroySurface(surface);
+}
+
 #endif /* NK_SDL3_RENDERER_IMPLEMENTATION_ONCE */
 #endif /* NK_SDL3_RENDERER_IMPLEMENTATION */
 ```

</p>
</details>

File: demo/sdl3/nuklear_sdl3_renderer.h [hunk 16]
```diff
@@ -0,0 +1,388 @@
...
+
+NK_API
+void nk_sdl_shutdown(struct nk_context* ctx)
+{
+    struct nk_sdl* sdl;
+    NK_ASSERT(ctx);
+    sdl = (struct nk_sdl*)ctx->userdata.ptr;
+    NK_ASSERT(sdl);
+
+    nk_font_atlas_clear(&sdl->atlas);
```
sleeptightAnsiC, line 374: When I was messing with custom font (mentioned [here](https://github.com/Immediate-Mode-UI/Nuklear/pull/779#issuecomment-3268640887)), this line would lead to the crash unless I make a check for any existing font:
```diff
-    nk_font_atlas_clear(&sdl->atlas);
+    if (sdl->atlas.font_num > 0)
+        nk_font_atlas_clear(&sdl->atlas);
```
However, this feels like something that should be fixed inside of [nk_font_atlas_clear](https://github.com/Immediate-Mode-UI/Nuklear/blob/171090c8e21584330207544dbde5e7c1ec581135/src/nuklear_font.c#L1334)

demo/sdl3/nuklear_sdl3_renderer.h
sleeptightAnsiC: Note: we will also need Window reference for SDL_StartTextInput and SDL_StopTextInput, if we ever use them.

EDIT: we should also make a check in `nk_sdl_handle_event` to ensure that the Window sending event is the same as the one in `nk_sdl` (and make an early bail, if it's not the same)

sleeptightAnsiC:
> Figure out the best method around SDL_StartTextInput() usage

> IMO the biggest unknown right now is solving how we can call SDL_StartTextInput and SDL_StopTextInput in nk_input_focus and nk_input_unfocus, respectively.



> > > Are you proposing some kind of event callback system for them?
> >
> > That's the first thing that comes to mind. I don't have deep knowledge of the Nuklear codebase, so I wasn't sure if that's done in other places, or what the best way to integrate something like that would be. Do you have any opinions?
>
> Maybe a strict callback isn't necessary as that would incur a small but present overhead for all backends.

I've started experimenting with this and found small solution.
This is very much work in progress but already seems to behave just fine.
I have to test few cases to make sure it isn't buggy.

<details><summary>patch</summary>
<p>

```diff
diff --git a/demo/sdl3/nuklear_sdl3_renderer.h b/demo/sdl3/nuklear_sdl3_renderer.h
index c57a423..1eb61f1 100644
--- a/demo/sdl3/nuklear_sdl3_renderer.h
+++ b/demo/sdl3/nuklear_sdl3_renderer.h
@@ -274,6 +274,20 @@ nk_sdl_handle_event(struct nk_context* ctx, SDL_Event *evt)
     NK_ASSERT(ctx);
     NK_ASSERT(evt);
 
+    static bool was_active = false;
+    const bool active = (ctx->active && ctx->active->edit.active) ||
+                        (ctx->active && ctx->active->popup.win && ctx->active->popup.win->edit.active);
+    SDL_Window *win = SDL_GetWindowFromEvent(evt);
+    struct nk_sdl* sdl = (struct nk_sdl*)ctx->userdata.ptr;
+    if (active != was_active && win == sdl->win)
+    {
+        if (!was_active && active)
+            SDL_StartTextInput(win);
+        else if (was_active && !active)
+            SDL_StopTextInput(win);
+        was_active = active;
+        SDL_LogInfo(SDL_LOG_CATEGORY_APPLICATION, "text input %s", active ? "started" : "stopped");
+    }
+
     switch(evt->type)
     {
         case SDL_EVENT_KEY_UP: /* KEYUP & KEYDOWN share same routine */
```

</p>
</details> 

<details><summary>video</summary>
<p>

https://github.com/user-attachments/assets/0882456f-55c1-48ad-9c3c-1d420178f297

</p>
</details> 

EDIT: I had to add additional branch for detecting Input from Popup. I tweaked the patch by hand, I hope it still applies.

vsaint1:
Working with pixel art or low art res viewport requires changing the viewport, if you call  SDL_SetRenderLogicalPresentation  the motion position events will be wrong since the viewport != window size, so here is the simplest solution.

Good job y'all, since i dont have much knowledge about patching, i'll just post the solution fix here if anyone encounter the same problem as me:

```cpp
 
 // must be called int `nk_sdl_handle_event` 
 SDL_ConvertEventToRenderCoordinates(sdl->renderer, evt);
 ```

**Comment by sleeptightAnsiC** [comment]
I'm thinking about creating new PR with all the changes mentioned here. I'm pretty sure, I fixed most of the stuff from the comments. What's left is just merging, some additional polishing, adding Makefile (instead of CMake) and some minor stuff.

Would anyone here be interested? or shall we keep our focus on this PR?

CC @RobLoach (sorry for pinging but I think you silenced this thread and this message is quite important)

**Comment by RobLoach** [comment]
Happy with anything that helps this move forwards.

**Comment by zoogies** [comment]
Rob, do you have any feedback on the state of this PR? I want to make sure we have a sense of what needs to be changed before this would be considered for a merge. Thanks!

**Comment by RobLoach** [comment]
@sleeptightAnsiC You had some additions over at https://github.com/Immediate-Mode-UI/Nuklear/compare/master...sleeptightAnsiC:Nuklear:personal ... Anything you think we should bring in? Would you like to add these in once the PR is merged, or prior?

**Comment by sleeptightAnsiC** [comment]
> @sleeptightAnsiC You had some additions over at https://github.com/Immediate-Mode-UI/Nuklear/compare/master...sleeptightAnsiC:Nuklear:personal ... Anything you think we should bring in? Would you like to add these in once the PR is merged, or prior?

@RobLoach That's my working branch where I test things and push stuff for my personal use. No worries, I will upstream all of those changes as separate PRs/Issues when I consider them stable and reasonable. Everything related to *this* PR is already mentioned in comments above.

That said, I will create a new PR for the SDL3 renderer demo as I mentioned in https://github.com/Immediate-Mode-UI/Nuklear/pull/779#issuecomment-3429841640 . I just wanted to wait a bit and give others in this thread a chance to share any final thoughts, since there were a lot of people involved here.

File: demo/sdl3/nuklear_sdl3_renderer.h (hunk 5 above)
Comment (sleeptightAnsiC) [reply demo/sdl3/nuklear_sdl3_renderer.h:354]: Yup, this line causes heap-buffer-overflow. I just caught it when running the demo with address sanitizer. The funny thing is that `text` field is located at the very end of `SDL_TextInputEvent` struct, so it only reads zeros from the padding, however, this is still a bug. Nice catch.

I am a bit confused though. I haven't found any info about this change in https://github.com/libsdl-org/SDL/blob/main/docs/README-migration.md#sdl_keyboardh

File: demo/sdl3/nuklear_sdl3_renderer.h (hunk 1 above)
Comment (sleeptightAnsiC) [reply demo/sdl3/nuklear_sdl3_renderer.h]: yup, this is broken when using non-ASCII characters

(also linking for cross reference because it might be related: https://github.com/Immediate-Mode-UI/Nuklear/pull/841 )

**Comment by rswinkle** [comment]
> > Figure out the best method around SDL_StartTextInput() usage
> 
> > IMO the biggest unknown right now is solving how we can call SDL_StartTextInput and SDL_StopTextInput in nk_input_focus and nk_input_unfocus, respectively.
> 
> > > > Are you proposing some kind of event callback system for them?
> > > 
> > > 
> > > That's the first thing that comes to mind. I don't have deep knowledge of the Nuklear codebase, so I wasn't sure if that's done in other places, or what the best way to integrate something like that would be. Do you have any opinions?
> > 
> > 
> > Maybe a strict callback isn't necessary as that would incur a small but present overhead for all backends.
> 
> I've started experimenting with this and found small solution. This is very much work in progress but already seems to behave just fine. I have to test few cases to make sure it isn't buggy.

I'm sorry for chiming in so late but I dealt with some headaches with getting this to work in my own SDL3 backend back in June and want to know if you ran into the same issues.

Does it handle multiple windows correctly? Have you tried it on Android/mobile? Honestly I can't remember everything I ran into exactly but based on my commit message the multiple windows thing was a issue that led to my current version.

My current method required adding a few lines to nuklear proper but it simplified the backend code to this:
```
	// Nuklear keeps track of it but we only use it if we need it for our
	// backend/platform
	if (ctx->input.text_input_active != ctx->input.text_input_active_prev) {
		if (ctx->input.text_input_active) {
			SDL_PropertiesID props = SDL_CreateProperties();
			SDL_SetNumberProperty(props, SDL_PROP_TEXTINPUT_CAPITALIZATION_NUMBER, SDL_CAPITALIZE_NONE);
			//SDL_StartTextInput(g->win);
			//props = SDL_PROP_TEXTINPUT_ANDROID_INPUTTYPE_NUMBER |
			//props = SDL_PROP_TEXTINPUT_CAPITALIZATION_NUMBER |
			SDL_StartTextInputWithProperties(g->win, props);
		} else {
			SDL_StopTextInput(g->win);
		}
	}
```
Obviously the general case would look cleaner/simpler but you get the idea.  BTW, how do you do the collapsible patch in a comment?  Is that automatic when you upload a patch file?

**Comment by sleeptightAnsiC** [comment]
@rswinkle 

> Does it handle multiple windows correctly?

It handles whatever `SDL_Window` is referenced by `nk_sdl` context. This should work just fine if you have one context per window, but will most likely fail if you need to share and/or pass the same context between several windows.

> Have you tried it on Android/mobile?

No, I haven't tried it (yet). I'm searching for any platform that utilizes Start/StopTextInput calls and would let me test it quickly. Maybe Steam in BigPicture mode? I've heard that [Steam-runtime ships with a modified version of SDL](https://wiki.libsdl.org/SDL3/README-steamos) with some virtual-keyboard support.

I'm also thinking about hooking [SDL_SetTextInputArea](https://wiki.libsdl.org/SDL3/SDL_SetTextInputArea) into this demo, but for this one I definitely need a better testing environment.

> Honestly I can't remember everything I ran into exactly but based on my commit message the multiple windows thing was a issue that led to my current version.

Hard to tell. If you've used the code from this PR in unchanged form, then it had no chance to work out of the box in multi-window scenario. The working copy I'm using right now has a lot of fixes for this, so maybe I fixed some of those issues already. I might be able to help with more specific info.

> Obviously the general case would look cleaner/simpler but you get the idea.

Yea, I understand. I will be posting a new PR in few days. I'm working on it _right now_, but I've encountered some issues. It's taking way more time than I expected... and I don't want to Draft it, because I already have to keep track of several related PRs/Issues (https://github.com/Immediate-Mode-UI/Nuklear/pull/779, https://github.com/Immediate-Mode-UI/Nuklear/pull/772, https://github.com/Immediate-Mode-UI/Nuklear/pull/834, https://github.com/Immediate-Mode-UI/Nuklear/pull/825, https://github.com/Immediate-Mode-UI/Nuklear/issues/768) :weary:

>  BTW, how do you do the collapsible patch in a comment? Is that automatic when you upload a patch file?

Just type `/` on the beginning of newline, and you will get a lot of Markdown templates. The one you're asking for is `/Details`. For syntax highlight for patches, you need to use a code-block like so:
````
```diff
patch content here
```
````
or `patch` instead of `diff`, but `diff` is enough when its content was generated with `git diff <file>`

(there are a lot of features like this one, take a look at some docs https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github )

rswinkle:
> @rswinkle
> 
> > Does it handle multiple windows correctly?
> 
> It handles whatever `SDL_Window` is referenced by `nk_sdl` context. This should work just fine if you have one context per window, but will most likely fail if you need to share and/or pass the same context between several windows.

Haha, I meant multiple nuklear windows.  The issue was related to how nuklear keeps track of its internal state relative to multiple windows and clicking between them. Going from an input field in one to an input field in another and every other possible focus transition you can think of but between windows rather than within one.

>Hard to tell. If you've used the code from this PR in unchanged form, then it had no chance to work out of the box in multi-window scenario. The working copy I'm using right now has a lot of fixes for this, so maybe I fixed some of those issues already. I might be able to help with more specific info.

I think I took what the latest version was back then and made lots of changes/fixes. I had recently done several pull requests that touched most/all of the backends so some of that was just bringing those same fixes to the SDL3 backend.

>Yea, I understand

I completely understand and empathize. It gets overwhelming fast when you're working on a simple fix/PR and then you find another issue, and another, and oh we should totally change this, but these should probably be different PRs, and that would be too big of a change without pre-approval, and then other stuff is being changed out from under you as well and you just want to be done.  sigh

> Just type `/`

Thanks, my google fu was really weak at 3 am last night. Will try it out later.

Also, I can probably test it out on Android once it's ready.  My project was primarily for Android and SDL3's Android build script works pretty well for simple structured projects.

demo/sdl3/nuklear_sdl3_renderer.h
sleeptightAnsiC: I've thought about this problem for a bit, and I think that if someone needs to use `userdata` in `nk_context`, that would probably be for something that demo/renderer already handles. In other cases, the use of `nk_sdl_set_userdata()` should be fine.

sleeptightAnsiC:
I think we can close this. All work has moved to https://github.com/Immediate-Mode-UI/Nuklear/pull/852 and was already merged.

PR #781, opened 2025-02-22
Add NumPad Enter to other backends

rswinkle:
@RobLoach you merged my last pull request so I'm taking that as approval to make similar changes to the other backends as appropriate.  I've done so to all except rawfb since that's kind of special/unfinished.

As I said before, I'm unable to easily test the Windows backends but they all use the same input method so if one works they should all work.

In the process of going through everything I found a couple issues I hope someone else can verify:

* glfw_gl4 demo crashes, something to do with the shaders, I didn't dig into it
* SFML Makefiles fail on linux, removing the "-s" suffixes on the libraries fixes it for me
* All SFML backends have a weird input bug

### DIFF
--- a/demo/allegro5/main.c
+++ b/demo/allegro5/main.c
@@ -120,7 +120,7 @@ int main(void)
     al_register_event_source(event_queue, al_get_mouse_event_source());
     al_register_event_source(event_queue, al_get_keyboard_event_source());
 
-    font = nk_allegro5_font_create_from_file("../../../extra_font/Roboto-Regular.ttf", 12, 0);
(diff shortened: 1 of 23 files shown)


RobLoach (approved):
Looks good to me. Thanks a lot for the add.

> SFML backends have a weird input bug

Do say more. Haven't tried sfml in a while.

RobLoach:
Definitely seems something with SFML. I haven't seen this happen in the other renderers.

RobLoach:
Made a follow up for that over at https://github.com/Immediate-Mode-UI/Nuklear/issues/783 . Overall, this change looks good. Thanks a lot.

PR #785, opened 2025-03-01
Add support for right control to sdl backends

rswinkle:
Can't believe I never noticed this issue before.

Pretty self explanatory: use SDL_GetModState() to get whether either control is down and use that instead of the key state array.

### DIFF
--- a/demo/sdl_opengl2/nuklear_sdl_gl2.h
+++ b/demo/sdl_opengl2/nuklear_sdl_gl2.h
@@ -263,6 +263,7 @@ NK_API int
 nk_sdl_handle_event(SDL_Event *evt)
 {
     struct nk_context *ctx = &sdl.ctx;
+    int ctrl_down = SDL_GetModState() & (KMOD_LCTRL | KMOD_RCTRL);
 
     switch(evt->type)
     {
@@ -270,7 +271,6 @@ nk_sdl_handle_event(SDL_Event *evt)
         case SDL_KEYDOWN:
             {
                 int down = evt->type == SDL_KEYDOWN;
-                const Uint8* state = SDL_GetKeyboardState(0);
                 switch(evt->key.keysym.sym)
                 {
                     case SDLK_RSHIFT: /* RSHIFT & LSHIFT share same routine */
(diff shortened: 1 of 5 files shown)


RobLoach:
Should we take the same `SDL_GetModState()` approach for the shift/alt keys too? Or do you think that's unnecessary.

rswinkle:
As far as I can tell we don't use ALT at all but I guess we do use SHIFT and track it specifically unlike CTRL.  I think it makes more sense to leave it how it is though because it would actually be slightly more work CPU wise to change it.  Either we have a shift_down variable/expression and pass it to nk_input in the same place we have now (when the key event is a SHIFT) which makes the shift_down calculation completely redundant with down, or we remove SHIFT from the switch and call nk_input for shift above it which means many many unnecessary function calls.

That was/is one of the issues with GLFW backends.  nk_input is really meant for events and so calling it every frame for a state is just inefficient even if it's sometimes functionally the same for the user.  I was actually debating adding all of these to the key callback and guarding all of them for that reason even though the user can't tell the difference:

`    nk_input_key(ctx, NK_KEY_TEXT_START, glfwGetKey(win, GLFW_KEY_HOME) == GLFW_PRESS);
    nk_input_key(ctx, NK_KEY_TEXT_END, glfwGetKey(win, GLFW_KEY_END) == GLFW_PRESS);
    nk_input_key(ctx, NK_KEY_SCROLL_START, glfwGetKey(win, GLFW_KEY_HOME) == GLFW_PRESS);
    nk_input_key(ctx, NK_KEY_SCROLL_END, glfwGetKey(win, GLFW_KEY_END) == GLFW_PRESS);
    nk_input_key(ctx, NK_KEY_SHIFT, glfwGetKey(win, GLFW_KEY_LEFT_SHIFT) == GLFW_PRESS||
                                    glfwGetKey(win, GLFW_KEY_RIGHT_SHIFT) == GLFW_PRESS);
`
and same with these but that'd be a bit more complicated
`
        nk_input_key(ctx, NK_KEY_COPY, 0);
        nk_input_key(ctx, NK_KEY_PASTE, 0);
        nk_input_key(ctx, NK_KEY_CUT, 0);
`

Speaking of though, I've been using SDL forever and never knew or forgot a long time ago about the mod member of the key event so thanks for pinging me so I would see that.  In looking into that after I saw another minor improvement I should have been aware of in the enum SDL_Keymod:

`KMOD_CTRL = KMOD_LCTRL | KMOD_RCTRL,`

And similarly for all the other 2 sided modifiers.  D`oh!

So that's a small improvement we can make to all 6 SDL backends, not sure it's even worth a separate pull request though.  I can do it if you want though.

PR #786, opened 2025-03-01
Fix glfw repeat behavior

rswinkle:
I was the [author](https://github.com/vurtun/nuklear/pull/765) for the [NK_KEYSTATE_BASED_INPUT](https://github.com/Immediate-Mode-UI/Nuklear/commit/69a2d50c9b7c4fd60e40db67cf3f8bc1d5e474d1) fix for the fix for the too-fast repeat behavior on glfw backends.

I was never really happy with it as, while it fixed repeat behavior for everyone else, it broke it for glfw.  If I had been less focused on Nuklear and more on GLFW (which I had never used before and still have almost 0 experience with) I would have realized that this is not a Nuklear problem and should not be handled by Nuklear proper but by the backend authors.

If I had paid attention I would have noticed we already handle the same problem with text input: we use the char callback and buffer the results to send all at once between nk_input_begin()/end().  We can do the same thing for special keys like Delete, Backspace, Left/Right etc.  They have a separate callback but the idea is the same, just store the values from the callback, send any collected events to Nuklear, and clear the array after input_end() for the next frame.

While I was making these changes I also added fb_scale to the vulkan backend since it was the only glfw backend that didn't have it.  I'm pretty sure I did it correctly but someone should verify it on a high-DPI monitor.

I wasn't sure if I should label this a draft or not.  There are a couple of minor issues I think might be worth changing* but I think it's mostly sound.

Oh and if this gets approved and this approach of "let the backend work around its own issues" becomes policy, it would obviate the need for #639.  Maybe there's still some rare edge case where it would be useful but for my 2 cents I think we should avoid unnecessary complexity/code in Nuklear proper that doesn't benefit all or almost all backends.  While we're at it we could get rid of NK_KEYSTATE_BASED_INPUT entirely and I could redeem myself completely.

*like I name it key_events[] in the glfw structure but the name the local k_state ...they kind of are both true in context.

### DIFF
--- a/demo/glfw_opengl2/main.c
+++ b/demo/glfw_opengl2/main.c
@@ -20,7 +20,6 @@
 #define NK_INCLUDE_DEFAULT_FONT
 #define NK_IMPLEMENTATION
 #define NK_GLFW_GL2_IMPLEMENTATION
-#define NK_KEYSTATE_BASED_INPUT
 #include "../../nuklear.h"
 #include "nuklear_glfw_gl2.h"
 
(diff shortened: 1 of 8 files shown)


RobLoach:
Thank you.

PR #791, opened 2025-03-11
Use built-in motion handler to circumvent FPS-limit/VSync issue

TheRainTrain:
The mouse delta when using the current nk_sdl_handle_event is inaccurate.  
This is apparent if you limit FPS or enable VSync, then move a window around, the window will lag behind the mouse in certain environments and desync immediately. (Tested on GNOME/Mutter on wayland)  
  
Essentially, the current solution seems to treat the mouse delta as being a guesstimated 20% lower than it is.  
  
The fix is using Nuklear's nk_input_motion instead

### DIFF
--- a/demo/sdl3/nuklear_sdl3_renderer.h
+++ b/demo/sdl3/nuklear_sdl3_renderer.h
@@ -342,11 +342,7 @@ nk_sdl_handle_event(struct nk_context* ctx, SDL_Event *evt)
             return 1;
 
         case SDL_EVENT_MOUSE_MOTION:
-            ctx->input.mouse.prev = ctx->input.mouse.pos;
-            ctx->input.mouse.delta.x = evt->motion.xrel;
-            ctx->input.mouse.delta.y = evt->motion.yrel;
-            ctx->input.mouse.pos.x = evt->motion.x;
-            ctx->input.mouse.pos.y = evt->motion.y;
+            nk_input_motion(ctx, evt->motion.x, evt->motion.y);
             return 1;
 
         case SDL_EVENT_TEXT_INPUT:


demo/sdl3/nuklear_sdl3_renderer.h
```diff
@@ -342,11 +342,7 @@ nk_sdl_handle_event(struct nk_context* ctx, SDL_Event *evt)
             return 1;
 
         case SDL_EVENT_MOUSE_MOTION:
-            ctx->input.mouse.prev = ctx->input.mouse.pos;
-            ctx->input.mouse.delta.x = evt->motion.xrel;
-            ctx->input.mouse.delta.y = evt->motion.yrel;
-            ctx->input.mouse.pos.x = evt->motion.x;
-            ctx->input.mouse.pos.y = evt->motion.y;
+            nk_input_motion(ctx, evt->motion.x, evt->motion.y);
```
RobLoach, line 345: The problem with `nk_input_motion()` is that it uses `int`s. So we lose precision here. Perhaps we introduce a new `nk_input_motionf()` function, or similar?

RobLoach (changes requested):

demo/sdl3/nuklear_sdl3_renderer.h
RobLoach: Or maybe we just need to remove the `.prev` change.
``` suggestion
            ctx->input.mouse.delta.x = evt->motion.xrel;
            ctx->input.mouse.delta.y = evt->motion.yrel;
            ctx->input.mouse.pos.x = evt->motion.x;
            ctx->input.mouse.pos.y = evt->motion.y;
```

TheRainTrain: Your proposed change works in my testing environment. Though I like the idea of using an actual non-demo nk_* function more, imprecision would definitely be an issue, this might be the way to go until the `master` branch implements something more precise.  
  
Maybe `nk_input_motionf` might be a good idea, or `nk_input_motion_precise` to be more clear on why you might want to use it.  
Would there be a practical reason `nk_input_motion` uses ints if a floating point is required for precision in this case?

PR #794, opened 2025-03-25
Fixed msvc and clang warnings

TheMostDiligent:


demo/common/overview.c
```diff
@@ -443,8 +443,8 @@ overview(struct nk_context *ctx)
                     nk_combo_end(ctx);
                 }
                 /* progressbar combobox */
-                sum = prog_a + prog_b + prog_c + prog_d;
-                sprintf(buffer, "%lu", sum);
+                sum = (unsigned long)(prog_a + prog_b + prog_c + prog_d);
+                snprintf(buffer, sizeof(buffer), "%lu", sum);
```
klei1984, line 447: Is it targeted for demos to be also ANSI C89 (ISO C90) compatible?

nuklear.h
klei1984, line 1: This file seems to be the script generated output file that should not be changed by hand. The /src/*.* files are the source code of the library.

demo/common/overview.c
TheMostDiligent: `snprintf` is C99. Is there any compiler an existence that does not support C99 though?

There is a definite issue here and in other similar places: `%lu` defines `unsigned long` argument of size 4. `sizeof(size_t)` is 8 on 64- bit systems, so there is an argument size mismatch.

klei1984: Apparently the reviewer's guide clarifies that for demos newer language features are not discouraged. But the library shall remain fully ANSI C89 compliant.

C99 introduced the `"%zu"` format specifier for `size_t`. I recommend removing the `unsigned long` casts and using `zu`.

RobLoach (changes requested):
Concur. The changes in nuklear.h will have to be applied in the src dir, or we'll lose those chagnes.

RobLoach (changes requested):
There have been some recent changes that may have fixes thiese warninggs. Mind another test?

TheMostDiligent:
There are still these warnings when compiling with MSVC:

```
10>overview.c(447,33): error C2220: the following warning is treated as an error
10>(compiling source file '../../../../../DiligentSamples/Samples/NuklearDemo/src/NuklearDemo.cpp')
10>overview.c(447,33): warning C4477: 'sprintf' : format string '%lu' requires an argument of type 'unsigned long', but variadic argument 1 has type 'size_t'
10>(compiling source file '../../../../../DiligentSamples/Samples/NuklearDemo/src/NuklearDemo.cpp')
10>    overview.c(447,33):
10>    consider using '%zu' in the format string
10>overview.c(459,33): warning C4477: 'sprintf' : format string '%lu' requires an argument of type 'unsigned long', but variadic argument 1 has type 'size_t'
10>(compiling source file '../../../../../DiligentSamples/Samples/NuklearDemo/src/NuklearDemo.cpp')
10>    overview.c(459,33):
10>    consider using '%zu' in the format string
```

TheMostDiligent:
All changes in this PR still apply + there are additional uninitialized struct warnings.

sleeptightAnsiC (changes requested):
@TheMostDiligent 

Shouldn't this be fixed by changing `"%lu"` formatting to the `"%zu"` one? It feels very strange to replace a portable `size_t` with non-portable `unsigned long`. Someone who wrote this used `size_t` for a reason.

The trick with `snprintf()`+`sizeof()` is fine, however, it doesn't seem related to this change (?) Did MSVC warn you about unsafe sprintf usage, or maybe you found it overflowing somewhere? **Note that few other examples are still using sprintf so it doesn't make much sense to only change it here. You may still get warnings elsewhere**

Same with all the changes to the library (seems unrelated). What is the reason for memzero'ing those objects? Did you got any warning about them?

Next time, it would be nice, if you could report such thing as an Issue before submitting any patches, or at least mention which exact warnings are you getting for which exact compiler version (and so on). It's really hard to review without all those details.

TheMostDiligent:
> Did MSVC warn you about unsafe sprintf usage

Yes

> What is the reason for memzero'ing those objects?

Warning about potentially uninitialized variable being used.

Besides the warning above, here are additional warnings that pop up:

10>nuklear.h(30083,1): error C2220: the following warning is treated as an error
10>nuklear.h(30186,1): error C2220: the following warning is treated as an error
10>nuklear.h(30083,1): warning C4701: potentially uninitialized local variable 'sym_background' used
10>nuklear.h(30186,1): warning C4701: potentially uninitialized local variable 'text' used
10>nuklear.h(23886,1): warning C4701: potentially uninitialized local variable 'label' used
10>nuklear.h(25571,1): warning C4701: potentially uninitialized local variable 'text' used

sleeptightAnsiC:
> All changes in this PR still apply

There is a mismatch between files in ./src/* and ./nuklear.h You need to apply the patch to corresponding files under src, and only then rebuild ./nuklear.h with ./src/paq.bat (sorry, we don't have a CI for detecting this at them moment)

> > Did MSVC warn you about unsafe sprintf usage
> 
> Yes

Okey, but could you post it here?

> 
> > What is the reason for memzero'ing those objects?
> 
> Warning about potentially uninitialized variable being used.

Again, could you post it, please?

Also, if there is really an uninitialized variable somewhere, we probably have some bug that we have to fix. Zero initializing could hide this bug (and any similar bugs) from UBSan.

> Besides the warning above, here are additional warnings that pop up:

Too little info to judge but *n*one of those seem fatal. You may have warnings-as-errors turned on (I think it was `/WX` in MSVC ?)

Could you please post full output with all the warnings, and which flags are you using, and more details overall?

TheMostDiligent:
I can't post the full log because these warnings appear in batches - you fix a few and then more appear. The ones I posted above is what I get with the latest version.

If you want to reproduce them, try building the demo in Visual Studio with MSVC with warning level 4 (`/W4`) and warnings as errors (`/WX`).

PR #795, opened 2025-03-28
fix: unintialized data in glyph

oflebbe:
Compiler complained about uninitialized glyph. The array isnt initialized here.

Thanks for maintaining nuklear!


nuklear.h
klei1984, line 1: This file seems to be the script generated output file that should not be changed by hand. The `/src/*.*` files are the source code of the library.

oflebbe: Thanks ! I forgot...

oflebbe: Since I created the PR from wrong branch in fork, I closed the PR and created #797 instead

PR #800, opened 2025-04-05
Revert "Fix bug where width padding was applied twice"

RobLoach:
Going through the commits, I found this commit is what was causing some border issues on the button. Mind testing this out?

Coming from https://github.com/Immediate-Mode-UI/Nuklear/issues/796

This reverts commit 14dd68ea0f4918ec3a77222ddb5c337dff4995f3.

### DIFF
--- a/nuklear.h
+++ b/nuklear.h
@@ -22034,14 +22034,16 @@ nk_layout_row_calculate_usable_space(const struct nk_style *style, enum nk_panel
     float panel_space;
 
     struct nk_vec2 spacing;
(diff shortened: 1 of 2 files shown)


klei1984:
I tested sdl_opengl3 and glfw_opengl4, both with anti aliasing (AA) on and off.

The left side is now not truncated, but the right side conditionally is. On master the following behavior is not observed, on this branch it is consistent on the two demos tested. Screenshot is taken from glfw_opengl4 with AA off:
kép

PR #808, opened 2025-05-24
Add CMake scripts to D3D11/D3D12 demos

IceyPillow:
This change helps other users, one command can generate a VS solution regardless of which Visual Studio IDE users have.

e.g.
cd ./demo/d3d12
cmake -S ./ -B ./cmake

### DIFF
--- a/demo/d3d11/CMakeLists.txt
+++ b/demo/d3d11/CMakeLists.txt
@@ -1,0 +1,15 @@
+cmake_minimum_required(VERSION 4.0)
+set(CMAKE_C_STANDARD 17)
+project(NuklearDemo LANGUAGES C)
+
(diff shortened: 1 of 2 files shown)


igromanru:
I don't know much about CMake. But shouldn't there be a CMakeLists.txt ?
How do I build it with Native Tools Command Prompt from VS 2022?

ModulePillow:
> I don't know much about CMake. But shouldn't there be a CMakeLists.txt ? How do I build it with Native Tools Command Prompt from VS 2022?

VS (exactly, it's MSBuild of VS) is a build system. CMake is a management system for all build systems.

I don't know much about VS Native Command; u can create an empty VS C++ solution on the GUI, and manually put *.h/*.cpp into it.

That's where CMake shines; several lines of CMake scripts define a whole project.

Issue #810, opened 2025-05-27
MinGW64 wrong types for nk_size and nk_ptr


m256i:
On my windows machine using UCRT64-MinGW64

```c
NK_STATIC_ASSERT(sizeof(nk_size) >= sizeof(void *));
NK_STATIC_ASSERT(sizeof(nk_ptr) >= sizeof(void *));
```

both fail as:
```c
#ifndef NK_SIZE_TYPE
#if defined(_WIN64) && defined(_MSC_VER)
#define NK_SIZE_TYPE unsigned __int64
#elif (defined(_WIN32) || defined(WIN32)) && defined(_MSC_VER)
#define NK_SIZE_TYPE unsigned __int32
#elif defined(__GNUC__) || defined(__clang__)
#if defined(__x86_64__) || defined(__ppc64__) || defined(__PPC64__) ||         \
  defined(__aarch64__)
#define NK_SIZE_TYPE unsigned long
#else
#define NK_SIZE_TYPE unsigned int
#endif
#else
#define NK_SIZE_TYPE unsigned long
#endif
#endif
```

aswell as:

```c
#ifndef NK_POINTER_TYPE
#if defined(_WIN64) && defined(_MSC_VER)
#define NK_POINTER_TYPE unsigned __int64
#elif (defined(_WIN32) || defined(WIN32)) && defined(_MSC_VER)
#define NK_POINTER_TYPE unsigned __int32
#elif defined(__GNUC__) || defined(__clang__)
#if defined(__x86_64__) || defined(__ppc64__) || defined(__PPC64__) ||         \
  defined(__aarch64__)
#define NK_POINTER_TYPE unsigned long
#else
#define NK_POINTER_TYPE unsigned int
#endif
#else
#define NK_POINTER_TYPE unsigned long
#endif
#endif
```

are erroneously resolved to ``unsigned long`` because MinGW appears to not define _MSC_VER.

sleeptightAnsiC:
For cross reference: https://github.com/Immediate-Mode-UI/Nuklear/issues/847#issuecomment-3497050976

sleeptightAnsiC:
Since https://github.com/Immediate-Mode-UI/Nuklear/pull/811 was merged, we can close this one, right?

PR #811, opened 2025-05-27
fixed defines for ``NK_SIZE_TYPE`` and ``NK_POINTER_TYPE`` on MinGW and windows clang/clang-cl.

m256i:
Issue as stated in #810 
modified defines for ``NK_SIZE_TYPE`` and ``NK_POINTER_TYPE`` to:
```c
    #elif defined(_WIN64) && (defined(__MINGW64__) || defined(__clang__))
      #define NK_SIZE/POINTER_TYPE unsigned long long
   ...
    #elif (defined(_WIN32) || defined(WIN32)) && (defined(__MINGW32__) || defined(__clang__))
      #define NK_SIZE/POINTER_TYPE unsigned long  
```
to account for MinGW 64 and 32 bit aswell as windows clang 64 and 32 bit and clang-cl 64 and 32 bit. 
i've put together a quick [goldbolt comparison](https://godbolt.org/z/W4YzMP555)
with MinGW gcc + clang and a gnu/linux gcc basecase, to demonstrate that the fix works as intended.

### DIFF
--- a/src/nuklear.h
+++ b/src/nuklear.h
@@ -152,7 +152,7 @@ extern "C" {
       #define NK_UINT32 unsigned int
     #endif
   #endif
-#ifndef NK_SIZE_TYPE
+  #ifndef NK_SIZE_TYPE
     #if defined(_WIN64) && defined(_MSC_VER)
       #define NK_SIZE_TYPE unsigned __int64
     #elif defined(_WIN64) && (defined(__MINGW64__) || defined(__clang__))


src/nuklear.h
```diff
@@ -152,11 +152,15 @@ extern "C" {
       #define NK_UINT32 unsigned int
     #endif
   #endif
-  #ifndef NK_SIZE_TYPE
+   #ifndef NK_SIZE_TYPE
```
sleeptightAnsiC: Looks like this unintentional space snuck into your changes. Could you remove it, please? 

This would break someone's else diff/PR as it is not mirrored in header-only-file.

sleeptightAnsiC (changes requested):
Tested under MSYS2 UCRT64 and VS2022, and it indeed fixes this issue.

**If you could only remove that one little space, it should be good to merge.**

Just for reference, my environment is:
```
$ uname -a
MINGW64_NT-10.0-22631 QW11090824 3.6.4-23a25d49.x86_64 2025-07-16 16:15 UTC x86_64 Msys
$ gcc --version
gcc.exe (Rev8, Built by MSYS2 project) 15.2.0
$ clang-cl --version
clang version 21.1.0 (https://github.com/msys2/MSYS2-packages dd887afdaff2641f5919b53ef1286f484a8799a7)
```
```
** Visual Studio 2022 Developer Command Prompt v17.10.5
> cl /?
Microsoft (R) C/C++ Optimizing Compiler Version 19.40.33813 for x86
> clang-cl --version
clang version 17.0.3
```

src/nuklear.h
RobLoach:
``` suggestion
#ifndef NK_SIZE_TYPE
```

src/nuklear.h
```diff
@@ -152,11 +152,15 @@ extern "C" {
       #define NK_UINT32 unsigned int
     #endif
   #endif
-  #ifndef NK_SIZE_TYPE
+#ifndef NK_SIZE_TYPE
```
RobLoach:
``` suggestion
  #ifndef NK_SIZE_TYPE
```

PR #818, opened 2025-06-20
Fix typo in nuklear.h for nk_layout_space_push

realDLPS:
Fixes typo of bounds parameter description for the function nk_layout_space_push from "laoyut" to "layout"


sleeptightAnsiC:
This also needs to be changed in `src/nuklear.h`

Generally, you should never touch the `./nuklear.h` itself, but rather apply the changes to src/* files and then rebuild that header as described in...
https://github.com/Immediate-Mode-UI/Nuklear/blob/9afb3dd6d14fba0c5e62f08809b494b6d580b9b1/src/Readme.md?plain=1#L3-L4

I guess, for simple typo fixes there is no need for updating changelog, so that one should be fine.

RobLoach:
Small change, it'll need to be in the source. Feel free to open up a new PR, or update as needed. Simple typos like this sometimes may not even be worth the changeset, so :shrug:

PR #825, opened 2025-08-09
Add sdl3 renderer demo

suboyar:
I have ported `sdl_renderer` to SDL3.

**Changes from `sdl_renderer` to `sdl3_renderer`:**
- Update function names and types for SDL3 API changes
- Remove deprecated flags (e.g. `SDL_WINDOW_SHOWN`) 
- Use the new `SDL_PropertiesID` system
- Remove legacy Metal cliprect workaround (no longer needed)
- Switch vertex colors from byte values (0-255) to floating point (0-1)

### DIFF
--- a/demo/sdl3_renderer/main.c
+++ b/demo/sdl3_renderer/main.c
@@ -257,7 +257,9 @@ main(void)
 cleanup:
     nk_sdl_shutdown();
     SDL_DestroyRenderer(renderer);
+    SDL_DestroyProperties(renderer_props);
     SDL_DestroyWindow(win);
+    SDL_DestroyProperties(win_props);
     SDL_Quit();
     return 0;
 }


sleeptightAnsiC:
There is already an advanced PR for SDL3 demo which uses SDL_renderer, see https://github.com/Immediate-Mode-UI/Nuklear/pull/779
Does this PR have any advantages over the other one? or is it just a duplicate?

suboyar:
I skimmed through #779, nuklear_sdl3_renderer.h seem quite similar, while their nuklear_sdl3_renderer.c (compared to my main.c) is missing some initialization property and uses SDL's own main callback. While I tried to make a 1:1 implementation of the provided sdl2 demo.

As for why I made this PR, I searched for "SLD3 renderer demo" which didn't yield any results.

sleeptightAnsiC:
> their nuklear_sdl3_renderer.c (compared to my main.c) is missing some initialization property

Is there anything important/critical in those properties? If not, this isn't really a problem since *.c files are more of an example. Having less code in them would be even preferable. 

> [their nuklear_sdl3_renderer.c] uses SDL's own main callback

It is now preferred to use those callbacks in SDL3, see https://wiki.libsdl.org/SDL3/README-main-functions

> I tried to make a 1:1 implementation of the provided sdl2 demo.

Actually, the other PR is missing `nk_sdl_handle_grab` which I mentioned here https://github.com/Immediate-Mode-UI/Nuklear/pull/779#discussion_r2331174857 It also does not have Makefile (uses Cmake for some reason) and doesn't follow naming conventions for it's folders/files (compared to other demos). 

> As for why I made this PR, I searched for "SLD3 renderer demo" which didn't yield any results.

Yea, I also had this issue. I started working with your PR, then realized that https://github.com/Immediate-Mode-UI/Nuklear/pull/779 exists. IMO, the other PR has really misleading name...

@suboyar Anyway, I'm currently working on https://github.com/Immediate-Mode-UI/Nuklear/pull/779 since there are more people involved there, but would love to know if the other PR misses anything that your PR solves better.

demo/sdl3_renderer/nuklear_sdl3_renderer.h
```diff
@@ -0,0 +1,353 @@
...
+        static const struct nk_draw_vertex_layout_element vertex_layout[] = {
+            {NK_VERTEX_POSITION, NK_FORMAT_FLOAT, NK_OFFSETOF(struct nk_sdl_vertex, position)},
+            {NK_VERTEX_TEXCOORD, NK_FORMAT_FLOAT, NK_OFFSETOF(struct nk_sdl_vertex, uv)},
+            {NK_VERTEX_COLOR, NK_FORMAT_R32G32B32A32_FLOAT, NK_OFFSETOF(struct nk_sdl_vertex, col)},
+            {NK_VERTEX_LAYOUT_END}
+        };
+
+        Uint64 now = SDL_GetTicks();
+        sdl.ctx.delta_time_seconds = (float)(now - sdl.time_of_last_frame) / 1000;
+        sdl.time_of_last_frame = now;
```
sleeptightAnsiC, line 106: Is `sdl.ctx.delta_time_seconds` being used anywhere? I think that https://github.com/Immediate-Mode-UI/Nuklear/pull/779 is missing this one. I was also wondering, if we could use delta time for proper double-click (mentioned [here](https://github.com/Immediate-Mode-UI/Nuklear/pull/779/files#r2328481792))

EDIT: ahh, now I understand, Nuklear needs this for hiding scroll bar!
```sh
/tmp/Nuklear/src $ grep -rn 'delta_time_seconds' .
./nuklear_panel.c:493:            window->scrollbar_hiding_timer += ctx->delta_time_seconds;
./nuklear.h:5737:    float delta_time_seconds;
```

sleeptightAnsiC:
Thanks for this PR. I've taken some notes from here and applied them to https://github.com/Immediate-Mode-UI/Nuklear/pull/852

Since the other PR has been already merged, we can close _this_ one.

suboyar:
Closing this in favor of #853

PR #834, opened 2025-10-03
add: support sdl3 + opengl 3.3 and opengles 3.0 

vsaint1:
- Support for both desktop (OpenGL 3.3) and mobile/web (OpenGL ES 3.0)
- Example main menu with settings (resolution, difficulty, volume, etc..)

Here are some examples tested on different platforms:

ui_web

ui_android

image


vsaint1:
Forgot to add

image

demo/sdl3_opengl3_opengles3/nuklear_sdl3_ogl3.h
```diff
@@ -0,0 +1,521 @@
...
+                nk_input_motion(ctx, x + evt->motion.xrel, y + evt->motion.yrel);
+            }
+            else nk_input_motion(ctx, evt->motion.x, evt->motion.y);
+            return 1;
+
+        case SDL_EVENT_TEXT_INPUT:
+            {
+
+                nk_glyph glyph;
+                memcpy(glyph, evt->text.text, NK_UTF_SIZE);
```
oreo639: This doesn't work. The type of `evt->text.text` changed between SDL2 and SDL3. With SDL2, this is an array of size `SDL_TEXTINPUTEVENT_TEXT_SIZE`, so while it isn't correct, it wouldn't crash. With SDL3, this is now a pointer to a UTF-8 string, so this will now crash reading OOB.

The OOB issue can be fixed by using `strncpy(glyph, evt->text.text, NK_UTF_SIZE)` instead of `memcpy()`, but that doesn't fix the issue that `evt->text.text` can have more than one character.

For the latter issue, smth like the following should work afaict:
```
nk_rune unicode;
int glyph_len;
int byte_len;
const char *text = evt->text.text;
glyph_len = byte_len = nk_utf_decode(text, &unicode, 4);
while (unicode != '\0' && glyph_len) {
	nk_input_unicode(ctx, unicode);
 	glyph_len = nk_utf_decode(text+byte_len, &unicode, 4);
    byte_len += glyph_len;
}
```
(This will be able to store up to `NK_INPUT_MAX` from the SDL3 text input buffer)

vsaint1: this is correct, didn't crashed but i need to make some minor changes, thanks for showing this!!!

vsaint1:
Fixed the event text and runtime linking SDL3 example
image

sleeptightAnsiC:
@vsaint1 Since https://github.com/Immediate-Mode-UI/Nuklear/pull/852 has been merged, you can probably take as much stuff from [demo/sdl3_renderer](https://github.com/Immediate-Mode-UI/Nuklear/tree/master/demo/sdl3_renderer) as possible, and simply swap the rendering backend to the openGL ones. That demo has been extremely tested by multiple people and on multiple platforms. It should solve most of the problems you encounter here.

PR #837, opened 2025-10-10
Correctly handle hover in popup windows

lockie:
Hello! Right now `nk_widget_is_hovered` and its mouse friends `nk_widget_mouse_clicked` and `nk_widget_has_mouse_click_down` do not work for popup windows, always returning 0. This PR fixes that by making exception to always look inside the windows with `NK_PANEL_SET_POPUP` layout type.

### DIFF
--- a/src/nuklear_widget.c
+++ b/src/nuklear_widget.c
@@ -72,7 +72,8 @@ nk_widget_is_hovered(const struct nk_context *ctx)
     struct nk_rect bounds;
     NK_ASSERT(ctx);
     NK_ASSERT(ctx->current);
-    if (!ctx || !ctx->current || ctx->active != ctx->current)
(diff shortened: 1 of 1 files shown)


RobLoach (approved):
Sounds good to me. Thanks a lot for pushing this forwards.

RobLoach (changes requested):
Actually, we'll have to make this change in the src file before it makes its way into nuklear.h. Mind fixing it there as well?

RafaDEV99 (approved):
Yep, this is very Good! However, as noted by @RobLoach, it first needs to be pushed to the src file (i.e., `src/nuklear.c`).

lockie:
Oh yes, that completely slipped my mind, sorry. I've pushed the commit to the branch.

PR #845, opened 2025-10-28
Fix the comment lines in section headers

MrOneTwo:
Comment lines for headers were inconsistent and didn't play well with doxygen.

This is how the https://immediate-mode-ui.github.io/Nuklear/Window.html looks now. It includes the source code comment headers.

image

This patch hopes to fix this.


MrOneTwo:
Apparently doxygen messes up on abbreviations like "i.e.". Here you can see the issue:

image

I have pushed a commit that fixes it `i.e. ` -> `i.e.\ `. It looks clunky and isn't self explanatory why it's there, but it helps 🤷

MrOneTwo:
I have also improved this mess

before:

image

after:

image

MrOneTwo:
@RobLoach hmmm I don't see these issues in `src`. The `nuklear_draw.c` has the correct heading compared to `src/nuklear.h`

sleeptightAnsiC:
@MrOneTwo I believe that Rob had `./nuklear.h` in mind, the one at root directory, not `./src/*`

(people usually do the opposite of what you have done when submitting PRs here, they apply all changes to single-header-file but not to the sources, so that's probably where the confusion comes from, hah)

You just need to run [./src/paq.sh](https://github.com/Immediate-Mode-UI/Nuklear/blob/master/src/paq.sh) or [./src/paq.bat](https://github.com/Immediate-Mode-UI/Nuklear/blob/master/src/paq.bat) to fix it :)

RobLoach:
Ah, this is just in src/nuklear.h. Since this is just comments, I think it's fine to avoid re-building it for now. The change will be brought in next time it's built.

Going to squash merge for now.

sleeptightAnsiC:
> I think it's fine to avoid re-building it for now. The change will be brought in next time it's built.

This is not okey. Other PRs may have some problems because of this, like merge conflicts, confusing diffs, and so on... I think we can prevent similar situations with CI

RobLoach:
Rebuilt manually. CI rebuild would be nice.

MrOneTwo:
Cool, thanks for merging :+1:

sleeptightAnsiC:
> > > I think it's fine to avoid re-building it for now. The change will be brought in next time it's built.
> > >
> > This is not okey. Other PRs may have some problems because of this, like merge conflicts, confusing diffs, and so on... I think we can prevent similar situations with CI
> >
> Rebuilt manually. CI rebuild would be nice.

I meant CI for detecting a mismatch, but I guess auto-rebuilding would also be nice.
(though, I'm not sure how to implement the latter one)

Issue #847, opened 2025-10-29
Some warning in Win10 for MSYS2 mingw64 about warning: ISO C90 does not support 'long long' [-Wlong-long]


732857315:
When I make /demo/sdl_opengl3:

` make
rm -f bin/demo.exe
cc main.c -std=c89 -Wall -Wextra -pedantic -DSDL_DISABLE_IMMINTRIN_H -o bin/demo.exe -lmingw32 -lSDL2main -lSDL2 -lopengl32 -lm -lGLU32 -lGLEW32
../../nuklear.h: In function 'nk_memset':
../../nuklear.h:6004:39: warning: ISO C90 does not support 'long long' [-Wlong-long]
 6004 | # define NK_PTR_TO_UINT(x) ((nk_size)(__PTRDIFF_TYPE__)(x))
      |                                       ^~~~~~~~~~~~~~~~
../../nuklear.h:6884:14: note: in expansion of macro 'NK_PTR_TO_UINT'
 6884 |     if ((t = NK_PTR_TO_UINT(dst) & nk_wmask) != 0) {
      |              ^~~~~~~~~~~~~~
../../nuklear.h: In function 'nk_buffer_align':
../../nuklear.h:6003:37: warning: ISO C90 does not support 'long long' [-Wlong-long]
 6003 | # define NK_UINT_TO_PTR(x) ((void*)(__PTRDIFF_TYPE__)(x))
      |                                     ^~~~~~~~~~~~~~~~
../../nuklear.h:6017:6: note: in expansion of macro 'NK_UINT_TO_PTR'
 6017 |     (NK_UINT_TO_PTR((NK_PTR_TO_UINT((nk_byte*)(x) + (mask-1)) & ~(mask-1))))
      |      ^~~~~~~~~~~~~~
../../nuklear.h:8557:22: note: in expansion of macro 'NK_ALIGN_PTR'
 8557 |             memory = NK_ALIGN_PTR(unaligned, align);
      |                      ^~~~~~~~~~~~
In file included from main.c:25:
../../nuklear.h:6004:39: warning: ISO C90 does not support 'long long' [-Wlong-long]
 6004 | # define NK_PTR_TO_UINT(x) ((nk_size)(__PTRDIFF_TYPE__)(x))
      |                                       ^~~~~~~~~~~~~~~~
../../nuklear.h:6003:55: note: in definition of macro 'NK_UINT_TO_PTR'
[...]

Issue #256, opened 2021-01-31
Failure to build


revolucas:
Because of the changes made to **nuklear_internal.h** and **src/nuklear.h** in this commit: https://github.com/Immediate-Mode-UI/Nuklear/commit/455b404d866820a5d9ad6efa4c1aa4877fb45ea1
I can't seem to build it into a shared dll without errors:

unresolved external symbol NK_COS
unresolved external symbol NK_DTOA
unresolved external symbol NK_MEMCPY
unresolved external symbol NK_MEMSET
unresolved external symbol NK_SIN

Vurtun's original repo, I have no issues. Seems like  quite a few changes, in this repo, made to the _single header version_ weren't mirrored into the _separated implementation version_.

walkthetalk:
same problem, now only implement mode can be used.

RobLoach:
Is this still an issue?

sleeptightAnsiC:
Yes, this is still an issue.

If you try to build per each translation unit (instead of using amalgamated header), it will fail because multiple sources depend on macro definitions from `src/nuklear_math.c` and `src/nuklear_util.c`.

Not sure, if this is the same exact problem as the OP had, but it is probably also affected by the same commit.

<details><summary>repro</summary>
<p>


```sh
$ uname -a
Linux MAL200424 6.14.9-arch1-1 #1 SMP PREEMPT_DYNAMIC Thu, 29 May 2025 21:42:15 +0000 x86_64 GNU/Linux

$ cc --version
cc (GCC) 15.1.1 20250425

$ git clone https://github.com/Immediate-Mode-UI/Nuklear.git --recursive

$ cd Nuklear/src

$ for i in $(ls ./*.c); do echo "$i"; cc -x c -c "$i" -o "$i.o" -DNK_INCLUDE_FIXED_TYPES -DNK_INCLUDE_DEFAULT_ALLOCATOR -DNK_INCLUDE_STANDARD_IO -DNK_INCLUDE_STANDARD_VARARGS -DNK_INCLUDE_STANDARD_BOOL -DNK_INCLUDE_VERTEX_BUFFER_OUTPUT -DNK_INCLUDE_DEFAULT_FONT -DNK_INCLUDE_COMMAND_USERDATA ; done
## see results below
```

</p>
</details> 

<details><summary>result</summary>
<p>

```sh
./nuklear_9slice.c
./nuklear_buffer.c
[...]

Issue #851, opened 2025-11-04
nk_do_property() uses nk_strtod() instead of NK_STRTOD()


sleeptightAnsiC:
You can provide your own definition of `NK_STRTOD()`, but `nk_do_property()` will ignore this and still use internal `nk_strtod()` instead of user-provided macro:

https://github.com/Immediate-Mode-UI/Nuklear/blob/fcd64f85e54b9d35eee13347748d70fa4d2e134e/src/nuklear.h#L3818-L3821

https://github.com/Immediate-Mode-UI/Nuklear/blob/fcd64f85e54b9d35eee13347748d70fa4d2e134e/src/nuklear_property.c#L283-L287

Should be an easy fix, unless there are more issues like this one, or if there is some hidden dependency issue that I missed.

RobLoach:
Thahnks for putting together these follow up issues.

PavelSharp:
This indeed resolves the inconsistency that currently exists in Nuklear. 
In the code, the `NK_PROPERTY_FLOAT` branch calls `nk_strtof`:
https://github.com/Immediate-Mode-UI/Nuklear/blob/fcd64f85e54b9d35eee13347748d70fa4d2e134e/src/nuklear_property.c#L278-L282

`nk_strtof` implementation is:
https://github.com/Immediate-Mode-UI/Nuklear/blob/fcd64f85e54b9d35eee13347748d70fa4d2e134e/src/nuklear_util.c#L224-L232

Therefore, in the `NK_PROPERTY_DOUBLE` branch we should definitely call `NK_STRTOD`. This will ensure consistent behavior between the `float` and `double` property handling.

sleeptightAnsiC:
Note: be careful when fixing _this_ or https://github.com/Immediate-Mode-UI/Nuklear/issues/256: `nk_strtod()` is missing a `#ifndef NK_STRTOD` guard.

sleeptightAnsiC:
Note: make sure to also deal with this at the same time: https://github.com/Immediate-Mode-UI/Nuklear/issues/700

PR #852, opened 2025-11-06
Add sdl3_renderer demo V4

sleeptightAnsiC:
This is yet another PR that implements SDL_renderer demo for libSDL3. I consider this to be a Version 4, because there are 3 other PRs that attempted to do this. This one fixes a lot of long occuring issues and brings some improvements specific to SDL3 API. I've put a lot of time and serious care into this, and I hope it will be useful for someone out there.

PR is split into two logical commits.

EDIT: I collapsed the notes because they were scary to look at, and probably outdated at this point. I doubt anyone fully reads them anyway.

<details><summary>notes [CLICK ME]</summary>
<p>

NOTEs:

- I've started with the code from https://github.com/Immediate-Mode-UI/Nuklear/pull/779 and adjusted it accordingly...
- ...but the full origin is something like: `Nuklear/demo/sdl_renderer` -> https://github.com/Immediate-Mode-UI/Nuklear/pull/772 -> https://github.com/Immediate-Mode-UI/Nuklear/pull/779 + https://github.com/Immediate-Mode-UI/Nuklear/pull/825 -> _this_ PR...
- ...and I can probably put [Co-authored-by](https://docs.github.com/en/pull-requests/committing-changes-to-your-project/creating-and-editing-commits/creating-a-commit-with-multiple-authors#creating-co-authored-commits-on-github) into commit message (just let me let me know who was involved)
- Fixed a ton of issues mentioned by other PRs and reviews. (Sorry... I've stopped tracking those long ago)
- I did NOT bring back `nk_sdl_handle_grab()`, despite that I considered it, due to reasoning from https://github.com/Immediate-Mode-UI/Nuklear/pull/779#discussion_r1966566010
- I did NOT change how `userdata` is handled, despite some concerns, per https://github.com/Immediate-Mode-UI/Nuklear/pull/779/files#r2487912696
- Every `FIXME` in _this_ PR is an issue within Nuklear, that I couldn't solve from the demo side. Each one of those should be already reported and linked to Github. If you see any FIXME without explanation or link, let me know.
- Added "CONFIG" section to main.c which let's you drop any linkage dependency on libm/libc, and use SDL-provided symbols instead (this is currently disabled by default, read FIXME about it)
- Unified the use of malloc/free. Now you can pass one nk_allocator to this demo, and it will be used everywhere when possible. You can also use `nk_sdl_allocator()` to get Nuklear-compatible allocator provided by SDL. This is consistent with how Nuklear API works, but not exactly with the rest of demos.
- Added `nk_sdl_style_set_debug_font()` which let's you use primitive SDL debug font and allows to remove the dependency on font baking entirely (not used by default)
- Made sure that you can use this demo without amalgamated Nuklear header (although this is broken on main branch due to https://github.com/Immediate-Mode-UI/Nuklear/issues/256 )
- Changed CMakeLists.txt to Makefile (because every other demo uses GNU Make, not Cmake)
- I've tested this on: Ubuntu-25.04 with Wayland/Gnome, ArchLinux with Xorg, Windows11 with MSYS2 (although I had some issues due to https://github.com/Immediate-Mode-UI/Nuklear/issues/847#issuecomment-3497050976 ), and on macOS Ventura 13.6.9 - so this is well tested on desktop...
- ...however, I had no chance to test this on mobile platforms (Android/iOS) and on web (Emscripten) - any testing on those will be really helpful
- Suggestions and nitpicks are welcome. I will be force-pushing into this branch as fast as I can.

TODOs (out of the scope of _this_ PR):
- Figure out how to hook [SDL_SetTextInputArea](https://wiki.libsdl.org/SDL3/SDL_SetTextInputArea) into this demo. I haven't found any nk_* functions or any context member that would let me do this.
- I've noticed few issues when using `nk_sdl_style_set_debug_font()`, most of those are related to the Examples being hardcoded to specific font size, but there is also one severe bug: the background of color picker always appears black. Not sure if this is my fault or something in Nuklear.

</p>
</details>

### DIFF
--- a/demo/sdl3_renderer/Makefile
+++ b/demo/sdl3_renderer/Makefile
@@ -1,0 +1,41 @@
+# this Makefile is specific to GNU Make and GCC'compatible compilers
+
+PKG_CONFIG ?= $(shell command -v pkg-config)
+ifeq (,$(PKG_CONFIG))
+    $(error missing pkg-config utility!)
+endif
+
+PKG_SDL3 ?= sdl3
+ifeq (,$(shell $(PKG_CONFIG) --path $(PKG_SDL3)))
+    $(error $(PKG_CONFIG) could not find: $(PKG_SDL3))
+endif
+
+OS ?= $(shell uname -s)
+BINEXT-Windows_NT = .exe
+BINEXT ?= $(BINEXIT-$(OS))
+
+TEMPDIR ?= ./bin
+BIN ?= $(TEMPDIR)/demo$(BINEXT)
+
+CFLAGS += -std=c89 -Wall -Wextra -Wpedantic
+CFLAGS += -O2
+#CFLAGS += -O0 -g
+#CFLAGS += -fsanitize=address
+#CFLAGS += -fsanitize=undefined
+CFLAGS += $(shell $(PKG_CONFIG) $(PKG_SDL3) --cflags)
+
+LIBS += -lm
+LIBS += $(shell $(PKG_CONFIG) $(PKG_SDL3) --libs)
+
+DEP ?= $(BIN).d
+
+SRC = main.c
+
+$(BIN):
+	mkdir -p $(dir $@)
+	$(CC) $(SRC) -o $@ -MD -MF $(DEP) $(CFLAGS) $(LIBS)
+
+$(BIN): $(SRC) ./Makefile ./nuklear_sdl3_renderer.h ./../../nuklear.h
+
+-include $(DEP)
+
--- a/demo/sdl3_renderer/main.c
+++ b/demo/sdl3_renderer/main.c
@@ -1,0 +1,416 @@
+/* nuklear - public domain */
+#include <stdio.h>
+#include <stdlib.h>
+#include <stdint.h>
+#include <stdarg.h>
+#include <string.h>
+#include <math.h>
+#include <assert.h>
+#include <limits.h>
+#include <time.h>
+
+#include <SDL3/SDL.h>
+
+/* This demo uses "main callbacks" which are new in SDL3
+ * Those provide highly portable entry point and event loop for the app
+ * see: https://wiki.libsdl.org/SDL3/README-main-functions
+ * */
+#define SDL_MAIN_USE_CALLBACKS
+#include <SDL3/SDL_main.h>
+
+/* ===============================================================
+ *
+ *                          CONFIG
+ *
+ * ===============================================================*/
+
+/* optional: sdl3_renderer does not need any of these defines
+ * (but some examples might need them, so be careful) */
+#define NK_INCLUDE_STANDARD_VARARGS
+#define NK_INCLUDE_STANDARD_IO
+
+/* note that sdl3_renderer comes with nk_sdl_style_set_debug_font()
+ * so you may wish to use that instead of font baking */
+#define NK_INCLUDE_FONT_BAKING
+#define NK_INCLUDE_DEFAULT_FONT
+
+/* note that sdl3_renderer comes with nk_sdl_allocator()
+ * and you probably want to use that allocator instead of the default ones */
+/*#define NK_INCLUDE_DEFAULT_ALLOCATOR*/
+
+/* mandatory: sdl3_renderer depends on those defines */
+#define NK_INCLUDE_COMMAND_USERDATA
+#define NK_INCLUDE_VERTEX_BUFFER_OUTPUT
+
+
+/* We can re-use the types provided by SDL which are extremely portable,
+ * so there is no need for Nuklear to detect those on its own */
+/*#define NK_INCLUDE_FIXED_TYPES*/
+#ifndef NK_INCLUDE_FIXED_TYPES
+    #define NK_INT8              Sint8
+    #define NK_UINT8             Uint8
+    #define NK_INT16             Sint16
+    #define NK_UINT16            Uint16
+    #define NK_INT32             Sint32
+    #define NK_UINT32            Uint32
+    /* SDL guarantees 'uintptr_t' typedef */
+    #define NK_SIZE_TYPE         uintptr_t
+    #define NK_POINTER_TYPE      uintptr_t
+#endif
+
+/* FIXME: We could also use the `bool` symbol provided by SDL,
+ * but this is currently broken due to internal Nuklear issue, see:
+ * https://github.com/Immediate-Mode-UI/Nuklear/issues/849
+ * */
+#define NK_INCLUDE_STANDARD_BOOL
+#ifndef NK_INCLUDE_STANDARD_BOOL
+    #define NK_BOOL               bool
+#endif
+
+/* We can re-use various portable libc functions provided by SDL */
+#define NK_ASSERT(condition)      SDL_assert(condition)
+#define NK_STATIC_ASSERT(exp)     SDL_COMPILE_TIME_ASSERT(, exp)
+#define NK_MEMSET(dst, c, len)    SDL_memset(dst, c, len)
+#define NK_MEMCPY(dst, src, len)  SDL_memcpy(dst, src, len)
+#define NK_VSNPRINTF(s, n, f, a)  SDL_vsnprintf(s, n, f, a)
+#define NK_STRTOD(str, endptr)    SDL_strtod(str, endptr)
+
+/* sadly, SDL3 does not provide "dtoa" (only integer version) */
+/*#define NK_DTOA (str, d)*/
+
+/* SDL can also provide us with math functions, but beware that Nuklear's own
+ * implementation can be slightly faster at the cost of some precision */
+#define NK_INV_SQRT(f)            (1.0f / SDL_sqrtf(f))
+#define NK_SIN(f)                 SDL_sinf(f)
+#define NK_COS(f)                 SDL_cosf(f)
+
+/* HACK: Nuklear pulls two stb libraries in order to use font baking
+ * those libraries pull in some libc headers internally, creating a linkage dependency,
+ * so you’ll most likely want to use SDL symbols instead */
+#define STBTT_ifloor(x)       ((int)SDL_floor(x))
+#define STBTT_iceil(x)        ((int)SDL_ceil(x))
+#define STBTT_sqrt(x)         SDL_sqrt(x)
+#define STBTT_pow(x,y)        SDL_pow(x,y)
+#define STBTT_fmod(x,y)       SDL_fmod(x,y)
+#define STBTT_cos(x)          SDL_cosf(x)
+#define STBTT_acos(x)         SDL_acos(x)
+#define STBTT_fabs(x)         SDL_fabs(x)
+#define STBTT_assert(x)       SDL_assert(x)
+#define STBTT_strlen(x)       SDL_strlen(x)
+#define STBTT_memcpy          SDL_memcpy
+#define STBTT_memset          SDL_memset
+#define stbtt_uint8           Uint8
+#define stbtt_int8            Sint8
+#define stbtt_uint16          Uint16
+#define stbtt_int16           Sint16
+#define stbtt_uint32          Uint32
+#define stbtt_int32           Sint32
+#define STBRP_SORT            SDL_qsort
+#define STBRP_ASSERT          SDL_assert
+/* There is no need to define STBTT_malloc/STBTT_free macros
+ * Nuklear will define those to user-provided nk_allocator */
+
+
+#define NK_IMPLEMENTATION
+#include "../../nuklear.h"
+#define NK_SDL3_RENDERER_IMPLEMENTATION
+#include "nuklear_sdl3_renderer.h"
+
+#define WINDOW_WIDTH 1200
+#define WINDOW_HEIGHT 800
+
+/* ===============================================================
+ *
+ *                          EXAMPLE
+ *
+ * ===============================================================*/
+/* These are some code examples to provide a small overview of what can be
+ * done with this library. To try out an example uncomment the defines */
+/*#define INCLUDE_ALL */
+/*#define INCLUDE_STYLE */
+/*#define INCLUDE_CALCULATOR */
+/*#define INCLUDE_CANVAS */
+#define INCLUDE_OVERVIEW
+/*#define INCLUDE_CONFIGURATOR */
+/*#define INCLUDE_NODE_EDITOR */
+
+#ifdef INCLUDE_ALL
+    #define INCLUDE_STYLE
+    #define INCLUDE_CALCULATOR
+    #define INCLUDE_CANVAS
+    #define INCLUDE_OVERVIEW
+    #define INCLUDE_CONFIGURATOR
+    #define INCLUDE_NODE_EDITOR
+#endif
+
+#ifdef INCLUDE_STYLE
+    #include "../../demo/common/style.c"
+#endif
+#ifdef INCLUDE_CALCULATOR
+    #include "../../demo/common/calculator.c"
+#endif
+#ifdef INCLUDE_CANVAS
+    #include "../../demo/common/canvas.c"
+#endif
+#ifdef INCLUDE_OVERVIEW
+    #include "../../demo/common/overview.c"
+#endif
+#ifdef INCLUDE_CONFIGURATOR
+    #include "../../demo/common/style_configurator.c"
+#endif
+#ifdef INCLUDE_NODE_EDITOR
+    #include "../../demo/common/node_editor.c"
+#endif
+
+/* ===============================================================
+ *
+ *                          DEMO
+ *
+ * ===============================================================*/
+
+struct nk_sdl_app {
+    SDL_Window* window;
+    SDL_Renderer* renderer;
+    struct nk_context * ctx;
+    struct nk_colorf bg;
+    enum nk_anti_aliasing AA;
+};
+
+static SDL_AppResult
+nk_sdl_fail()
+{
+    SDL_LogError(SDL_LOG_CATEGORY_CUSTOM, "Error: %s", SDL_GetError());
+    return SDL_APP_FAILURE;
+}
+
+SDL_AppResult
+SDL_AppInit(void** appstate, int argc, char* argv[])
+{
+    struct nk_sdl_app* app;
+    struct nk_context* ctx;
+    float font_scale;
+    NK_UNUSED(argc);
+    NK_UNUSED(argv);
+
+    if (!SDL_Init(SDL_INIT_VIDEO | SDL_INIT_EVENTS)) {
+        return nk_sdl_fail();
+    }
+
+    app = SDL_malloc(sizeof(*app));
+    if (app == NULL) {
+        return nk_sdl_fail();
+    }
+
+    if (!SDL_CreateWindowAndRenderer("Nuklear: SDL3 Renderer", WINDOW_WIDTH, WINDOW_HEIGHT, SDL_WINDOW_RESIZABLE, &app->window, &app->renderer)) {
+        SDL_free(app);
+        return nk_sdl_fail();
+    }
+    *appstate = app;
+
+    if (!SDL_SetRenderVSync(app->renderer, 1)) {
+        SDL_LogError(SDL_LOG_CATEGORY_CUSTOM, "SDL_SetRenderVSync failed: %s", SDL_GetError());
+    }
+
+    app->bg.r = 0.10f;
+    app->bg.g = 0.18f;
+    app->bg.b = 0.24f;
+    app->bg.a = 1.0f;
+
+
+    font_scale = 1;
+    {
+        /* This scaling logic was kept simple for the demo purpose.
+         * On some platforms, this might not be the exact scale
+         * that you want to use. For more information, see:
+         * https://wiki.libsdl.org/SDL3/README-highdpi */
+        const float scale = SDL_GetWindowDisplayScale(app->window);
+        SDL_SetRenderScale(app->renderer, scale, scale);
+        font_scale = scale;
+    }
+
+    ctx = nk_sdl_init(app->window, app->renderer, nk_sdl_allocator());
+    app->ctx = ctx;
+
+#if 0
+    {
+        /* If you don't want to use advanced Nuklear font baking API
+         * you can use simple ASCII debug font provided by SDL
+         * just change the `#if 0` above to `#if 1` */
+        nk_sdl_style_set_debug_font(ctx);
+
+        /* Note that since debug font is extremely small (only 8x8 pixels),
+         * scaling it does not make much sense. The font would appear blurry. */
+        NK_UNUSED(font_scale);
+
+        /* You may wish to change a few style options, here are few recommendations: */
+        ctx->style.button.rounding = 0.0f;
+        ctx->style.menu_button.rounding = 0.0f;
+        ctx->style.property.rounding = 0.0f;
+        ctx->style.property.border = 0.0f;
+        ctx->style.option.border = -1.0f;
+        ctx->style.checkbox.border = -1.0f;
+        ctx->style.property.dec_button.border = -2.0f;
+        ctx->style.property.inc_button.border = -2.0f;
+        ctx->style.tab.tab_minimize_button.border = -2.0f;
+        ctx->style.tab.tab_maximize_button.border = -2.0f;
+        ctx->style.tab.node_minimize_button.border = -2.0f;
+        ctx->style.tab.node_maximize_button.border = -2.0f;
+        ctx->style.checkbox.spacing = 5.0f;
+
+        /* It's better to disable anti-aliasing when using small fonts */
+        app->AA = NK_ANTI_ALIASING_OFF;
+    }
+#else
+    {
+        struct nk_font_atlas *atlas;
+        struct nk_font_config config = nk_font_config(0);
+        struct nk_font *font;
+
+        /* set up the font atlas and add desired font; note that font sizes are
+         * multiplied by font_scale to produce better results at higher DPIs */
+        atlas = nk_sdl_font_stash_begin(ctx);
+        font = nk_font_atlas_add_default(atlas, 13 * font_scale, &config);
+        /*font = nk_font_atlas_add_from_file(atlas, "../../../extra_font/DroidSans.ttf", 14 * font_scale, &config);*/
+        /*font = nk_font_atlas_add_from_file(atlas, "../../../extra_font/Roboto-Regular.ttf", 16 * font_scale, &config);*/
+        /*font = nk_font_atlas_add_from_file(atlas, "../../../extra_font/kenvector_future_thin.ttf", 13 * font_scale, &config);*/
+        /*font = nk_font_atlas_add_from_file(atlas, "../../../extra_font/ProggyClean.ttf", 12 * font_scale, &config);*/
+        /*font = nk_font_atlas_add_from_file(atlas, "../../../extra_font/ProggyTiny.ttf", 10 * font_scale, &config);*/
+        /*font = nk_font_atlas_add_from_file(atlas, "../../../extra_font/Cousine-Regular.ttf", 13 * font_scale, &config);*/
+        nk_sdl_font_stash_end(ctx);
+
+        /* this hack makes the font appear to be scaled down to the desired
+         * size and is only necessary when font_scale > 1 */
+        font->handle.height /= font_scale;
+        /*nk_style_load_all_cursors(ctx, atlas->cursors);*/
+        nk_style_set_font(ctx, &font->handle);
+
+        app->AA = NK_ANTI_ALIASING_ON;
+    }
+#endif
+
+    return SDL_APP_CONTINUE;
+}
+
+SDL_AppResult
+SDL_AppEvent(void *appstate, SDL_Event* event)
+{
+    struct nk_sdl_app* app = (struct nk_sdl_app*)appstate;
+
+    switch (event->type) {
+        case SDL_EVENT_QUIT:
+            return SDL_APP_SUCCESS;
+        case SDL_EVENT_WINDOW_DISPLAY_SCALE_CHANGED:
+            /* You may wish to rescale the renderer and Nuklear during this event.
+             * Without this the UI and Font could appear too small or too big.
+             * This is not handled by the demo in order to keep it simple,
+             * but you may wish to re-bake the Font whenever this happens. */
+            SDL_Log("Unhandled scale event! Nuklear may appear blurry");
+            return SDL_APP_CONTINUE;
+    }
+
+    /* Remember to always rescale the event coordinates,
+     * if your renderer uses custom scale. */
+    SDL_ConvertEventToRenderCoordinates(app->renderer, event);
+
+    nk_sdl_handle_event(app->ctx, event);
+
+    return SDL_APP_CONTINUE;
+}
+
+SDL_AppResult
+SDL_AppIterate(void *appstate)
+{
+    struct nk_sdl_app* app = (struct nk_sdl_app*)appstate;
+    struct nk_context* ctx = app->ctx;
+
+#ifdef INCLUDE_CONFIGURATOR
+    static struct nk_color color_table[NK_COLOR_COUNT];
+    NK_MEMCPY(color_table, nk_default_color_style, sizeof(color_table));
+#endif
+
+    nk_input_end(ctx);
+
+    /* GUI */
+    if (nk_begin(ctx, "Demo", nk_rect(50, 50, 230, 250),
+        NK_WINDOW_BORDER|NK_WINDOW_MOVABLE|NK_WINDOW_SCALABLE|
+        NK_WINDOW_MINIMIZABLE|NK_WINDOW_TITLE))
+    {
+        enum {EASY, HARD};
+        static int op = EASY;
+        static int property = 20;
+
+        nk_layout_row_static(ctx, 30, 80, 1);
+        if (nk_button_label(ctx, "button")) {
+            SDL_Log("button pressed");
+        }
+        nk_layout_row_dynamic(ctx, 30, 2);
+        if (nk_option_label(ctx, "easy", op == EASY)) op = EASY;
+        if (nk_option_label(ctx, "hard", op == HARD)) op = HARD;
+        nk_layout_row_dynamic(ctx, 25, 1);
+        nk_property_int(ctx, "Compression:", 0, &property, 1000, 1, 1);
+
+        nk_layout_row_dynamic(ctx, 20, 1);
+        nk_label(ctx, "background:", NK_TEXT_LEFT);
+        nk_layout_row_dynamic(ctx, 25, 1);
+        if (nk_combo_begin_color(ctx, nk_rgb_cf(app->bg), nk_vec2(nk_widget_width(ctx),400))) {
+            nk_layout_row_dynamic(ctx, 120, 1);
+            app->bg = nk_color_picker(ctx, app->bg, NK_RGBA);
+            nk_layout_row_dynamic(ctx, 25, 1);
+            app->bg.r = nk_propertyf(ctx, "#R:", 0, app->bg.r, 1.0f, 0.01f,0.005f);
+            app->bg.g = nk_propertyf(ctx, "#G:", 0, app->bg.g, 1.0f, 0.01f,0.005f);
+            app->bg.b = nk_propertyf(ctx, "#B:", 0, app->bg.b, 1.0f, 0.01f,0.005f);
+            app->bg.a = nk_propertyf(ctx, "#A:", 0, app->bg.a, 1.0f, 0.01f,0.005f);
+            nk_combo_end(ctx);
+        }
+    }
+    nk_end(ctx);
+
+    /* -------------- EXAMPLES ---------------- */
+    #ifdef INCLUDE_CALCULATOR
+        calculator(ctx);
+    #endif
+    #ifdef INCLUDE_CANVAS
+        canvas(ctx);
+    #endif
+    #ifdef INCLUDE_OVERVIEW
+        overview(ctx);
+    #endif
+    #ifdef INCLUDE_CONFIGURATOR
+        style_configurator(ctx, color_table);
+    #endif
+    #ifdef INCLUDE_NODE_EDITOR
+        node_editor(ctx);
+    #endif
+    /* ----------------------------------------- */
+
+    SDL_SetRenderDrawColorFloat(app->renderer, app->bg.r, app->bg.g, app->bg.b, app->bg.a);
+    SDL_RenderClear(app->renderer);
+
+    nk_sdl_render(ctx, app->AA);
+    nk_sdl_update_TextInput(ctx);
+
+    /* show if TextInput is active for debug purpose. Feel free to remove this. */
+    SDL_SetRenderDrawColor(app->renderer, 0xFF, 0xFF, 0xFF, 0xFF);
+    SDL_RenderDebugTextFormat(app->renderer, 10, 10, "TextInputActive? %s",
+                              SDL_TextInputActive(app->window) ? "Yes" : "No");
+
+    SDL_RenderPresent(app->renderer);
+
+    nk_input_begin(ctx);
+    return SDL_APP_CONTINUE;
+}
+
+void
+SDL_AppQuit(void* appstate, SDL_AppResult result)
+{
+    struct nk_sdl_app* app = (struct nk_sdl_app*)appstate;
+    NK_UNUSED(result);
+
+    if (app) {
+        nk_sdl_shutdown(app->ctx);
+        SDL_DestroyRenderer(app->renderer);
+        SDL_DestroyWindow(app->window);
+        SDL_free(app);
+    }
+}
+
--- a/demo/sdl3_renderer/nuklear_sdl3_renderer.h
+++ b/demo/sdl3_renderer/nuklear_sdl3_renderer.h
@@ -1,0 +1,707 @@
+/* nuklear - public domain */
+
+/*
+ * ==============================================================
+ *
+ *                              API
+ *
+ * ===============================================================
+ */
+
+#ifndef NK_SDL3_RENDERER_H_
+#define NK_SDL3_RENDERER_H_
+
+#if SDL_MAJOR_VERSION < 3
+    #error "nk_sdl3_renderer requires at least SDL 3.0.0"
+#endif
+#ifndef NK_INCLUDE_COMMAND_USERDATA
+    #error "nk_sdl3_renderer requires the NK_INCLUDE_COMMAND_USERDATA define"
+#endif
+#ifndef NK_INCLUDE_VERTEX_BUFFER_OUTPUT
+    #error "nk_sdl3_renderer requires the NK_INCLUDE_VERTEX_BUFFER_OUTPUT define"
+#endif
+
+/* We have to redefine it because demos do not include any headers
+ * This is the same default value as the one from "src/nuklear_internal.h" */
+#ifndef NK_BUFFER_DEFAULT_INITIAL_SIZE
+    #define NK_BUFFER_DEFAULT_INITIAL_SIZE (4*1024)
+#endif
+
+NK_API struct nk_context*   nk_sdl_init(SDL_Window *win, SDL_Renderer *renderer, struct nk_allocator allocator);
+#ifdef NK_INCLUDE_FONT_BAKING
+NK_API struct nk_font_atlas* nk_sdl_font_stash_begin(struct nk_context* ctx);
+NK_API void                 nk_sdl_font_stash_end(struct nk_context* ctx);
+#endif
+NK_API int                  nk_sdl_handle_event(struct nk_context* ctx, SDL_Event *evt);
+NK_API void                 nk_sdl_render(struct nk_context* ctx, enum nk_anti_aliasing);
+NK_API void                 nk_sdl_update_TextInput(struct nk_context* ctx);
+NK_API void                 nk_sdl_shutdown(struct nk_context* ctx);
+NK_API nk_handle            nk_sdl_get_userdata(struct nk_context* ctx);
+NK_API void                 nk_sdl_set_userdata(struct nk_context* ctx, nk_handle userdata);
+NK_API void                 nk_sdl_style_set_debug_font(struct nk_context* ctx);
+NK_API struct nk_allocator  nk_sdl_allocator(void);
+
+#endif /* NK_SDL3_RENDERER_H_ */
+
+/*
+ * ==============================================================
+ *
+ *                          IMPLEMENTATION
+ *
+ * ===============================================================
+ */
+#ifdef NK_SDL3_RENDERER_IMPLEMENTATION
+#ifndef NK_SDL3_RENDERER_IMPLEMENTATION_ONCE
+#define NK_SDL3_RENDERER_IMPLEMENTATION_ONCE
+
+#ifndef NK_SDL_DOUBLE_CLICK_LO
+#define NK_SDL_DOUBLE_CLICK_LO 0.02
+#endif
+#ifndef NK_SDL_DOUBLE_CLICK_HI
+#define NK_SDL_DOUBLE_CLICK_HI 0.2
+#endif
+
+struct nk_sdl_device {
+    struct nk_buffer cmds;
+    struct nk_draw_null_texture tex_null;
+    SDL_Texture *font_tex;
+};
+
+struct nk_sdl_vertex {
+    float position[2];
+    float uv[2];
+    float col[4];
+};
+
+struct nk_sdl {
+    SDL_Window *win;
+    SDL_Renderer *renderer;
+    struct nk_user_font* debug_font;
+    struct nk_sdl_device ogl;
+    struct nk_context ctx;
+#ifdef NK_INCLUDE_FONT_BAKING
+    struct nk_font_atlas atlas;
+#endif
+    struct nk_allocator allocator;
+    nk_handle userdata;
+    Uint64 last_left_click;
+    Uint64 last_render;
+    bool insert_toggle;
+    bool edit_was_active;
+};
+
+NK_API nk_handle
+nk_sdl_get_userdata(struct nk_context* ctx) {
+    struct nk_sdl* sdl;
+    NK_ASSERT(ctx);
+    sdl = (struct nk_sdl*)ctx->userdata.ptr;
+    NK_ASSERT(sdl);
+    return sdl->userdata;
+}
+
+NK_API void
+nk_sdl_set_userdata(struct nk_context* ctx, nk_handle userdata) {
+    struct nk_sdl* sdl;
+    NK_ASSERT(ctx);
+    sdl = (struct nk_sdl*)ctx->userdata.ptr;
+    NK_ASSERT(sdl);
+    sdl->userdata = userdata;
+}
+
+NK_INTERN void *
+nk_sdl_alloc(nk_handle user, void *old, nk_size size)
+{
+    NK_UNUSED(user);
+    /* FIXME: nk_sdl_alloc should use SDL_realloc here, not SDL_malloc
+     * but this could cause a double-free due to bug within Nuklear, see:
+     * https://github.com/Immediate-Mode-UI/Nuklear/issues/768
+     * */
+#if 0
+    return SDL_realloc(old, size);
+#else
+    NK_UNUSED(old);
+    return SDL_malloc(size);
+#endif
+}
+
+NK_INTERN void
+nk_sdl_free(nk_handle user, void *old)
+{
+    NK_UNUSED(user);
+    SDL_free(old);
+}
+
+NK_API struct nk_allocator
+nk_sdl_allocator()
+{
+    struct nk_allocator allocator;
+    allocator.userdata.ptr = 0;
+    allocator.alloc = nk_sdl_alloc;
+    allocator.free = nk_sdl_free;
+    return allocator;
+}
+
+NK_INTERN void
+nk_sdl_device_upload_atlas(struct nk_context* ctx, const void *image, int width, int height)
+{
+    struct nk_sdl* sdl;
+    NK_ASSERT(ctx);
+    NK_ASSERT(image);
+    NK_ASSERT(width > 0);
+    NK_ASSERT(height > 0);
+
+    sdl = (struct nk_sdl*)ctx->userdata.ptr;
+    NK_ASSERT(sdl);
+
+    /* Clean up if the texture already exists. */
+    if (sdl->ogl.font_tex != NULL) {
+        SDL_DestroyTexture(sdl->ogl.font_tex);
+        sdl->ogl.font_tex = NULL;
+    }
+
+    sdl->ogl.font_tex = SDL_CreateTexture(sdl->renderer, SDL_PIXELFORMAT_ARGB8888, SDL_TEXTUREACCESS_STATIC, width, height);
+    NK_ASSERT(sdl->ogl.font_tex);
+    SDL_UpdateTexture(sdl->ogl.font_tex, NULL, image, 4 * width);
+    SDL_SetTextureBlendMode(sdl->ogl.font_tex, SDL_BLENDMODE_BLEND);
+}
+
+NK_API void
+nk_sdl_update_TextInput(struct nk_context* ctx)
+{
+    struct nk_sdl* sdl;
+    bool active;
+    NK_ASSERT(ctx);
+    sdl = (struct nk_sdl*)ctx->userdata.ptr;
+    NK_ASSERT(sdl);
+
+    /* Determine if Nuklear is using any top-level "edit" widget.
+     * Popups take higher priority because they block any incomming input.
+     * This will not work, if the widget is not updating context state properly. */
+    if (!ctx->active)
+        active = false;
+    else if (ctx->active->popup.win)
+        active = ctx->active->popup.win->edit.active;
+    else
+        active = ctx->active->edit.active;
+
+    /* decide, if TextInputActive should be unchanged/stoped/started
+     * and change its state accordingly for owned SDL Window */
+    if (active != sdl->edit_was_active)
+    {
+        const bool window_edit_active = SDL_TextInputActive(sdl->win);
+
+        /* If you ever hit this check, it means that the demo and your app
+         * (or something else) are all trying to manage TextInputActive state.
+         * This can cause subtle bugs where the state won't be what you expect.
+         * You can safely remove this assert and the demo will keep working,
+         * but make sure it does not cause any issues for you */
+        NK_ASSERT(window_edit_active == sdl->edit_was_active && "something else changed TextInputActive state for this Window");
+
+        if (!window_edit_active && !sdl->edit_was_active && active)
+            SDL_StartTextInput(sdl->win);
+        else if (window_edit_active && sdl->edit_was_active && !active)
+            SDL_StopTextInput(sdl->win);
+        sdl->edit_was_active = active;
+    }
+
+    /* FIXME:
+     * for full SDL3 integration, you also need to find current edit widget
+     * bounds and the text cursor offset, and pass this data into SDL_SetTextInputArea.
+     * This is currently not possible to do safely as Nuklear does not support it.
+     * https://wiki.libsdl.org/SDL3/SDL_SetTextInputArea
+     * https://github.com/Immediate-Mode-UI/Nuklear/pull/857
+     */
+}
+
+NK_API void
+nk_sdl_render(struct nk_context* ctx, enum nk_anti_aliasing AA)
+{
+    /* setup global state */
+    struct nk_sdl* sdl;
+    NK_ASSERT(ctx);
+    sdl = (struct nk_sdl*)ctx->userdata.ptr;
+    NK_ASSERT(sdl);
+
+    { /* setup internal delta time that Nuklear needs for animations */
+        const Uint64 ticks = SDL_GetTicks();
+        ctx->delta_time_seconds = (float)(ticks - sdl->last_render) / 1000.0f;
+        sdl->last_render = ticks;
+    }
+
+    {
+        SDL_Rect saved_clip;
+        bool clipping_enabled;
+        int vs = sizeof(struct nk_sdl_vertex);
+        size_t vp = NK_OFFSETOF(struct nk_sdl_vertex, position);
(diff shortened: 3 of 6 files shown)


demo/sdl3_renderer/Makefile
```diff
@@ -0,0 +1,41 @@
+# this Makefile is specific to GNU Make and GCC'compatible compilers
```
sleeptightAnsiC, line 1: I did rewrite this Makefile because I was using it for testing. I was about to revert it back to the old one from demo/sdl_renderer, but I think this is actually more portable and I kept it. It should work just fine on MSYS2, Cygwin, w64devkit, Linux, macOS, Termux, and literally everything that uses GNU Make as `make` command.

demo/sdl3_renderer/nuklear_sdl3_renderer.h
```diff
@@ -0,0 +1,663 @@
...
+    char *str;
+    size_t buflen;
+    int i;
+    struct nk_sdl* sdl = (struct nk_sdl*)usr.ptr;
+    NK_ASSERT(sdl);
+    if (len <= 0 || text == NULL) return;
+
+    /* 'text' is an arbitrary place in the buffer where the copy selection starts,
+     * 'len' is NOT the buffer lenght, but instead a count of UTF glyphs
+     * that the selection is trying to copy, we need to count 'buflen' ourselfs */
```
sleeptightAnsiC: Per https://github.com/Immediate-Mode-UI/Nuklear/pull/779/files#r1965047664 , I've tested this and the clipboard handling was broken when copy/pasting UTF glyphs. I fixed it, however, now I'm not so sure weather it was a bug in the demo or in Nuklear itself. This implementation (code below) **"just works"** with whatever is going on in the library right now.

This might be related to: https://github.com/Immediate-Mode-UI/Nuklear/pull/841

demo/sdl3_renderer/nuklear_sdl3_renderer.h
```diff
@@ -0,0 +1,663 @@
...
+
+    /* We only care about Window currently used by Nuklear */
+    if (sdl->win != SDL_GetWindowFromEvent(evt)) {
+        return 0;
+    }
+
+    /* SDL_ConvertEventToRenderCoordinates may mutate 'evt' so we use its copy from now on */
+    NK_MEMCPY(&evtcpy, evt, sizeof(evtcpy));
+
+    if (!SDL_ConvertEventToRenderCoordinates(sdl->renderer, &evtcpy)) {
```
sleeptightAnsiC: @vsaint1 I've added this per your recommendation https://github.com/Immediate-Mode-UI/Nuklear/pull/779#issuecomment-3364041662
and confirmed with SDL upstream that this function should never fail here https://github.com/libsdl-org/SDL/issues/14360
It doesn't really change anything on my side, but I kinda understand how it works (I think).
Would you mind sharing any code to test this further? or some steps?

sleeptightAnsiC:
Tagging anyone who was involved in this (sorry for a bit of spam)

@RobLoach @rswinkle @zoogies @vsaint1 @klei1984 @suboyar

demo/sdl3_renderer/nuklear_sdl3_renderer.h
vsaint1: - Add this line to test

```cpp
 SDL_SetRenderLogicalPresentation(_renderer, 320, 180, SDL_LOGICAL_PRESENTATION_STRETCH);
 ```

File: demo/common/overview.c [hunk 4]
```diff
@@ -601,6 +601,8 @@ overview(struct nk_context *ctx)
                 static int box_len;
                 nk_flags active;
 
+                nk_labelf(ctx, NK_TEXT_LEFT, "Any Input active: %s", nk_edit_is_any_active(ctx) ? "YES" : "no");
+
```
RobLoach: Was this for testing purposes? I think it's fine to keep this, though keeping the casing the same may be nicer to look at...
``` suggestion
                nk_labelf(ctx, NK_TEXT_LEFT, "Any Input active: %s", nk_edit_is_any_active(ctx) ? "Yes" : "No");

```

demo/sdl3_renderer/main.c
```diff
@@ -0,0 +1,408 @@
...
+/* mandatory: sdl3_renderer depends on those defines */
+#define NK_INCLUDE_COMMAND_USERDATA
+#define NK_INCLUDE_VERTEX_BUFFER_OUTPUT
+
+
+/* FIXME: Symbol configuration is currently disabled by default,
+ * due to symbol resolution issues in some Examples, see:
+ * https://github.com/Immediate-Mode-UI/Nuklear/issues/850
+ * */
+#if 0
```
RobLoach: This would be a follow up issue after this is merged?

demo/sdl3_renderer/main.c
```diff
@@ -0,0 +1,408 @@
...
+    #define stbtt_int8            Sint8
+    #define stbtt_uint16          Uint16
+    #define stbtt_int16           Sint16
+    #define stbtt_uint32          Uint32
+    #define stbtt_int32           Sint32
+    #define STBRP_SORT            SDL_qsort
+    #define STBRP_ASSERT          SDL_assert
+    /* There is no need to define STBTT_malloc/STBTT_free macros
+     * Nuklear will define those to user-provided nk_allocator */
+#endif
```
RobLoach: Whether or not we're font baking, I still think it's worth defining these if they haven't been defined already. Since SDL would assumingly  be available, it's worth assuming these functions will be there.

demo/sdl3_renderer/main.c
```diff
@@ -0,0 +1,408 @@
...
+        node_editor(ctx);
+    #endif
+    /* ----------------------------------------- */
+
+    SDL_SetRenderDrawColor(
+            app->renderer,
+            (Uint8)(app->bg.r * 255.0f),
+            (Uint8)(app->bg.g * 255.0f),
+            (Uint8)(app->bg.b * 255.0f),
+            SDL_ALPHA_OPAQUE);
```
RobLoach: Worth considering [SDL_SetRenderDrawColorFloat()](https://wiki.libsdl.org/SDL3/SDL_SetRenderDrawColorFloat) here?

demo/sdl3_renderer/Makefile
RobLoach: Seems fine with me 👍

demo/sdl3_renderer/nuklear_sdl3_renderer.h
PavelSharp: After reviewing almost all existing backends, it appears that when a backend implements a clipboard copy function, the `len` parameter is consistently **treated as** _the number of bytes_, **not** the _number of code points_. I believe this strongly indicates that the current behavior in Nuklear’s internal code is unintentional.
Given this, accepting PR #841 would fix clipboard behavior across all backends, and for consistency and interoperability, this backend should not be an exception. 
However, as a temporary solution, if we only want to test the behavior(**"just works"**), we can guard the change behind a preprocessor macro
<details>
<summary>Here is the list of tested backends and the results</summary>

- nuklear_d3d11.h(nk_d3d11_clipboard_copy)
- nuklear_d3d12.h(nk_d3d12_clipboard_copy)
- nuklear_d3d9.h(nk_d3d9_clipboard_copy) 
- nuklear_gdi.h(nk_gdi_clipboard_copy)
- gdi_native_nuklear/nuklear_gdi.h (nk_gdi_clipboard_copy)
- nuklear_gdip.h (nk_gdip_clipboard_copy)
- nuklear_glfw_gl2.h (nk_glfw3_clipboard_copy)
- nuklear_glfw_gl3.h (nk_glfw3_clipboard_copy)
- nuklear_glfw_gl4.h (nk_glfw3_clipboard_copy)
- nuklear_glfw_vulkan.h (nk_glfw3_clipboard_copy)
- nuklear_sdl_gl2.h (nk_sdl_clipboard_copy)
- nuklear_sdl_gl3.h (nk_sdl_clipboard_copy)
- nuklear_sdl_gles2.h (nk_sdl_clipboard_copy)
- nuklear_sdl_renderer.h (nk_sdl_clipboard_copy)
- nuklear_sdl_vulkan.h (nk_sdl_clipboard_copy)
- nuklear_sfml_gl2.h (nk_sfml_clipboard_copy NOT IMPL)
- nuklear_sfml_gl3.h (nk_sfml_clipboard_copy NOT IMPL)
- nuklear_xlib.h (nk_xlib_copy)
- nuklear_xlib_gl2.h (NOT FOUND)
- nuklear_xlib_gl3.h (NOT FOUND)
- nuklear_xcb.h (NOT IMPL)

</details>

RobLoach (commented):
Thanks again for the push one this.

demo/sdl3_renderer/nuklear_sdl3_renderer.h
sleeptightAnsiC: Thanks. Just tested with this and it works fine.

demo/sdl3_renderer/main.c
```diff
@@ -0,0 +1,408 @@
...
+
+/* We can re-use various portable libc functions provided by SDL */
+#define NK_ASSERT(condition)      SDL_assert(condition)
+#define NK_STATIC_ASSERT(exp)     SDL_COMPILE_TIME_ASSERT(, exp)
+#define NK_MEMSET(dst, c, len)    SDL_memset(dst, c, len)
+#define NK_MEMCPY(dst, src, len)  SDL_memcpy(dst, src, len)
+#define NK_VSNPRINTF(s, n, f, a)  SDL_vsnprintf(s, n, f, a)
+/* FIXME: Nuklear ignores NK_STRTOD macro, due to:
+ * https://github.com/Immediate-Mode-UI/Nuklear/issues/851
+ * */
```
sleeptightAnsiC: I think, I should remove this note. Defining `NK_STRTOD` does not break anything, it's simply being ignored silently. Having a note here is unnecessary and this is already being tracked by https://github.com/Immediate-Mode-UI/Nuklear/issues/851

demo/common/overview.c
sleeptightAnsiC: This is for testing, but I intentionally put it into PR.
People should be able to find example usage of something that is part of public API

Changed the casing by force pushing.

demo/sdl3_renderer/main.c
sleeptightAnsiC: Makes sense.

I removed the `#ifdef NK_INCLUDE_FONT_BAKING` guard and force pushed it.

demo/sdl3_renderer/main.c
sleeptightAnsiC: Removed this note note and force pushed.

demo/sdl3_renderer/main.c
sleeptightAnsiC: Yes. I just changed it to the *Float() one and forced pushed.

I didn't really touch anything related to rendering. It should be pretty much the same as it was back in https://github.com/Immediate-Mode-UI/Nuklear/pull/779 If you see anything else worth changing like this one, let me know.

demo/sdl3_renderer/main.c
sleeptightAnsiC: This should be fixed later after merging, as part of https://github.com/Immediate-Mode-UI/Nuklear/issues/850

However, we can hack around this problem, I can simply include few headers into this file and the Examples should build just fine, even after removing `#if 0` guard. After taking a look at other demos, [this is exactly what they're doing](https://github.com/Immediate-Mode-UI/Nuklear/blob/fcd64f85e54b9d35eee13347748d70fa4d2e134e/demo/sdl_renderer/main.c#L2-L10).

~~What do you think?~~

demo/sdl3_renderer/nuklear_sdl3_renderer.h
sleeptightAnsiC: What do you think @RobLoach ? Shall I bring back the old (broken) behavior for the sake of consistency with the rest of the demos? I can just put two implementations here and guard one with `#if 0`, just like Pavel mentioned above.

I'm not entirely sure if https://github.com/Immediate-Mode-UI/Nuklear/pull/841 is a correct solution to this problem. I just added a comment there explaining some thoughts: https://github.com/Immediate-Mode-UI/Nuklear/pull/841#issuecomment-3501578904

demo/sdl3_renderer/nuklear_sdl3_renderer.h
```diff
@@ -0,0 +1,663 @@
+/* nuklear - public domain */
```
sleeptightAnsiC, line 1: @RobLoach (just tagging you here because I believe you skipped this file during your last review)

EDIT: Rob answered this here: https://github.com/Immediate-Mode-UI/Nuklear/pull/852#issuecomment-3509060919

demo/sdl3_renderer/nuklear_sdl3_renderer.h
RobLoach: Let's roll with what's here for SDL3, and then work on fixing it afterwards? Consistency is important, but I don't want it to be a blocker.

demo/sdl3_renderer/nuklear_sdl3_renderer.h
```diff
@@ -0,0 +1,663 @@
...
+nk_sdl_clipboard_paste(nk_handle usr, struct nk_text_edit *edit)
+{
+    char *text;
+    NK_UNUSED(usr);
+
+    /* this function returns empty string on failure, not NULL */
+    text = SDL_GetClipboardText();
+    NK_ASSERT(text);
+    if (text[0] != '\0')
+        nk_textedit_paste(edit, text, SDL_utf8strlen(text));
```
PavelSharp: Please note that all backends based on DirectX/GLFW/SDL tend to pass the number of bytes to `nk_textedit_paste`. Because of that, the paste functionality in each of those backends is currently silently broken, and this is exactly what PR #841 addresses. Passing code point count (as done here) makes the behavior inconsistent and fragile for future changes

sleeptightAnsiC: This is the exact same problem as mentioned in https://github.com/Immediate-Mode-UI/Nuklear/pull/852#discussion_r2499130791 and I know about it.

No worries, I will remember about `nk_sdl_clipboard_paste` when editing `nk_sdl_clipboard_copy`

I think I will just put a 2nd implementation next to it and guard it with `#if 0`, so this will be trivial the change for people who use https://github.com/Immediate-Mode-UI/Nuklear/pull/841, or after it gets merged.

demo/sdl3_renderer/nuklear_sdl3_renderer.h
sleeptightAnsiC: I just added the 2nd implementation next to it and guarder them with preprocess. This is still still using the same behavior, but it will be trivial to change when https://github.com/Immediate-Mode-UI/Nuklear/pull/841 gets merged.

demo/sdl3_renderer/nuklear_sdl3_renderer.h
sleeptightAnsiC: I just added the 2nd implementation next to it and guarder them with preprocess. This is still using the same behavior, but it will be trivial to change when https://github.com/Immediate-Mode-UI/Nuklear/pull/841 gets merged.

demo/sdl3_renderer/main.c
sleeptightAnsiC: Since I was already force pushing into this PR, I did exactly what I said in my previous comment.

PavelSharp:
@sleeptightAnsiC 
At the moment, I have successfully built SDL3 for Android 13 and run a few standard examples. Building for Android wasn’t too difficult, so I might consider releasing some instruction as part of our Wiki. 
Next, I’m planning to integrate Nuklear, which means testing will include keyboard input and DPI correctness in addition to the usual checks. Maybe you have any other requests or suggestions?

PavelSharp:
:tada:Congratulations!
More than 10 years of development and now (aside from a bunch of small bugs) it actually works on Android!
Screenshot 1: running `sdl3_renderer` as-is
Screenshot 2: after adjusting the scale factor(`scale_x`, `scale_y`) and, it fully works at a basic level(navigating menus, interacting with widgets, UI behaves almost correctly)
What's even better it looks like most of the remaining issues(I guess) aren’t directly related to the `sdl3_renderer` itself, which is very good news. But to be sure, we still need to test it further.
[**UPD**] Added a new screenshot. No issues here(except for scaling problem), just making things look nice.
<details>
<summary>CLICK</summary>

изображение

Screenshot 3
изображение

</details>

sleeptightAnsiC:
@PavelSharp

> Maybe you have any other requests or suggestions?

Sorry, I missed your initial comment. I guess you already checked what I was wondering about, which are TextInput Events, virtual keyboard, and weather Touch events are automatically translated into Mouse events (I guess they are).

> it actually works on Android!

Thanks. This is quite surprising, because I never tested it outside of the desktop platforms. For full mobile support we need to figure out how to hook [SDL_SetTextInputArea](https://wiki.libsdl.org/SDL3/SDL_SetTextInputArea) into this demo. I think, I will just install Android Studio and implement this someday, but this is not required for this PR.

> Building for Android wasn’t too difficult, so I might consider releasing some instruction as part of our Wiki.

We can maybe add a small simple `Android.mk` just for this demo (?) I'm pretty sure that SDL3 has a similar file in it's source tree. Cmake could be better, but adding Cmake support is like opening a can of worms.

> What's even better it looks like most of the remaining issues(I guess) aren’t directly related to the sdl3_renderer itself, which is very good news. But to be sure, we still need to test it further.

Yea, I think most the problems would come from Nuklear and Examples. If you see something that we can patch in the demo, let me know and I will force push it.

PavelSharp:
@sleeptightAnsiC 

> What's even better it looks like most of the remaining issues(I guess) aren’t directly related to the sdl3_renderer

Honestly, my opinion is still a bit uncertain. As I said, I had to modify these lines:
```
scale_x = 2.5f * (float)(render_w) / (float)(window_w);
scale_y = 2.5f * (float)(render_h) / (float)(window_h);
```
So the main task right now is to understand why the resulting scaling was incorrect. We should lern more about [SDL3 and DPI](https://wiki.libsdl.org/SDL3/README-highdpi)

PavelSharp:
**News.** In the current version of this PR, keyboard input for all `nk_property_*` is broken. Pressing any physical key is ignored except for Backspace.
The likely cause is `nk_edit_is_any_active`, or more specifically, that we don’t account for other widgets (including user-defined ones) that may also request text input.
We probably need a more sophisticated input handling system, which would be better implemented as part of the backend rather than the nuklear core itself. But...
[**UPD**] 
Actually, I just got an idea. We could introduce a new callback into `nk_do_edit` that gets triggered when the widget reach the `NK_EDIT_ACTIVATED` (and/or `NK_EDIT_DEACTIVATED`) flag. As a nice bonus, this callback could also prepare the bounding rectangle, so we can call `SDL_SetTextInputArea` directly.  This would also allow user-defined widgets to trigger the same callback.

What do you think about this approach?

sleeptightAnsiC:
I'll take a look into it.

Callback would be perfect, but I don't want to break ABI/API accidentally. I suspect I simply missed some case for `nk_edit_is_any_active`, I already have some ideas.

When it comes to font scaling, this code is not mine, it comes from other demos and I didn't dare to touch it. I'll see what can be done here.

I also just realized that I did broke commit history. There should be two commits, not one... I probably smashed those during latest force-push.

PavelSharp:
@sleeptightAnsiC 

> When it comes to scale, this code is not mine, it comes from other demos and I didn't dare to touch it. I'll see what can be done here.

Currently, I am working on the High DPI support issue and have the draft solution. Patch will be ready soon

sleeptightAnsiC:
EDIT: already fixed. Hiding...

Converted to draft so this won't be merged by mistake in current form.
All reviews are still welcome, and this is pretty much done.
I just didn't have a time to address the latest concerns. I hope I will revisit this soon.

RobLoach:
Think it would be okay to merge, and fix the remaining in follow ups? Or should we wait?

sleeptightAnsiC:
EDIT: fixed in latest force-push. Hiding...

Right now, I should at least fix the _"keyboard input for all nk\_property\_* is broken"_ https://github.com/Immediate-Mode-UI/Nuklear/pull/852#issuecomment-3507261086, and _"there should be two commits, not one"_ https://github.com/Immediate-Mode-UI/Nuklear/pull/852#issuecomment-350845973, and we are pretty much done, unless something worse pops up. The rest can be patched later.

I don't consider the lack of [SDL_SetTextInputArea](https://wiki.libsdl.org/SDL3/SDL_SetTextInputArea) as a real issue. It's a very advanced feature that everyone will probably want to solve on its own. Only god really knows how this thing is supposed to work :weary: (https://github.com/libsdl-org/SDL/issues/13166).

The problem with DPI scaling that Pavel mentioned (https://github.com/Immediate-Mode-UI/Nuklear/pull/852#issuecomment-3506801190) is also not really an issue. This is something that is solvable by the application. It would be nice to have a good example of this working correctly, but not a must.

@RobLoach In the meantime, would you mind taking a look at `demo/sdl3_renderer/nuklear_sdl3_renderer.h` and other recent changes? I just want to be aware of any concerns ahead of time, so this PR won't take forever to merge.

RobLoach:
Looked pretty good to me. Wasn't able to run it since I'll need a CMake implementation, but we can fix any of the other small things up afterwards.

PavelSharp:
> I just want to be aware of any concerns ahead of time, so this PR won't take forever to merge.

Just to note. SDL3 has [dropped support for C89](https://github.com/libsdl-org/SDL/issues/6691) in favor of C99/11, so we should definitely keep this in mind.

sleeptightAnsiC:
> > I just want to be aware of any concerns ahead of time, so this PR won't take forever to merge.
> 
> Just to note. SDL3 has [dropped support for C89](https://github.com/libsdl-org/SDL/issues/6691) in favor of C99/11, so we should definitely keep this in mind.

That's not exactly true. Every public header should still be in highly portable C (ANSI/C89). They only changed it for internal sources. I don't believe this changes anything for us.

Every major C library uses C89 for public headers, so you can link it with more compilers, but internally they use whatever they want (usually C99/11 with GNU extensions).

(trust me bro, I have "Ansi C" in my name :smile: )

PavelSharp:
High DPI support patch is done and ready. 🎉🥳

✅ Key features:
- Runtime / hot DPI scaling based on SDL3 events (e.g. on Windows 10 changing scale to 125% in Settings will immediately trigger UI rebuild and font recalculation).
- Tested on Android 13 as well.
- Comments were added to clarify DPI-related logic and event flow.

I guess, with current Nuklear API there is essentially no better DPI handling possible, we adjust scale via `SDL_SetRenderScale` and rebake fonts on demand. `ProggyClean` won’t remain razor-sharp when scaled (expected limitation), but overall DPI behavior is consistent and correct. So, we have introduced High DPI support for this demo/renderer.

<details>
<summary>high_dpi.patch</summary>

```diff
diff --git a/old_main.c b/main.c
index edfd5a9..402f6ec 100644
--- a/old_main.c
+++ b/main.c
@@ -183,6 +183,96 @@ nk_sdl_fail()
     return SDL_APP_FAILURE;
 }
 
+static void
+sdl_demo_set_debug_font(void* appstate)
+{
+    struct nk_sdl_app* app = (struct nk_sdl_app*)appstate;
+    struct nk_context* ctx = app->ctx;
+    nk_sdl_style_set_debug_font(ctx);
+
+    /* note that since debug font is extremely small (only 8x8 pixels),
+     * you may wish to change a few style options, here are few recommendations: */
+    ctx->style.button.rounding = 0.0f;
+    ctx->style.menu_button.rounding = 0.0f;
+    ctx->style.property.rounding = 0.0f;
+    ctx->style.property.border = 0.0f;
+    ctx->style.option.border = -1.0f;
+    ctx->style.checkbox.border = -1.0f;
+    ctx->style.property.dec_button.border = -2.0f;
+    ctx->style.property.inc_button.border = -2.0f;
+    ctx->style.tab.tab_minimize_button.border = -2.0f;
+    ctx->style.tab.tab_maximize_button.border = -2.0f;
+    ctx->style.tab.node_minimize_button.border = -2.0f;
+    ctx->style.tab.node_maximize_button.border = -2.0f;
+    ctx->style.checkbox.spacing = 5.0f;
+
+    /* it's better to disable anti-aliasing when using blocky fonts */
+    app->AA = NK_ANTI_ALIASING_OFF;
+}
+
+static void
+sdl_demo_set_font(void* appstate, float font_scale)
+{
+    struct nk_sdl_app* app = (struct nk_sdl_app*)appstate;
+    struct nk_context* ctx = app->ctx;
+    struct nk_sdl* sdl = (struct nk_sdl*)ctx->userdata.ptr;
+
+    struct nk_font_atlas* atlas;
+    struct nk_font_config config = nk_font_config(0);
+    struct nk_font* font;
+    /* set up the font atlas and add desired font; note that font sizes are
+     * multiplied by font_scale to produce better results at higher DPIs */
+    /* FIXME: This looks odd, but all existing renderer no calls nk_font_atlas_clear
+     * inside nk_sdl_font_stash_begin. This might cause a leak if called multiple times */
+    atlas = nk_sdl_font_stash_begin(ctx);
+    font = nk_font_atlas_add_default(atlas, 13 * font_scale, &config);
+    /*font = nk_font_atlas_add_from_file(atlas, "../../../extra_font/DroidSans.ttf", 14 * font_scale, &config);*/
+    /*font = nk_font_atlas_add_from_file(atlas, "../../../extra_font/Roboto-Regular.ttf", 16 * font_scale, &config);*/
+    /*font = nk_font_atlas_add_from_file(atlas, "../../../extra_font/kenvector_future_thin.ttf", 13 * font_scale, &config);*/
+    /*font = nk_font_atlas_add_from_file(atlas, "../../../extra_font/ProggyClean.ttf", 12 * font_scale, &config);*/
+    /*font = nk_font_atlas_add_from_file(atlas, "../../../extra_font/ProggyTiny.ttf", 10 * font_scale, &config);*/
+    /*font = nk_font_atlas_add_from_file(atlas, "../../../extra_font/Cousine-Regular.ttf", 13 * font_scale, &config);*/
+    nk_sdl_font_stash_end(ctx);
+
+    /* this hack makes the font appear to be scaled down to the desired
+     * size and is only necessary when font_scale > 1 */
+    font->handle.height /= font_scale;
+    /*nk_style_load_all_cursors(ctx, atlas->cursors);*/
+    nk_style_set_font(ctx, &font->handle);
+
+    app->AA = NK_ANTI_ALIASING_ON;
+}
+
+static void
+sdl_demo_on_scale_changed(void* appstate, float scale)
+{
+    struct nk_sdl_app* app = (struct nk_sdl_app*)appstate;
+    struct nk_context* ctx = app->ctx;
+    struct nk_sdl* sdl;
+
+    NK_ASSERT(scale > 0);
+    SDL_SetRenderScale(app->renderer, scale, scale);
+    /* If you don't want to use advanced Nuklear font baking API
+     * you can use simple ASCII debug font provided by SDL
+     * just change the `#if 0` above to `#if 1` */
+#if 0
+    {
+        sdl = (struct nk_sdl*)ctx->userdata.ptr;
+        NK_ASSERT(sdl);
+        /* According to SDL3 documentation, debug font glyphs are always
+         * 8 pixels high, so no scaling is needed when the display DPI changes */
+        if (!sdl->debug_font)
+            sdl_demo_set_debug_font(appstate);
+    }
+#else
+    {
+        NK_UNUSED(sdl);
+        /*scale*2 will give even better quality for ProggyClean*/
+        sdl_demo_set_font(appstate, scale);
+    }
+#endif
+}
+
 SDL_AppResult
 SDL_AppInit(void** appstate, int argc, char* argv[])
 {
@@ -201,7 +291,7 @@ SDL_AppInit(void** appstate, int argc, char* argv[])
         return nk_sdl_fail();
     }
 
-    if (!SDL_CreateWindowAndRenderer("Nuklear: SDL3 Renderer", WINDOW_WIDTH, WINDOW_HEIGHT, SDL_WINDOW_RESIZABLE, &app->window, &app->renderer)) {
+    if (!SDL_CreateWindowAndRenderer("Nuklear: SDL3 Renderer", WINDOW_WIDTH, WINDOW_HEIGHT, SDL_WINDOW_RESIZABLE | SDL_WINDOW_HIGH_PIXEL_DENSITY, &app->window, &app->renderer)) {
         SDL_free(app);
         return nk_sdl_fail();
     }
@@ -216,77 +306,10 @@ SDL_AppInit(void** appstate, int argc, char* argv[])
     app->bg.b = 0.24f;
     app->bg.a = 1.0f;
 
-
-    font_scale = 1;
-    {
-        int render_w, render_h;
-        int window_w, window_h;
-        float scale_x, scale_y;
-        SDL_GetCurrentRenderOutputSize(app->renderer, &render_w, &render_h);
-        SDL_GetWindowSize(app->window, &window_w, &window_h);
-        scale_x = (float)(render_w) / (float)(window_w);
-        scale_y = (float)(render_h) / (float)(window_h);
-        SDL_SetRenderScale(app->renderer, scale_x, scale_y);
-        font_scale = scale_y;
-    }
-
     ctx = nk_sdl_init(app->window, app->renderer, nk_sdl_allocator());
     app->ctx = ctx;
 
-#if 0
-    {
-        /* If you don't want to use advanced Nuklear font baking API
-         * you can use simple ASCII debug font provided by SDL
-         * just change the `#if 0` above to `#if 1` */
-        nk_sdl_style_set_debug_font(ctx);
-        NK_UNUSED(font_scale);
-
-        /* note that since debug font is extremely small (only 8x8 pixels),
-         * you may wish to change a few style options, here are few recommendations: */
-        ctx->style.button.rounding = 0.0f;
-        ctx->style.menu_button.rounding = 0.0f;
-        ctx->style.property.rounding = 0.0f;
-        ctx->style.property.border = 0.0f;
-        ctx->style.option.border = -1.0f;
-        ctx->style.checkbox.border = -1.0f;
-        ctx->style.property.dec_button.border = -2.0f;
-        ctx->style.property.inc_button.border = -2.0f;
-        ctx->style.tab.tab_minimize_button.border = -2.0f;
-        ctx->style.tab.tab_maximize_button.border = -2.0f;
-        ctx->style.tab.node_minimize_button.border = -2.0f;
-        ctx->style.tab.node_maximize_button.border = -2.0f;
-        ctx->style.checkbox.spacing = 5.0f;
-
-        /* it's better to disable anti-aliasing when using blocky fonts */
-        app->AA = NK_ANTI_ALIASING_OFF;
-    }
-#else
-    {
-        struct nk_font_atlas *atlas;
-        struct nk_font_config config = nk_font_config(0);
-        struct nk_font *font;
-
-        /* set up the font atlas and add desired font; note that font sizes are
-         * multiplied by font_scale to produce better results at higher DPIs */
-        atlas = nk_sdl_font_stash_begin(ctx);
-        font = nk_font_atlas_add_default(atlas, 13 * font_scale, &config);
-        /*font = nk_font_atlas_add_from_file(atlas, "../../../extra_font/DroidSans.ttf", 14 * font_scale, &config);*/
-        /*font = nk_font_atlas_add_from_file(atlas, "../../../extra_font/Roboto-Regular.ttf", 16 * font_scale, &config);*/
-        /*font = nk_font_atlas_add_from_file(atlas, "../../../extra_font/kenvector_future_thin.ttf", 13 * font_scale, &config);*/
-        /*font = nk_font_atlas_add_from_file(atlas, "../../../extra_font/ProggyClean.ttf", 12 * font_scale, &config);*/
-        /*font = nk_font_atlas_add_from_file(atlas, "../../../extra_font/ProggyTiny.ttf", 10 * font_scale, &config);*/
-        /*font = nk_font_atlas_add_from_file(atlas, "../../../extra_font/Cousine-Regular.ttf", 13 * font_scale, &config);*/
-        nk_sdl_font_stash_end(ctx);
-
-        /* this hack makes the font appear to be scaled down to the desired
-         * size and is only necessary when font_scale > 1 */
-        font->handle.height /= font_scale;
-        /*nk_style_load_all_cursors(ctx, atlas->cursors);*/
-        nk_style_set_font(ctx, &font->handle);
-
-        app->AA = NK_ANTI_ALIASING_ON;
-    }
-#endif
+    sdl_demo_on_scale_changed(app, SDL_GetWindowDisplayScale(app->window));
 
     return SDL_APP_CONTINUE;
 }
@@ -297,6 +320,11 @@ SDL_AppEvent(void *appstate, SDL_Event* event)
     struct nk_sdl_app* app = (struct nk_sdl_app*)appstate;
 
     switch (event->type) {
+        case SDL_EVENT_WINDOW_DISPLAY_SCALE_CHANGED:
+            sdl_demo_on_scale_changed(app, SDL_GetWindowDisplayScale(app->window));
+            /* In this version, DPI is managed by this demo itself,
+             * so we avoid forwarding this event to Nuklear to prevent potential double handling */
+            return SDL_APP_CONTINUE;
         case SDL_EVENT_QUIT:
             return SDL_APP_SUCCESS;
     }
```

</details>

rswinkle:
I spent a little time today trying to figure out why properties weren't triggering the edit and this is the fix I came up with (not sure if setting the name really matters but I was trying to match code in nk_edit_buffer():

```diff
diff --git a/src/nuklear_property.c b/src/nuklear_property.c
index 68b47ba..a234628 100644
--- a/src/nuklear_property.c
+++ b/src/nuklear_property.c
@@ -405,6 +405,8 @@ nk_property(struct nk_context *ctx, const char *name, struct nk_property_variant
         win->property.name = hash;
         win->property.select_start = *select_begin;
         win->property.select_end = *select_end;
+        win->edit.active = nk_true;
+        win->edit.name = hash;
         if (*state == NK_PROPERTY_DRAG) {
             ctx->input.mouse.grab = nk_true;
             ctx->input.mouse.grabbed = nk_true;
@@ -420,6 +422,7 @@ nk_property(struct nk_context *ctx, const char *name, struct nk_property_variant
         win->property.select_start = 0;
         win->property.select_end = 0;
         win->property.active = 0;
+        win->edit.active = nk_false;
     }
 }
 NK_API void

```
Seems to be working correctly.

I also added a text field to the small demo window to test the differences and discovered something else.  Text input stays active for a text field even if you click on (for instance) a color drop down, even if you actually click around in the color box or drag the nested channel properties. But it goes inactive if it was the property that was being edited. Just another small thing to dig into in the combobox/dropdown code.

sleeptightAnsiC:
[replying to multiple messages at the same time]

@rswinkle So this is more of a bug inside of Nuklear, right? If I understand your [comment](https://github.com/Immediate-Mode-UI/Nuklear/pull/852#issuecomment-3518277340), the widgets are not setting their `edit` flags as they should, and this should be fixed "per-widget" (?) I was thinking about re-implementing `nk_edit_is_any_active`, but if the widget context is not properly set, I should probably take a look into this first. 

@PavelSharp I think, I fully understand your [patch](https://github.com/Immediate-Mode-UI/Nuklear/pull/852#issuecomment-3509287687), but I find the font rebaking to be a bit of an overkill. It increases the complexity of `main.c` and this file is meant to be replaced by the demo user anyway. Maybe it would be better to simply set the scale initially with `SDL_GetWindowDisplayScale`, just like you did, and leave the note (and maybe some log message) under `SDL_EVENT_WINDOW_DISPLAY_SCALE_CHANGED` that the font will break if you're not handling it yourself. This is such an edge case that I don't want to add more control flow only for it.

@RobLoach I can bring back the cmake file. I didn't like the old one from https://github.com/Immediate-Mode-UI/Nuklear/pull/779, because it was silently downloading SDL sources instead of reusing the system-wide copy of it. I guess, since SDL3 is now primary cmake-based, it makes sense to have it here.

Anyway, thank you guys for your input and patches. I hope I will deal with this soon. Sorry, for the slowdown on my part. I am a bit exhausted with debugging Nuklear and grepping its old commit messages

rswinkle:
> @rswinkle So this is more of a bug inside of Nuklear, right? If I understand your [comment](https://github.com/Immediate-Mode-UI/Nuklear/pull/852#issuecomment-3518277340), the widgets are not setting their `edit` flags as they should, and this should be fixed "per-widget" (?) I was thinking about re-implementing `nk_edit_is_any_active`, but if the widget context is not properly set, I should probably take a look into this first.

Yes, these are definitely bugs in Nuklear. They happen with my method of doing StartText/StopText too (see diff).

The issue is that nk_property has its own method for tracking its edit active/inactive state which is separate from the "global" context text_edit active state and wasn't updating the latter.  It calls a function nk_do_edit() but that is part of a nested series of functions used for text editing and the global state setting was in an outer call after nk_do_edit(). Similarly, I'm sure when I look at the combobox/dropdown code will have its own state and will be missing the lines to turn off an active edit. I'll do that later if someone else hasn't already by then.

```diff
diff --git a/src/external/nuklear/nuklear.h b/src/external/nuklear/nuklear.h
index 38b9b58..f046e06 100644
--- a/src/external/nuklear/nuklear.h
+++ b/src/external/nuklear/nuklear.h
@@ -3506,6 +3506,7 @@ enum nk_edit_events {
     NK_EDIT_DEACTIVATED = NK_FLAG(3), /**!< edit widget went from state active to state inactive */
     NK_EDIT_COMMITED    = NK_FLAG(4)  /**!< edit widget has received an enter and lost focus */
 };
+
 NK_API nk_flags nk_edit_string(struct nk_context*, nk_flags, char *buffer, int *len, int max, nk_plugin_filter);
 NK_API nk_flags nk_edit_string_zero_terminated(struct nk_context*, nk_flags, char *buffer, int max, nk_plugin_filter);
 NK_API nk_flags nk_edit_buffer(struct nk_context*, nk_flags, struct nk_text_edit*, nk_plugin_filter);
@@ -4698,9 +4699,19 @@ struct nk_keyboard {
     int text_len;
 };
 
+/*
+ * Not sure if we shouldn't just use nk_edit_events but that's for individual fields
+ * in a window and we need change over the whole GUI...might as well have unique enums
+ * for clarity
+enum nk_edit_text_change { NK_NO_TEXT_INPUT_CHANGE, NK_START_TEXT, NK_STOP_TEXT };
+do we even need this?
+*/
+
 struct nk_input {
     struct nk_keyboard keyboard;
     struct nk_mouse mouse;
+    nk_bool text_input_active_prev;
+    nk_bool text_input_active;
 };
 
 NK_API nk_bool nk_input_has_mouse_click(const struct nk_input*, enum nk_buttons);
diff --git a/src/external/nuklear/nuklear_input.c b/src/external/nuklear/nuklear_input.c
index 48ad586..a1ba04e 100644
--- a/src/external/nuklear/nuklear_input.c
+++ b/src/external/nuklear/nuklear_input.c
@@ -25,6 +25,9 @@ nk_input_begin(struct nk_context *ctx)
     in->mouse.delta.y = 0;
     for (i = 0; i < NK_KEY_MAX; i++)
         in->keyboard.keys[i].clicked = 0;
+
+    in->text_input_active_prev = in->text_input_active;
+    in->text_input_active = nk_false;
 }
 NK_API void
 nk_input_end(struct nk_context *ctx)
diff --git a/src/external/nuklear/nuklear_window.c b/src/external/nuklear/nuklear_window.c
index 8c286e4..6a12836 100644
--- a/src/external/nuklear/nuklear_window.c
+++ b/src/external/nuklear/nuklear_window.c
@@ -302,6 +302,8 @@ nk_end(struct nk_context *ctx)
     if (!ctx || !ctx->current)
         return;
 
+    ctx->input.text_input_active |= ctx->current->edit.active;
+
     layout = ctx->current->layout;
     if (!layout || (layout->type == NK_PANEL_WINDOW && (ctx->current->flags & NK_WINDOW_HIDDEN))) {
         ctx->current = 0;
```

PavelSharp:
@SleeptightAnsiC, maybe I’m missing something, but after looking deeper into this [issue](https://github.com/Immediate-Mode-UI/Nuklear/pull/852#issuecomment-3507261086), I’m convinced that the only truly safe and future-proof solution for handling text input and virtual keyboard behavior is to **add a callback**. Ideally behind a macro guard so it doesn’t break the ABI or API.

Here are my main arguments:

1. **Avoiding fragile internal changes**
Any other approach risks introducing major internal modifications to Nuklear’s core logic, something that could easily lead to regressions and new bugs down the line. Include some complex scenarios that we might not see when testing with the standard examples only
A callback, on the other hand, keeps the logic contained and easy to reason about.

2. **Currently, there are already many bugs and fixing them may lead to new bugs**
2.1 As noted by @rswinkle in his [comment](https://github.com/Immediate-Mode-UI/Nuklear/pull/852#issuecomment-3518277340)
2.2 As noted by me in #856 

4. **Respecting user-defined widgets**
A callback also gives us a clean and extensible way for custom widgets (not based on `nk_do_edit`) to request and release text input properly, something impossible to do safely under the current proposal.

5. **Safe and backward-compatible**
If implemented with a macro guard, this feature remains completely optional and non-breaking.
Existing applications would behave exactly the same, while backends or advanced users could enable the callback when needed.

6. **Contextual data**(mostly for `SDL_SetTextInputArea`)
The callback could provide the input area rectangle and other stuff, allowing the backend to know:
**when to show the virtual keyboard (`edit` becomes active)**
**and when to hide it (e.g. `edit` turns read-only in this frame or loses focus)**

7. Consistency with existing Nuklear design
This is just my opinion, but adding a virtual keyboard callback is not a hacky workaround, it’s actually a clean and natural extension of Nuklear’s design. It fits perfectly into the existing family of callbacks (copy, paste, etc.) and keeps the logic modular, explicit, and easy to reason about, exactly in the spirit of Nuklear’s minimalistic philosophy.

I’d really like to hear what you think about this, @rswinkle and @RobLoach.

If nobody minds, I’ll go ahead and start working on it soon❤️

sleeptightAnsiC:
> maybe I’m missing something, but after looking deeper into this issue, I’m convinced that the only truly safe and future-proof solution for handling text input and virtual keyboard behavior is to add a callback.

I don't see how adding a callback would solve any issue here.

It could be a nice addition, but the _callback_ needs to be _call'ed_ by _something_.
That _thing_ is a code that is currently broken. Inserting a callback would not fix it.
You need to fix the the underlying issue first, before doing anything else.

I'm not gonna introduce any callback as part of this PR and I won't change my mind about it.
You can do this as part of different PR.

PavelSharp:
@sleeptightAnsiC 

> I don't see how adding a callback would solve any issue here.

I'm sure it's obvious. So, my plan is to introduce a callback that would be invoked inside `nk_do_edit` depending on the flags `NK_EDIT_ACTIVATED` and `NK_EDIT_DEACTIVATED`. The callback would be a function pointer stored in `nk_context` and set by the backend, which is fully consistent with the style of existing callbacks like copy and paste.

It’s also important to note that `nk_do_edit` is only called by text fields (e.g. `nk_edit_string`) and `nk_property`. This means the callback would directly address the exact issue we are discussing, without affecting unrelated parts of Nuklear.

> I won't change my mind about it.

I believe I’ve given enough reasons why a new callback is preferable. Modifying `nk_edit_is_any_active` would require many changes in Nuklear, which almost equivalent to new bugs.

sleeptightAnsiC:
> So, my plan is to introduce a callback that would be invoked inside nk_do_edit 
> (...)
> It’s also important to note that nk_do_edit is only called by text fields (e.g. nk_edit_string) and nk_property. This means the callback would directly address the exact issue we are discussing, without affecting unrelated parts of Nuklear.

If that's so simple, then just move the edit state de/activation to this one function, and the problem will be "fixed". The state changes are already here, the callback would be a new thing. One does not contradict the other, but the old implementation should be fixed.

### DISCLAIMER: From now on I'm gonna ignore any text that has a sign of AI/ChatGPT use (sorry)

rswinkle:
> I’d really like to hear what you think about this, @rswinkle and @RobLoach.

I have to agree with @sleeptightAnsiC on this.

> 1. **Avoiding fragile internal changes**
>    Any other approach risks introducing major internal modifications to Nuklear’s core logic, something that could easily lead to regressions and new bugs down the line. Include some complex scenarios that we might not see when testing with the standard examples only
>    A callback, on the other hand, keeps the logic contained and easy to reason about.

I would not call either my approach or the current approach fragile. As sleeptightAnsiC said, all approaches at their optimal/minimal state amount to the same thing, and in fact the callback method probably requires the most in actual lines changed. The logic is not/would not be in the callback. It is a matter of Nuklear internal state regardless. Knowing where/when to insert the callback is the same knowledge you would need for any other method.

> 2. **Currently, there are already many bugs and fixing them may lead to new bugs**

Again, these are Nuklear bugs that need to be fixed in any case and the code changes are very small. It's a matter of discovery/precision.

> 3. **Respecting user-defined widgets**
>    A callback also gives us a clean and extensible way for custom widgets (not based on `nk_do_edit`) to request and release text input properly, something impossible to do safely under the current proposal.

The request and release does not happen at the widget level, that's part of the issue. It has to be aggregated over all the windows/widgets over the whole frame being rendered, and at that point, you know if either Start or Stop should be called based on whether it changed from the previous. Whether you use my method, adding a couple of members to try to simplify the test by collating as we go, or try to get the same result using a function like nk_is_any_active() combining hopefully all necessary checks (or both), the solution is the same.  Ideally all widgets using text input should use the same method directly or (since that's obviously not the case) they should pass that state along to the text_edit state as the "master" state.

> 4. **Safe and backward-compatible**
>    If implemented with a macro guard, this feature remains completely optional and non-breaking.
>    Existing applications would behave exactly the same, while backends or advanced users could enable the callback when needed.

There is no breaking change regardless.  Nuklear is not used as a dynamic library that will be updated separately from the application(s) that use it. It is meant to be compiled statically. And the changes are not what I would consider "breaking": Either a couple added field to an internal structure, an extra utility function meant almost exclusively for backend implementers not average users, or both.  No change to the existing API, and nothing removed that could break someone's code.

> 5. **Contextual data**(mostly for `SDL_SetTextInputArea`)

I still don't really understand the point of this function but all the logic above applies here too afaict. The difficulty is in getting the information in the first place.

> 6. Consistency with existing Nuklear design
>    This is just my opinion, but adding a virtual keyboard callback is not a hacky workaround, it’s actually a clean and natural extension of Nuklear’s design. It fits perfectly into the existing family of callbacks (copy, paste, etc.) and keeps the logic modular, explicit, and easy to reason about, exactly in the spirit of Nuklear’s minimalistic philosophy.

None of them are hacky workarounds. As I said all would require roughly the same logic/changes, the callback would just require a bit more code for the macro stuff at the top. My first attempt at doing this was a macro based callback system a la NK_IS_WORD_BOUNDARY() which I worked on a while back. It was two optional macros NK_ON_EDIT_FOCUS()/NK_ON_EDIT_UNFOCUS().  I ended up deleting it and replacing it twice, ending up on my current version because I realized everything that I discussed above. Not only do we not want or need to call Start/StopText() within Nuklear (even indirectly), but everything boils down to correctly and efficiently tracking the boolean state so the backend/user can do it themselves if they want to. 

What should be documented is all the knowledge about how that state is tracked which is what someone implementing a new text input widget not based on the nk_do_edit() family needs to know. As long as they remember the few lines to set the state appropriately then everything would just work.


>It’s also important to note that nk_do_edit is only called by text fields (e.g. nk_edit_string) and nk_property. This means the callback would directly address the exact issue we are discussing, without affecting unrelated parts of Nuklear.
>I believe I’ve given enough reasons why a new callback is preferable. Modifying nk_edit_is_any_active would require many changes in Nuklear, which almost equivalent to new bugs.

See above. It's not that many changes and you would have to touch the same pieces of the code for the callback method.

sleeptightAnsiC:
EDIT: Hiding... This did not work out the way I hoped. See: https://github.com/Immediate-Mode-UI/Nuklear/pull/852#issuecomment-3535023688

Also, a small side note: I'm planning to change `nk_edit_is_any_active` to `nk_edit_get_active`. It will return the pointer to the active edit widget (or NULL), instead of true/false, so the semantic remains similar but you get more control. If I understand right, you can't have two edit widgets active (unless it's a popup stealing the focus) so this should make sense. This will hopefully allow us for hooking that SDL function. I'm not sure, however, if implementing this edit_get... utility is entirely possible. I don't remember weather global contex had a reference to the edit widget (I hope it has). And that tricky SDL function needs a special cursor offset... so yea, the problem still wouldn't be fully solved.

sleeptightAnsiC:
EDIT: Hiding... I missed the message where it was initially mentioned. This is already reported by https://github.com/Immediate-Mode-UI/Nuklear/issues/856 

TODO (not required for this PR):

There is also one annoying thing with the edit fields: if the edit is active and you hover over different window, the edit becomes inactive (as expected) but the edit cursor still appears as if it was active (unexpected):
image
image

For comparison, this is how terminals deal with this [active on left, inactive on right]:
image

src/nuklear_edit.c
```diff
@@ -833,4 +833,12 @@ nk_edit_string_zero_terminated(struct nk_context *ctx, nk_flags flags,
...
     return result;
 }
+NK_API nk_bool
+nk_edit_is_any_active(const struct nk_context *ctx)
+{
+    NK_ASSERT(ctx);
+    if (!ctx) return nk_false;
 
+    return (ctx->active && ctx->active->edit.active) ||
+           (ctx->active && ctx->active->popup.win && ctx->active->popup.win->edit.active);
```
sleeptightAnsiC: replaying to https://github.com/Immediate-Mode-UI/Nuklear/pull/852#issuecomment-3518639254

> I also added a text field to the small demo window to test the differences and discovered something else. Text input stays active for a text field even if you click on (for instance) a color drop down, even if you actually click around in the color box or drag the nested channel properties. But it goes inactive if it was the property that was being edited. Just another small thing to dig into in the combobox/dropdown code.

Reproduced! and I'm pretty sure there is a bug on this line, because combobox/colorpicker is actually a "popup" under the hood. It is definitely wrong to consider the edit from the window active, while the popup without active edit is open.

image

sleeptightAnsiC: Can be easily fixed with:

```diff
diff --git a/src/nuklear_edit.c b/src/nuklear_edit.c
index 6e34664..89d7c59 100644
--- a/src/nuklear_edit.c
+++ b/src/nuklear_edit.c
@@ -839,6 +839,9 @@ nk_edit_is_any_active(const struct nk_context *ctx)
     NK_ASSERT(ctx);
     if (!ctx) return nk_false;
 
-    return (ctx->active && ctx->active->edit.active) ||
-           (ctx->active && ctx->active->popup.win && ctx->active->popup.win->edit.active);
+    if (!ctx->active)
+        return nk_false;
+    if (ctx->active->popup.win)
+        return (nk_bool)(ctx->active->popup.win->edit.active);
+    return (nk_bool)(ctx->active->edit.active);
 }
 ```

image

sleeptightAnsiC: Replying to https://github.com/Immediate-Mode-UI/Nuklear/pull/852#issuecomment-3518639254

> Similarly, I'm sure when I look at the combobox/dropdown code will have its own state and will be missing the lines to turn off an active edit. I'll do that later if someone else hasn't already by then.

I'm pretty sure, this was the same issue.

sleeptightAnsiC:
EDIT: outdated... hiding...

Replaying to https://github.com/Immediate-Mode-UI/Nuklear/pull/852#issuecomment-3518277340

> not sure if setting the name really matters but I was trying to match code in nk_edit_buffer

This is generally desired and probably every widget should do this, however, I just tested this and some are not setting/clearing the edit.name hash, which complicates my life a lot. Hah (I'm crying inside)

---

Replaying to https://github.com/Immediate-Mode-UI/Nuklear/pull/852#issuecomment-3519350868

> I'm planning to change nk_edit_is_any_active to nk_edit_get_active
> [...]
> I don't remember weather global contex had a reference to the edit widget (I hope it has).

Welp, it does not. It only has the `edit.name` hash and this doesn't really help, because this hash is only useful for whoever set it in the first place and can be literally everything. We could try to compare the `window->edit.name` with something like `window->property.name` and do some magic accordingly, but that would be short-sighted and fragile.

---

~At this point, I think I can add two fields to `nk_edit_state`, one for edit bounds and one for cursor bounds, and I can try hooking them somewhere inside of `nk_do_edit`. I'm not entirely sure if that would be a right decision. I've taken a look at what Pavel tried to do in https://github.com/Immediate-Mode-UI/Nuklear/pull/857 and he generally had a similar idea, but chosen `nk_keyboard` instead. Me attempting to fix it here could make his changes from https://github.com/Immediate-Mode-UI/Nuklear/pull/857 incompatible, so that's another thing I'm trying to avoid.~

EDIT: this last part would not work, see my next comment.

PavelSharp:
> At this point, I think I can add two fields to `nk_edit_state`

I believe that the `nk_edit_state` is not quite right place since we should take into account user-defined widgets which are not based on `nk_do_edit` and `nk_edit_state`

sleeptightAnsiC:
> > At this point, I think I can add two fields to `nk_edit_state`
> 
> I believe that the `nk_edit_state` is not quite right place since we should take into account user-defined widgets which are not based on `nk_do_edit` and `nk_edit_state`

I mistaken `struct nk_edit_state` for `struct nk_text_edit`, so it wouldn't work anyway, since `nk_do_edit` only has access to the latter one. (or maybe I didn't, I'm not entirely sure where I got this assumption from)

However, I don't believe we can take user-defined text widgets into consideration, if those aren't being implemented via `nk_do_edit` (or something similar). Many primitives from nk_edit_* API will not work for those widgets anyway, so why would someone expect the new call (nk_edit_is_active/nk_edit_get_active) to work without it?

Your solution from https://github.com/Immediate-Mode-UI/Nuklear/pull/857 also triggers the ~~callbacks~~ updates from `nk_do_edit`, so how the _"not based on nk_do_edit"_ case is being solved there? What do you mean by _"user-defined widgets which are not based on nk_do_edit"_ ?

PavelSharp:
> Your solution from https://github.com/Immediate-Mode-UI/Nuklear/pull/857 also triggers the callbacks from nk_do_edit, so how the "not based on nk_do_edit" case is being solved there? What do you mean by "user-defined widgets which are not based on nk_do_edit" ?

Let's imagine a user-defined widget that wants to notify Nuklear's input system (and backend), that it is about to receive text keyboard input.

In this case, the widget is responsible for calling `nk_input_want_text_keyboard`, and if that succeeds, it must populate the `text_cursor` and `text_area` fields. In other words, the responsibility for requesting text keyboard input and filling these fields lies entirely with the widget.

A clear example of this is `nk_do_edit`, which already handles this, but any other widget could do the same.

The main point is that `text_want`, `text_area`, and `text_cursor` belong to the input system’s notification mechanism - a new concept which was introduced by my PR. We update `nk_keyboard` with these fields, to form the input notification system. Since the backend checks them before `nk_input_begin`, no callbacks are needed.

[**NEW PART**]

> However, I don't believe we can take user-defined text widgets into consideration, if those aren't being implemented via nk_do_edit (or something similar).

In fact, the existence of such widgets is also assumed by @rswinkle  in his[ message:](https://github.com/Immediate-Mode-UI/Nuklear/pull/852#issuecomment-3519281011)

> What should be documented is all the knowledge about how that state is tracked which is what someone implementing a new text input widget not based on the nk_do_edit() family needs to know. As long as they remember the few lines to set the state appropriately then everything would just work.

sleeptightAnsiC:
I've taken a better look at https://github.com/Immediate-Mode-UI/Nuklear/pull/857 and I'm a bit confused.

> the widget is responsible for calling nk_input_want_text_keyboard, and if that succeeds, it must populate the text_cursor and text_area fields [...] a new concept which was introduced by my PR

Why would people follow this very specific _"concept"_, if they're not even following an already established solution like `nk_do_edit()` ? This reasoning contradicts itself and can be applied the other way around.

> We update nk_keyboard with these fields, to form the input notification system. Since the backend checks them before nk_input_begin, no callbacks are needed.

This is not a notification, no one is getting notified about anything. The user of https://github.com/Immediate-Mode-UI/Nuklear/pull/857 would still have to pull that information manually from `ctx->input.keyboard` (or similar) which also expects that the widget must set and update it correctly.

This really feels like repeating the old mistake. We already proven that widgets are not to be trusted with setting the context state correctly, and against all reasoning, this new "concept" introduces yet another state that we have to track and care about. It does not create any convention that custom widgets could safely follow.

I don't get it. All your previous messages were about a _"callback"_ system (I see 43 matches when I search for this word on this page). You've tried very hard to convince us to _"calback"_ solution, so why this new thing ended up being another manual _"state"_ update? I'm really confused here. Note that [I only disagreed about callbacks being a part of _this_ PR](https://github.com/Immediate-Mode-UI/Nuklear/pull/852#issuecomment-3519049014), I still thought you would go for it later in another PR (https://github.com/Immediate-Mode-UI/Nuklear/pull/857).

PavelSharp:
> This is not a notification, no one is getting notified about anything. The user of https://github.com/Immediate-Mode-UI/Nuklear/pull/857 would still have to pull that information manually from ctx->input.keyboard (or similar)

That’s not quite right. The backend is the one that needs to read this information, the user doesn’t have to worry about it.
I only used the word “notification” to help explain the idea. I often work with publisher-subscriber patterns, so the term just felt natural to me in this context.

> also expects that the widget must set and update it correctly.

Yes, if a user wants to create a custom, non-standard text-input widget, they should use this new simple API. In this very rare situation, this behavior is completely normal, and it has already been mentioned here(quoting it for the second time):

> What should be documented is all the knowledge about how that state is tracked which is what someone implementing a new text input widget not based on the nk_do_edit() family needs to know. As long as they remember the few lines to set the state appropriately then everything would just work.

**Next**

> This really feels like repeating the old mistake. We already proven that widgets are not to be trusted with setting the context state correctly, and against all reasoning

It’s possible that the earlier arguments weren’t convincing, because #857 **works very well and very stable**. I haven’t been able to find any issues, even when testing different combinations of clicks and focus changes. 
Also, #857 results in only about 30 lines of final changes and it’s important to recognize that these 30 lines are the conclusion of a long and difficult effort to properly integrate SDL3.

> this new "concept" introduces yet another state that we have to track and care about. It does not create any convention that custom widgets could safely follow.

**First**, this new state exists only until the next call to `nk_input_begin`, which normally corresponds to a single frame. Because of that, introducing this state is harmless. 

**Second**, please read my quote again:

> calling nk_input_want_text_keyboard, and if that succeeds, it must populate the text_cursor and text_area fields.

So, the rule is very simple.

> I don't get it. All your previous messages were about a "callback" system (I see 43 matches when I search for this word on this page). You've tried very hard to convince us to "calback" solution, so why this new thing ended up being another manual "state" update? I'm really confused here. 

In the end, my opinion changed. You may want to look more closely at #857, where it is explicitly mentioned:

> **The chosen design is** [**inspired by Dear ImGui**](https://github.com/ocornut/imgui/blob/c254db763709e12eeabea54c52631c2b972b7d58/backends/imgui_impl_sdl3.cpp#L178-L207). ImGui demonstrates that callback-based control can be unnecessarily heavy

**as well as:**

> Since the backend checks them before `nk_input_begin`, no callbacks are needed.

sleeptightAnsiC:
The latest force-push brings following changes:
- Fix for initial font scaling, and added a note about how to deal with potential rescaling at runtime
- Fix for nk_property not updating 'edit.active' state
- Added a check that triggers when something else than demo is trying to SDL_SetTextInputActive
- Moved the SDL_ConvertEventToRenderCoordinates from `nk_sdl_handle_event` to the `main.c` layer (I think the developer should be handling this)
- Given the latest discoveries about how Nuklear Widgets handle its states, I...
- ...nuked `nk_edit_is_any_active()` and moved its functionality to demo to `nk_sdl_update_TextInput()`
- ...nuked the debug label from `overview.c` and moved its functionality to demo to `main.c`
- ...introduced new demo function `nk_sdl_update_TextInput()` and left there a FIXME about possible SDL_SetTextInputArea implementation
- PR is again split into two separate commits, so I re-converted it from Draft

I hope, this will bring a compromise to how TextInput should be tracked and updated, at least for now. PR no longer changes nk_* API. I made it so it will be easy to tweak TextInputActive and solve TextInputArea in the future. I may have some disagreement with Pavel, but I don't want to sabotage his work from https://github.com/Immediate-Mode-UI/Nuklear/pull/857 This is the most "friendly" solution I can see at the moment, and I hope that everyone will understand.

Please, do not drag this PR for any longer with features that can be solved later.
Let's retest this thing one last time, merge it, and move onto other problems.
There is so much I want to upstream and this demo is taking all my attention.

@PavelSharp @rswinkle  Maybe, let's hide our latest off-topic comments, and/or move their content to https://github.com/Immediate-Mode-UI/Nuklear/pull/857 ?

src/nuklear_edit.c
sleeptightAnsiC: fixed with latest change

PavelSharp:
@sleeptightAnsiC 

> Maybe, let's hide our latest comments, and/or move their content to #857?

Yes, I agree. This discussion relates much more to #857. So let's hide our messages here, and I'll move it.

> I may have some disagreement with Pavel, but I don't want to sabotage his work from https://github.com/Immediate-Mode-UI/Nuklear/pull/857

Actually, I'm open to your suggestions. If you have _specific_ ideas about what changes we should make in #857, please feel free to share them. I appreciate it

sleeptightAnsiC:
> Wasn't able to run it since I'll need a CMake implementation

@RobLoach Hey, if you want to test this PR, the old file from https://github.com/Immediate-Mode-UI/Nuklear/pull/779 works just fine after few changes. Here is an updated version [attachment below]. Just make sure to drop it into the same directory as this demo.

[CMakeLists.txt](https://github.com/user-attachments/files/23565131/CMakeLists.txt)

I'm not sure why it broke the last time I tested it. This time it "just" worked. :flushed: I guess I can add it back as a separate commit (?) I'm not really a Cmake specialist, so I can't tell how good/portable it is.

RobLoach:
Thanks so much for all of the push on this, @sleeptightAnsiC , @PavelSharp , @rswinkle , @zoogies . So happy to have this finally in.

zoogies:
Woo! Thanks, all. Glad to see this merged! Shoutout to @sleeptightAnsiC for taking the lead on this

sleeptightAnsiC:
It's over... I'm released from my burden. Thanks for all testing and feedback. This was way bigger and harder to deal with, than I ever expected it to be, hah. Hopefully, all bugs found along the way will be soon followed by another fixes :smiley:

demo/common/overview.c
alimourad-stack: Fixed in e8d0585e — good catch on the tie-out not being a backstop for this.

Both coercions now throw:
- `legTotal` rejects a null month by name, since if *both* legs lost the same month `rate + volume` still balances against the parent and the split would foot while silently omitting a month.
- `buildPeriodSplit` rejects a null `parent.var`, which would otherwise let the split tie out against a row rendering an em dash.

Also dropped the now-dead `parent.var ?? 0` in the mismatch message, since the new guard narrows the type.

sleeptightAnsiC: huh?

PR #860, opened 2025-11-16
demo/xcb_cairo/nuklear_xcb.h: do not discard xcb_client_message_event_t

Shamar:
Currently nk_xcb_handle_event discard everything that was sent from other clients through xcb_send_event.

To enable a minimal IPC over such channel, this patch adds a new option NK_XCB_CLIENT_MESSAGE to the nk_xcb_event_type enum and stores the message recieved in xcb_ctx->last_client_message.

This can be useful in particular when a background process or thread completes a data retrieval or processing and want to wake up the GUI.

### DIFF
--- a/demo/xcb_cairo/nuklear_xcb.h
+++ b/demo/xcb_cairo/nuklear_xcb.h
@@ -238,13 +238,12 @@ NK_API void nk_xcb_free(struct nk_xcb_context *xcb_ctx)
     free(xcb_ctx);
 }
 
-static const xcb_client_message_event_t EMPTY_CLIENT_MESSAGE;
-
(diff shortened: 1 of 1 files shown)


riri:
LGTM, maybe just move the empty message variable as a local static in  `nk_xcb_handle_event`, avoiding a global variable

Shamar:
> LGTM, maybe just move the empty message variable as a local static in `nk_xcb_handle_event`, avoiding a global variable

Good catch @riri.
Addressed in https://github.com/Immediate-Mode-UI/Nuklear/pull/860/changes/dc4e69d5038eb9794d5fb7220508b91bc8ecd477

PR #862, opened 2025-11-18
Fix several issues related to building from Nuklear/src/* files

sleeptightAnsiC:
This PR fixes several issues related to build process and symbol resolution through preprocesor.

I separated it into ~4~ **6!** logical commits, each one fixing slightly different issue.

I could submit them as separate PRs, but this would be very impractical to deal with, since all those changes have some logical overlap in the code.

For better explanation of each change, please read the corresponding commit message.

Closes: https://github.com/Immediate-Mode-UI/Nuklear/issues/256
Closes: https://github.com/Immediate-Mode-UI/Nuklear/issues/851

### DO NOT SQUASH!

### DIFF
--- a/.github/ci_compile_sources.sh
+++ b/.github/ci_compile_sources.sh
@@ -1,0 +1,48 @@
+#!/bin/sh
+# This shellscript compiles each Nuklear/src/*.c source separetly
+# in order to check for simple compilation failures by our CI runner.
+
+# FIXME: This script compiles WITHOUT LINKING, which means that it won't catch
+# any potential failures at link time. We don't have any example and/or demo
+# that would work without amalgamated header, so we can't really test linking
+# at the moment (but we defenetily should do this in the future!)
+
+# FIXME: This script currently lives in Nuklear/.github/* folder because
+# there are no other scripts like this one, and having it somewhere else
+# could confuse people. Same reason why it wasn't made as Makefile.
+
+set -e
+
+CC=cc
+SRCDIR=./src
+
+set -- ""
+set -- "$@" -std=c89
+set -- "$@" -Wall
+set -- "$@" -Wextra
+set -- "$@" -pedantic
+CFLAGS=$*
+
+set -- ""
+set -- "$@" -DNK_INCLUDE_FIXED_TYPES
+set -- "$@" -DNK_INCLUDE_DEFAULT_ALLOCATOR
+set -- "$@" -DNK_INCLUDE_STANDARD_IO
+set -- "$@" -DNK_INCLUDE_STANDARD_VARARGS
+set -- "$@" -DNK_INCLUDE_STANDARD_BOOL
+set -- "$@" -DNK_INCLUDE_VERTEX_BUFFER_OUTPUT
+set -- "$@" -DNK_INCLUDE_DEFAULT_FONT
+set -- "$@" -DNK_INCLUDE_COMMAND_USERDATA
+CPPFLAGS=$*
+
+retcode=0
+
+for i in "${SRCDIR}"/*.c ; do
+    printf "CC %s\n" "${i}"
+    # shellcheck disable=SC2086
+    if ! $CC -x c -c "${i}" -o /dev/null $CFLAGS $CPPFLAGS ; then
+        retcode=1
+    fi
+done
+
+exit "${retcode}"
+
--- a/.github/workflows/ccpp.yml
+++ b/.github/workflows/ccpp.yml
@@ -49,3 +49,5 @@ jobs:
       run: make -C demo/xcb_cairo
     - name: build example
       run: make -C example
+    - name: compile sources
+      run: sh -e ./.github/ci_compile_sources.sh


sleeptightAnsiC:
EDIT: added CI for this in https://github.com/Immediate-Mode-UI/Nuklear/pull/862/changes/3db34e4db2f6ad8eec92302897f69afcccefe3a9

In order to test the compilation from sources (without amalgamation header), you can use this:
```sh
$ git clone https://github.com/sleeptightAnsiC/Nuklear.git --branch fix__math_util_build
$ cd Nuklear/src
$ for i in $(ls ./*.c); do echo "$i"; cc -x c -c "$i" -o "$i.o" -DNK_INCLUDE_FIXED_TYPES -DNK_INCLUDE_DEFAULT_ALLOCATOR -DNK_INCLUDE_STANDARD_IO -DNK_INCLUDE_STANDARD_VARARGS -DNK_INCLUDE_STANDARD_BOOL -DNK_INCLUDE_VERTEX_BUFFER_OUTPUT -DNK_INCLUDE_DEFAULT_FONT -DNK_INCLUDE_COMMAND_USERDATA ; done
```

PavelSharp:
Readed and verified by me. I have carefully reviewed the changes and compiled it successfully. The work looks good, and very canonical in style of Nuklear. We may still need to do deep testing, although the changes are small.

demo/sdl3_renderer/main.c
```diff
@@ -75,8 +75,9 @@
 #define NK_VSNPRINTF(s, n, f, a)  SDL_vsnprintf(s, n, f, a)
 #define NK_STRTOD(str, endptr)    SDL_strtod(str, endptr)
 
-/* sadly, SDL3 does not provide "dtoa" (only integer version) */
-/*#define NK_DTOA (str, d)*/
+/* HACK: SDL3 does not provide "dtoa" (only integer version)
+ * so we use sprintf with %g formatting to get almost the same result */
+#define NK_DTOA(str, d) (SDL_snprintf(str, 999, "%g", d), str)
```
sleeptightAnsiC: EDIT: I've decided that having a hacky fragile macro is not really desired, and fixed it with https://github.com/Immediate-Mode-UI/Nuklear/pull/862/changes/7f0a86e99246147eb9d5ae52d3a3264336365335

This is quite _evil_ but works just fine. Comma-operator with "`((...), str)`" at the very end makes sure that we get the proper return value. The `999` indicates that we don't care about buffer length (dtoa is unsafe anyway). Formatting with `%g` turns longer values into scientific notation (nk_dtoa does a very same [thing](https://github.com/Immediate-Mode-UI/Nuklear/blob/9afb3dd6d14fba0c5e62f08809b494b6d580b9b1/src/nuklear_util.c#L523-L538)). I hope I didn't miss any case.

src/nuklear.h
```diff
@@ -3817,6 +3817,7 @@ NK_API int nk_strtoi(const char *str, char **endptr);
 NK_API float nk_strtof(const char *str, char **endptr);
 #ifndef NK_STRTOD
 #define NK_STRTOD nk_strtod
+#define NK_STRTOD_IMPLEMENTATION
```
sleeptightAnsiC: Note: `nk_strtod` is the only public utility function that you can overwrite with the macro. Other similar ones have their declarations kept private inside of `src/nuklear_internal.h`. I thought this is worth mentioning (but I don't remember why, lol)

sleeptightAnsiC (commented):
So, I've retested everything, and the only problem I found is this non-fatal warning [see below]. It happens only if you provide your own `NK_DTOA`, and only nuklear_sdl3_renderer is able to trigger it right now. Basically, nk_dtoa implementation was using all those functions listed below, and when user provides it's own implementation, they remain "`defined but not used`". It breaks nothing, just triggers a warning, that's all.

```sh
~/Nuklear (fix__math_util_build) $ cd demo/sdl3_renderer/
~/Nuklear/demo/sdl3_renderer (fix__math_util_build) $ make
mkdir -p bin/
cc main.c -o bin/demo -MD -MF ./bin/demo.d -std=c89 -Wall -Wextra -Wpedantic -O2  -lm -lSDL3
In file included from main.c:116:
../../nuklear.h:6626:1: warning: ‘nk_log10’ defined but not used [-Wunused-function]
 6626 | nk_log10(double n)
      | ^~~~~~~~
../../nuklear.h:6602:1: warning: ‘nk_ifloord’ defined but not used [-Wunused-function]
 6602 | nk_ifloord(double x)
      | ^~~~~~~~~~
../../nuklear.h:6587:1: warning: ‘nk_pow’ defined but not used [-Wunused-function]
 6587 | nk_pow(double x, int n)
      | ^~~~~~
```

There is nothing wrong about this warning really and I don't believe this is even worth fixing.

That said, I'm turning this back into mergable PR.

sleeptightAnsiC:
~~@RobLoach ready to merge, unless you have some objections :)~~

EDIT: ooops, I need to take a better look at it again: https://github.com/Immediate-Mode-UI/Nuklear/pull/862#discussion_r2542435759

demo/sdl3_renderer/main.c
sleeptightAnsiC: > I hope I didn't miss any case.

Actually, I missed the `str == NULL` case, I thought that SDL handles it ~(because [it does... just not correctly](https://github.com/libsdl-org/SDL/blob/168f78b8b76c4a91cd744b03a816613a8273e250/src/stdlib/SDL_string.c#L1836-L1838)).~

~EDIT: Actually actually, SDL makes a correct nullcheck, but [only for it's build-in version](https://github.com/libsdl-org/SDL/blob/168f78b8b76c4a91cd744b03a816613a8273e250/src/stdlib/SDL_string.c#L2140-L2145). If it uses the libc version, it has no such check, lol :D~

~I'll fix it in few days. I should probably test this thing more deeply.~

EDIT: nah, the doc for SDL_snprintf clearly says that neither of pointers should be null, so it does not matter what SDL does internally. dtoa is unsafe and should never take null. However, I still want to ensure this macro is fully correct, and I don't like its current form. I think, I will just reimplement it into the function.

RafaDEV99 (approved):
Ok, this looks good to me! Additionally, it is beneficial to use as many code implementation macros for this use case (because before it was a mess)

sleeptightAnsiC:
Latest force-push fixed all concerns mentioned by my previous comments.

I also added a CI runner in order to prevent https://github.com/Immediate-Mode-UI/Nuklear/issues/256 from reoccurring. We don't really have any other solution for testing those kinds of issues, so I figured out the shellscript would be enough for now.

Anyone here mind retesting this one, please?

src/nuklear.h
RobLoach: This brings up a good point. Should we also protect with `NK_STRTOD_IMPLEMENTATION_ONCE` defines?

src/nuklear_util.c
```diff
@@ -13,8 +13,7 @@ NK_LIB nk_bool nk_is_upper(int c){return (c >= 'A' && c <= 'Z') || (c >= 0xC0 &&
 NK_LIB int nk_to_upper(int c) {return (c >= 'a' && c <= 'z') ? (c - ('a' - 'A')) : c;}
 NK_LIB int nk_to_lower(int c) {return (c >= 'A' && c <= 'Z') ? (c - ('a' + 'A')) : c;}
 
-#ifndef NK_MEMCPY
-#define NK_MEMCPY nk_memcopy
+#ifdef NK_MEMCPY_IMPLEMENTATION
```
RobLoach: Should we only be implementing the function if it's not needed?

If you're implemnting your own nk_memcopy, we don't need this implementation.

sleeptightAnsiC: I'm a bit confused about what you're asking here.

The `NK_MEMCPY_IMPLEMENTATION` macro will NOT be defined when you're implementing your own memcpy (aka `#define NK_MEMCOPY my_own_memcpy` )

The `nk_memcopy` is private (marked as `NK_LIB` and hidden within `src/nuklear_internal.h` ). You shall not try to re-implement and/or use it inside of your application code. Use `NK_MEMCOPY` for this purpose instead.

This implementation is only being compiled when `#ifdef NK_MEMCPY_IMPLEMENTATION` **_AND_** `#ifdef NK_IMPLEMENTATION`, and this can only happen once, and only when you aren't providing your own memcpy.

So what is the problem?

Maybe the naming convention is simply too confusing?
How about we change it: NK_XXX_IMPLEMENTATION -> NK_XXX_NEEDED ?
Note that all those NK_XXX_IMPLEMENTATION macros are private anyway.

src/nuklear.h
sleeptightAnsiC: > Should we also protect with NK_STRTOD_IMPLEMENTATION_ONCE defines?

No, we don't need include-like-guards here.

I belive we have some misunderstanding due to how this macro is named. See https://github.com/Immediate-Mode-UI/Nuklear/pull/862#discussion_r2726008172

> Note: nk_strtod is the only public utility function that you can overwrite with the macro. Other similar ones have their declarations kept private inside of src/nuklear_internal.h. I thought this is worth mentioning (but I don't remember why, lol)

I think, now I remember why... You see, this symbol is public, end user should ALWAYS be able to access it... and yet, it is being conditionally defined with `#ifndef NK_STRTOD`. This seems like a potential bug to me.

src/nuklear_util.c
RobLoach: Needed does seem more informative since we are using implementation elsewhere. Thanks.

sleeptightAnsiC: Renamed to NK_XXX_NEEDED in latest force-push

Good for calling this out. I totally missed that we have bunch of NK_NO_STB_XXX_IMPLEMENTATION macros all over the place. So that naming would be indeed confusing.

sleeptightAnsiC:
> The ci_compile_sources.sh seems unrelated to this change? I'm fine with it being in there though.

@RobLoach That was intentional. It tests for similar regressions. See: https://github.com/Immediate-Mode-UI/Nuklear/pull/862#issuecomment-3797241538

RobLoach:
Thanks for adding the docs around it too 👍

PR #864, opened 2025-12-11
Fix windows gdiplus demo runtime panic

tisyang:
On windows 11, run demo/gdip, will have a runtime panic like below:

屏幕截图 2025-12-11 160939

    thread 11332 panic: left shift of 255 by 24 places cannot be represented in type 'int' in `nuklear_gdip.h:422:0`:
    return (c.a <<  24) | (c.r << 16) | (c.g << 8) | c.b;

As `ARGB` is `DWORD`(aka `u32`), `c.a` is `nk_byte`(aka `u8`), `c.a << 24` will result in `int` type(Integer Promotion), then the runntime checker will panic.

So, cast `c.a` to `ARGB` will fix this.

### DIFF
--- a/demo/gdip/nuklear_gdip.h
+++ b/demo/gdip/nuklear_gdip.h
@@ -419,7 +419,7 @@ static struct {
 
 static ARGB convert_color(struct nk_color c)
 {
-    return (c.a << 24) | (c.r << 16) | (c.g << 8) | c.b;
+    return ((ARGB)c.a << 24) | (c.r << 16) | (c.g << 8) | c.b;
 }
 
 static void


tisyang:
I realize that, the panic is cause by zig compiler:

This panic is caused by Zig's runtime safety checks detecting undefined behavior, integer flow is UB. These checks are enabled in Debug builds but disabled in ReleaseFast mode.

So, it's not a code bug for C.

sleeptightAnsiC:
Zig runs C with a modified version of Clang, so this might be an actual code bug. Not sure which Sanitizer triggers this (probably UBsan) but you may be able to reproduce this without Zig.

How about you report it as a separate Issue? I may take a look into it someday. 

I wonder why this is being triggered only for `255<<24`. Maybe because C has this strage rule where it implicitly converts "small" integer types to `int` during all kinds of arithmetic operations, and it may cause an overflow in this exact case (?) 

EDIT: I'm talking about: https://stackoverflow.com/questions/46073295/implicit-type-promotion-rules

tisyang:
> Zig runs C with a modified version of Clang, so this might be an actual code bug. Not sure which Sanitizer triggers this (probably UBsan) but you may be able to reproduce this without Zig.
> 
> How about you report it as a separate Issue? I may take a look into it someday.
> 
> I wonder why this is being triggered only for `255<<24`. Maybe because C has this strage rule where it implicitly converts "small" integer types to `int` during all kinds of arithmetic operations, and it may cause an overflow in this exact case (?)
> 
> EDIT: I'm talking about: https://stackoverflow.com/questions/46073295/implicit-type-promotion-rules

@sleeptightAnsiC In short, zig's c runtime library raise the panic.

I have made a very simple c code with zig build:

```c
// main.c
#include <stdio.h>

int main(int argc, char **argv)
{
    unsigned char c = 255;
    printf("255<<24=%d\n", c<<24);
    return 0;
}

```
```zig
// build.zig
const std = @import("std");

pub fn build(b: *std.Build) void {
    const target = b.standardTargetOptions(.{});
    const optimize = b.standardOptimizeOption(.{});

    const exe = b.addExecutable(.{
        .name = "test-int",
        .root_module = b.createModule(.{
            .target = target,
            .optimize = optimize,
        }),
    });
    exe.addCSourceFile(.{
        .file = b.path("main.c"),
        .flags = &.{},
    });
    exe.linkLibC();
    b.installArtifact(exe);

    const run_cmd = b.addRunArtifact(exe);
    run_cmd.step.dependOn(b.getInstallStep());
    if (b.args) |args| {
        run_cmd.addArgs(args);
    }
    const run_step = b.step("run", "Run the app");
    run_step.dependOn(&run_cmd.step);
}

```

On windows, when build with zig and run, either `Debug` or `-Doptimize=ReleaseFast`, will panic with same error `panic: left shift of 255 by 24 places cannot be represented in type 'int'`:
屏幕截图 2025-12-15 115823

The error occurs in zig's mingw c runtime library.

I also built the code with tdm-gcc(aka, mingw64) and run, no panic:
屏幕截图 2025-12-15 120041

So, the panic should only be raised in zig build, in **zig's c runtime library**.

As a test in linux x86, I built the example under debian and run end up with same panic:
屏幕截图_20251215_134119

sleeptightAnsiC:
> In short, zig's c runtime library raise the panic. [...] I also built the code with tdm-gcc(aka, mingw64) and run, no panic [...] So, the panic should only be raised in zig build, in zig's c runtime library.

No, this is not exactly how it works. Zig adds runtime "Sanitizers" by default when you build your project in `Debug` mode. You can get the same result in GCC by adding `-fsanitize=undefined` . In fact, I just reproduced your issues in C alone, with GCC and Clang, see: https://godbolt.org/z/ob1WadhW9

EDIT: So your PR would "fix" it here. I just wonder, if this is an actual UB, or just a false-positive from Sanitizer. It feels really strange that the shift is promoted to `signed int`.

tisyang:
> > In short, zig's c runtime library raise the panic. [...] I also built the code with tdm-gcc(aka, mingw64) and run, no panic [...] So, the panic should only be raised in zig build, in zig's c runtime library.
> 
> No, this is not exactly how it works. Zig adds runtime "Sanitizers" by default when you build your project in `Debug` mode. You can get the same result in GCC by adding `-fsanitize=undefined` . In fact, I just reproduced your issues in C alone, with GCC and Clang, see: https://godbolt.org/z/ob1WadhW9
> 
> EDIT: So your PR would "fix" it here. I just wonder, if this is an actual UB, or just a false-positive from Sanitizer. It feels really strange that the shift is promoted to `signed int`.

Oh, I didn't inspect this deeply before.

But, the `Promoted to signed int` is straightforward， as `int` can hold all value of `unsigned char`, so `unsigned char` is promoted to `int` before shift, and then shift will raise the Sanitizer complain.

PR #865, opened 2025-12-14
demo/sdl3_renderer/main.c: call nk_input_begin at the end of SDL_AppInit

sh1l0h:
The very first `nk_input_end` call in `SDL_AppIterate` occurs without a preceding `nk_input_begin` call. While not critical, this violates the expected pairing of these functions.

### DIFF
--- a/demo/sdl3_renderer/main.c
+++ b/demo/sdl3_renderer/main.c
@@ -409,6 +409,7 @@ SDL_AppQuit(void* appstate, SDL_AppResult result)
     NK_UNUSED(result);
 
     if (app) {
+        nk_input_end(app->ctx);
         nk_sdl_shutdown(app->ctx);
         SDL_DestroyRenderer(app->renderer);
         SDL_DestroyWindow(app->window);


sleeptightAnsiC:
Hi,

> While not critical, this violates the expected pairing of these functions.

Does the lack of first nk_input_begin cause any observable issue? or is there any documentation that mentions this being invalid? We intentionally made it this way and never had any issues since.

The [nk_input_end](https://github.com/Immediate-Mode-UI/Nuklear/blob/9afb3dd6d14fba0c5e62f08809b494b6d580b9b1/src/nuklear_input.c#L29-L43) only memzero'es few values, there is nothing wrong if you call it without preceding call to nk_input_begin. In fact, the first input cycle will do nothing, if you don't have any UI state created.

Also, if you're already adding nk_input_begin to SDL_AppInit, you should probably also add nk_input_end to SDL_AppQuit, just for the sake of consistency. Otherwise, even with your patch, you still have unbalanced begin/end calls.

sh1l0h:
Hey,

Fair enough. The input documentation does describe the pairing as the expected pattern, but if it's intentional and working as designed, that's fine.

You can close this if you'd prefer to keep it as-is.

Cheers

demo/sdl3_renderer/main.c
sleeptightAnsiC, line 411: > Also, if you're already adding nk_input_begin to SDL_AppInit, you should probably also add nk_input_end to SDL_AppQuit, just for the sake of consistency. Otherwise, even with your patch, you still have unbalanced begin/end calls.

```diff
diff --git a/demo/sdl3_renderer/main.c b/demo/sdl3_renderer/main.c
index d7e54e2..24d3502 100644
--- a/demo/sdl3_renderer/main.c
+++ b/demo/sdl3_renderer/main.c
@@ -406,6 +406,7 @@ SDL_AppQuit(void* appstate, SDL_AppResult result)
     NK_UNUSED(result);
 
     if (app) {
+        nk_input_end(app->ctx);
         nk_sdl_shutdown(app->ctx);
         SDL_DestroyRenderer(app->renderer);
         SDL_DestroyWindow(app->window);
```

^ I have no idea how to suggest one line addition so just posting a diff (I hate Github...)

sleeptightAnsiC:
> The input documentation does describe the pairing as the expected pattern

https://github.com/Immediate-Mode-UI/Nuklear/blob/729516425ae272809736c15eadd240a904f98f7e/src/nuklear.h#L522-L528

You're right, and although the docs do not explicitly forbid this, the unpaired input_end/begin does feel like a hack (it clearly depends on implementation details in order to work).

I'm only afraid that having additional input_begin/end may suggest something that doesn't exist, because this is really just a dead code, no actual input recording/consuming happens in those cases.

@RobLoach Feel free to merge. I would be happy if you could apply the diff from my comment above, but it isn't that big of a deal.

demo/sdl3_renderer/main.c
RobLoach: Umm, this maybe? Haha. Feel free to submit a follow up PR.
``` suggestion
    if (app) {
        nk_input_end(app->ctx);
```

RobLoach: I just commited in the branch

sleeptightAnsiC:
> I just commited in the branch

@RobLoach Looks good! Squash&Merge ?

sleeptightAnsiC:
@sh1l0h Thanks, and sorry for this one taking so long. I had it noted somewhere down the backlog, and just forgot about it. If you see anything else strange in `demo/sdl3_renderer`, let me know. Reporting an Issue before submitting PR is probably more preferable because, as you can see, it's not always so straight forward to accept it.

Issue #872, opened 2026-01-03
demos do not handle NK_UINT_DRAW_INDEX flag, resulting in broken rendering


sleeptightAnsiC:
Most of the demos do not handle `NK_UINT_DRAW_INDEX` in any way, which means that whenever said flag is defined, either the vertex data or the draw calls will break, which results in broken rendering like so: 

<img width="1195" height="796" alt="Image" src="https://github.com/user-attachments/assets/c05d7342-a02e-48c2-8b54-c2ee63490809" />

This was found by @Xeverous and mentioned on Nuklear discord server https://discord.com/channels/922234432071548968/922234432071548973/1456733473355206666

```
Xeverous — Yesterday at 8:38 PM
I have found an issue where defining NK_UINT_DRAW_INDEX completely destroys the rendering, I see shaded rectangles, no UI or any text. Happens with SDL backend. Will try to reproduce on pure C and minimal code.
I confirm. This is completely broken. Add -DNK_UINT_DRAW_INDEX to demo/sdl_opengl2/Makefile and see yourself.

sleeptightAnsiC — 12:40 PM
This is probably also broken with most other demos. Only two demos (d3d11 and d3d12) have a condition related to this flag. I suspect you could take a look at how d3d11 solves it, and try applying the same. https://github.com/Immediate-Mode-UI/Nuklear/blob/9afb3dd6d14fba0c5e62f08809b494b6d580b9b1/demo/d3d11/nuklear_d3d11.h#L89-L93

sleeptightAnsiC — 12:44 PM
Does this patch fixes the issue for you?
[see patch below]
^ This should work. I just tested it on my side.
```

[patch]

```diff
diff --git a/demo/sdl_opengl2/nuklear_sdl_gl2.h b/demo/sdl_opengl2/nuklear_sdl_gl2.h
index 18509f1..f9a0144 100644
--- a/demo/sdl_opengl2/nuklear_sdl_gl2.h
+++ b/demo/sdl_opengl2/nuklear_sdl_gl2.h
@@ -158,7 +158,11 @@ nk_sdl_render(enum nk_anti_aliasing AA)
                 (GLint)((height - (GLint)(cmd->clip_rect.y + cmd->clip_rect.h)) * scale.y),
                 (GLint)(cmd->clip_rect.w * scale.x),
                 (GLint)(cmd->clip_rect.h * scale.y));
+        #ifdef NK_UINT_DRAW_INDEX
[...]

PR #873, opened 2026-01-08
Reduce differences between the OpenGL 3 and OpenGL ES 2 demos

ccawley2011:
The differences between the OpenGL 3 and OpenGL ES 2 demos are:
- Vertex Array Objects were introduced in OpenGL 3.0 or OpenGL ES 3.0 - older versions require extensions.
- Memory mapped buffers require extensions - `glMapBufferRange` is in OpenGL 3.0 or OpenGL ES 3.0 and could be used as an alternative, but `glMapBuffer` is only available in OpenGL ES with extensions.
- The GLSL shader code is different.
- Different link libraries are required depending on which API should be used (see discussion in #777).

### DIFF
--- a/demo/x11_opengl3/nuklear_xlib_gl3.h
+++ b/demo/x11_opengl3/nuklear_xlib_gl3.h
@@ -77,8 +77,6 @@ typedef void (*nkglDeleteVertexArrays)(GLsizei, const GLuint*);
 typedef void(*nkglVertexAttribPointer)(GLuint, GLint, GLenum, GLboolean, GLsizei, const GLvoid*);
 typedef void(*nkglEnableVertexAttribArray)(GLuint);
 typedef void(*nkglDisableVertexAttribArray)(GLuint);
-/* GL_ARB_framebuffer_object */
-typedef void(*nkglGenerateMipmap)(GLenum target);
 /* GLSL/OpenGL 2.0 core */
 typedef GLuint(*nkglCreateShader)(GLenum);
 typedef void(*nkglShaderSource)(GLuint, GLsizei, const GLchar**, const GLint*);
@@ -114,7 +112,6 @@ static nkglDeleteVertexArrays glDeleteVertexArrays;
 static nkglVertexAttribPointer glVertexAttribPointer;
 static nkglEnableVertexAttribArray glEnableVertexAttribArray;
 static nkglDisableVertexAttribArray glDisableVertexAttribArray;
-static nkglGenerateMipmap glGenerateMipmap;
 static nkglCreateShader glCreateShader;
 static nkglShaderSource glShaderSource;
 static nkglCompileShader glCompileShader;
@@ -159,9 +156,7 @@ struct opengl_info {
     int glsl_available;
     int vertex_buffer_obj_available;
     int vertex_array_obj_available;
-    int map_buffer_range_available;
     int fragment_program_available;
-    int frame_buffer_object_available;
 };
 #endif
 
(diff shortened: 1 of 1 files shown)


RobLoach (approved):
Good to have some consistancy.

demo/glfw_opengl3/nuklear_glfw_gl3.h
```diff
@@ -151,40 +157,47 @@ nk_glfw3_device_create(struct nk_glfw* glfw)
     glGetProgramiv(dev->prog, GL_LINK_STATUS, &status);
     assert(status == GL_TRUE);
 
+    /* TODO: Detect at runtime */
```
sleeptightAnsiC, line 160: Is this unfinished? Are you planning to implement this in the force push?

(same for other files)

demo/glfw_opengl3/nuklear_glfw_gl3.h
```diff
@@ -37,6 +39,10 @@ struct nk_glfw_device {
     GLint uniform_tex;
     GLint uniform_proj;
     GLuint font_tex;
+    GLsizei vs;
+    size_t vp, vt, vc;
```
sleeptightAnsiC, line 43: What was the reason for moving this into the device object? We don't use this anywhere. It's only for initial setup in nk_glfw3_device_create and nothing else. Am I missing something?

demo/glfw_opengl3/nuklear_glfw_gl3.h
```diff
@@ -298,8 +337,8 @@ nk_glfw3_render(struct nk_glfw* glfw, enum nk_anti_aliasing AA, int max_vertex_b
                 (GLint)((glfw->height - (GLint)(cmd->clip_rect.y + cmd->clip_rect.h)) * glfw->fb_scale.y),
                 (GLint)(cmd->clip_rect.w * glfw->fb_scale.x),
                 (GLint)(cmd->clip_rect.h * glfw->fb_scale.y));
-            glDrawElements(GL_TRIANGLES, (GLsizei)cmd->elem_count, GL_UNSIGNED_SHORT, (const void*) offset);
-            offset += cmd->elem_count * sizeof(nk_draw_index);
```
sleeptightAnsiC, line 302: The `sizeof(nk_draw_index)` is not guaranteed to be `1` (you can change this size with preprocessor macro) so I'm not so sure you can safely remove this

sleeptightAnsiC (commented):
I find some of those changes a bit confusing, but I suspect you wanted to keep the code the same, in all files, at all costs. Sadly, because we have to deal with so many demos, the review process is really hard, so apologies if I missed something obvious.

Overall, I think that merging this without feature detection implemented doesn't seem very useful. Right now, most of those additions is just a dead code.

However, just like I mentioned in https://github.com/Immediate-Mode-UI/Nuklear/pull/777#issuecomment-3721549042 , I do agree that we need this.

demo/glfw_opengl3/nuklear_glfw_gl3.h
ccawley2011: The type of offset has changed from `nk_size` to `const nk_draw_index *`, so multiplying by `sizeof(nk_draw_index)` is no longer needed.

demo/glfw_opengl3/nuklear_glfw_gl3.h
ccawley2011: It's also used in `nk_glfw3_render` when vertex array objects are unsupported.

sleeptightAnsiC: Ah, yes, I missed it. Sorry

demo/glfw_opengl3/nuklear_glfw_gl3.h
ccawley2011: What is the current consensus regarding PR #777? Should I wait for that to be merged, or should this one be merged first and that one rebased on top? Or should the existing loader code from `x11_opengl3` be adapted for the other demos instead of using GLAD?

sleeptightAnsiC: > What is the current consensus regarding PR https://github.com/Immediate-Mode-UI/Nuklear/pull/777? Should I wait for that to be merged

I don't believe you will get https://github.com/Immediate-Mode-UI/Nuklear/pull/777 merged any time soon. I would ignore it for now.

> Or should the existing loader code from x11_opengl3 be adapted for the other demos instead of using GLAD?

I think this will be a way to go in the future. Right now, we have way too many duplicate code in the demos, so we need to deal with this problem before doing anything else.

@RobLoach (tactical ping)

demo/glfw_opengl3/nuklear_glfw_gl3.h
sleeptightAnsiC: Alright. I think, even if this is wrong, it wouldn't prevent issues like https://github.com/Immediate-Mode-UI/Nuklear/issues/872 anyway, so fine by me.

demo/glfw_opengl3/nuklear_glfw_gl3.h
RobLoach: Cleaning up the dup code would be nice. GLAD would ease up the system dependencies. This one consolidates the code. I like both, but sounds like there's still some stuff to figure out?

PR #874, opened 2026-01-21
Fix D3D12 pixel shader not passing texture index

therathatter:
User textures in the D3D12 backend are completely broken due to the fact that the 'texture_index' is always zero within the pixel shader due to the variable being static, so it always defaults to the font atlas texture.

Before:
image

After:
image

Also - if this gets merged, can we also please merge https://github.com/Immediate-Mode-UI/Nuklear/pull/487? This fix works fine for me and is necessary to even get the demo to work at all on modern Windows SDK versions.


sleeptightAnsiC:
@RobLoach checked with MSVC 19.14.33813 (VS2022) in Windows11 VM, and it works just fine. Let's merge it :)

> Also - if this gets merged, can we also please merge https://github.com/Immediate-Mode-UI/Nuklear/pull/487? This fix works fine for me and is necessary to even get the demo to work at all on modern Windows SDK versions.

That other PR has unresolved request

therathatter:
Thanks!

Issue #835, opened 2025-10-08
many code paths use asserts and null checks for the same thing at the same time


sleeptightAnsiC:
Nuklear uses following pattern through its source tree:
```
NK_ASSERT(foo);
NK_ASSERT(bar);
if (!foo || !bar) return;
```
(where `foo` and `bar` are pointers / arguments to some function)

example: https://github.com/Immediate-Mode-UI/Nuklear/blob/171090c8e21584330207544dbde5e7c1ec581135/src/nuklear_buffer.c#L38-L41

This is confusing in many places because: (a) there are functions that I would expect to accept NULL, e.g. any destructor, but they crash due to NK_ASSERT; (b) if a code works just fine with a null check, why does it need to be guarded with NK_ASSERT? (c) if I cannot write the code that reaches those null checks then I should probably remove them.

I could point to few locations where it does not make much sense. For example [nk_font_atlas_clear](https://github.com/Immediate-Mode-UI/Nuklear/blob/171090c8e21584330207544dbde5e7c1ec581135/src/nuklear_font.c#L1334) should work when passing NULL or empty atlas, but the assert causes an unexpected crash [[reference]](https://github.com/Immediate-Mode-UI/Nuklear/pull/779#discussion_r2331856071)

Is there some specific reason why Nuklear uses this idiom?

---

For reference, almost every occurrence of this pattern can be found with following AWK script:

<details><summary>Details</summary>
<p>

```sh
$ git clone https://github.com/Immediate-Mode-UI/Nuklear.git
$ cd Nuklear/src

$ awk '                                                                                                                                                                                                                              
function buf_reset() { buf = "" }                                                                                                                                                                                                                              
function buf_append() { buf = buf "\n" FILENAME ":" FNR ":" $0 }                                                                                                                                                                                               
BEGIN { buf_reset() }                                                                                                                                                                                                                                          
/ *NK_ASSERT(.*);/ { buf_append(); next; }                                                                                                                                                                                                                     
/ *if(.*) return;/ { if (buf) { buf_append(); print buf; buf_reset(); } next; }                                                                                                                                                                                
/.*/ { buf_reset(); next; }                                                                                                                                                                                                                                    
' ./*.c
```

```

./nuklear_buffer.c:38:    NK_ASSERT(b);
./nuklear_buffer.c:39:    NK_ASSERT(a);
./nuklear_buffer.c:40:    NK_ASSERT(initial_size);
[...]

PR #878, opened 2026-01-27
Fix crash in nk_buffer_realloc and forbid the use of realloc

### LINKED ISSUE #768: Memory safety (use after free) issue in `nk_buffer_realloc`
## Actual Behavior

The `nk_allocator` interface requires an allocation function which takes a pointer as argument. This suggests that we should use `realloc()` if said pointer is not `NULL` (but the default implementation actually ignores this argument, and simply use `malloc()`).

In the function `nk_buffer_realloc()`, there is a call to the allocation function which passes the buffer memory as argument. But then, if the result of this function is different from the old buffer memory pointer, the old pointer is freed:

https://github.com/Immediate-Mode-UI/Nuklear/blob/master/nuklear.h#L8592-L8595

This is introduce a memory safety issue (use after free). The `realloc()` function might return **the same memory block** at a **different address**:

https://manpages.org/realloc/3

> The `realloc()` function returns a pointer to the newly allocated memory, which is suitably aligned for any built-in type and **may be different from `ptr`**, or `NULL` if the request fails.

This means that if `realloc()` moves the pointer (and it often does), the `nk_buffer_realloc()` will free what it expects to be another allocation, and then the rest of the lib will try to use an allocation that was actually freed.

## Desired Behavior

**Option 1:** You should only check for `NULL` and trust the behavior of `nk_buffer_realloc()` (which would rely on `realloc()` instead of `malloc()`, no more free+memcpy.

**Option 2:** Remove the `ptr` argument of the `nk_plugin_alloc` callback, so that it is 100% clear that `realloc()` is not expected here.

sleeptightAnsiC:
Fixes: https://github.com/Immediate-Mode-UI/Nuklear/issues/768

Please read the linked issue for explanation.

The default implementation of nk_allocator callback (aka nk_malloc) is fundamentally broken, and should have used realloc instead of malloc in the first place. This must have led to the workaround in nk_buffer_realloc that would crash whenever you use an allocator with correct realloc assumption.

We cannot change nk_malloc at this point as people could have implemented their own allocators based on same assumptions, nor we can change the nk_allocator interface because it would be a breaking change (and because we may actually want to use realloc in a future) but we can ensure that Nuklear itself never tries to reallocate memory in its internal code (which is already happening anyway, this only enforces it)

This is kinda ugly, but it still leaves a room for making a breaking change later.

### DIFF
--- a/src/nuklear_buffer.c
+++ b/src/nuklear_buffer.c
@@ -11,7 +11,16 @@ NK_LIB void*
 nk_malloc(nk_handle unused, void *old,nk_size size)
 {
     NK_UNUSED(unused);
-    NK_UNUSED(old);
+    /*
+     * Due to historical reasons, Nuklear always calls alloc(user,old,size)
+     * with the "old" pointer always being NULL. For full explanation, see:
+     * https://github.com/Immediate-Mode-UI/Nuklear/issues/768
+     *
+     * FIXME: The best would be to replace this function with "nk_realloc",
+     * but some custom allocators could still depend on "malloc" assumption
+     * so changing this now (without major bump) would break existing code.
+     */
+    NK_ASSERT(!old && "Nuklear's internal code must never call realloc !");
     return malloc(size);
 }
 NK_LIB void
@@ -104,15 +113,16 @@ nk_buffer_realloc(struct nk_buffer *b, nk_size capacity, nk_size *size)
         return 0;
 
     buffer_size = b->memory.size;
-    temp = b->pool.alloc(b->pool.userdata, b->memory.ptr, capacity);
+    NK_ASSERT(capacity >= buffer_size && "shrinking was never supported here");
+
+    /* HACK: this simulates realloc with malloc+memcpy+free
+     * for backwards compatibility reasons, see the note in nk_malloc */
+    temp = b->pool.alloc(b->pool.userdata, 0, capacity);
     NK_ASSERT(temp);
     if (!temp) return 0;
-
     *size = capacity;
-    if (temp != b->memory.ptr) {
-        NK_MEMCPY(temp, b->memory.ptr, buffer_size);
-        b->pool.free(b->pool.userdata, b->memory.ptr);
-    }
+    NK_MEMCPY(temp, b->memory.ptr, buffer_size);
+    b->pool.free(b->pool.userdata, b->memory.ptr);
 
     if (b->size == buffer_size) {
         /* no back buffer so just set correct size */
--- a/demo/sdl3_renderer/nuklear_sdl3_renderer.h
+++ b/demo/sdl3_renderer/nuklear_sdl3_renderer.h
@@ -112,16 +112,7 @@ NK_INTERN void *
 nk_sdl_alloc(nk_handle user, void *old, nk_size size)
 {
     NK_UNUSED(user);
-    /* FIXME: nk_sdl_alloc should use SDL_realloc here, not SDL_malloc
-     * but this could cause a double-free due to bug within Nuklear, see:
-     * https://github.com/Immediate-Mode-UI/Nuklear/issues/768
-     * */
-#if 0
     return SDL_realloc(old, size);
-#else
-    NK_UNUSED(old);
-    return SDL_malloc(size);
-#endif
 }
 
 NK_INTERN void
--- a/nuklear.h
+++ b/nuklear.h
@@ -8524,7 +8524,16 @@ NK_LIB void*
 nk_malloc(nk_handle unused, void *old,nk_size size)
 {
     NK_UNUSED(unused);
-    NK_UNUSED(old);
+    /*
+     * Due to historical reasons, Nuklear always calls alloc(user,old,size)
+     * with the "old" pointer always being NULL. For full explanation, see:
+     * https://github.com/Immediate-Mode-UI/Nuklear/issues/768
+     *
+     * FIXME: The best would be to replace this function with "nk_realloc",
+     * but some custom allocators could still depend on "malloc" assumption
+     * so changing this now (without major bump) would break existing code.
+     */
+    NK_ASSERT(!old && "Nuklear's internal code must never call realloc !");
     return malloc(size);
 }
 NK_LIB void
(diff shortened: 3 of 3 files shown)


src/nuklear_buffer.c
```diff
@@ -104,15 +113,13 @@ nk_buffer_realloc(struct nk_buffer *b, nk_size capacity, nk_size *size)
         return 0;
 
     buffer_size = b->memory.size;
-    temp = b->pool.alloc(b->pool.userdata, b->memory.ptr, capacity);
+    temp = b->pool.alloc(b->pool.userdata, 0, capacity);
     NK_ASSERT(temp);
```
sleeptightAnsiC, line 121: ~~(not related to this PR)~~

I feel like this assert should be removed, and if it can't be removed then we have some kind of bug here. Nuklear is expected to never crash upon hitting OOM, it should always gracefully return, thus this assert doesn't make sense to me.

EDIT: Unresolving... This smells like a good example of https://github.com/Immediate-Mode-UI/Nuklear/issues/835 ~~(TODO: mention it in that Issue, after this PR gets merged)~~

~~EDIT: I decided to fix this anyway: https://github.com/Immediate-Mode-UI/Nuklear/compare/efefccd7db7ad5c25f8c73b27f07c5d0c57556a1..c43b6ab8a9a15c016152ddf2ca0e68c458ec7bc6~~

EDIT EDIT EDIT: this was reverted, haha https://github.com/Immediate-Mode-UI/Nuklear/pull/878#discussion_r2777006218

src/nuklear_buffer.c
```diff
@@ -104,15 +113,13 @@ nk_buffer_realloc(struct nk_buffer *b, nk_size capacity, nk_size *size)
...
     if (!temp) return 0;
 
     *size = capacity;
-    if (temp != b->memory.ptr) {
-        NK_MEMCPY(temp, b->memory.ptr, buffer_size);
-        b->pool.free(b->pool.userdata, b->memory.ptr);
-    }
+    NK_MEMCPY(temp, b->memory.ptr, buffer_size);
```
sleeptightAnsiC, line 124: ~~(not related to this PR)~~

I think this memcpy will cause buff-overrun, if Nuklear ever tries to shrink the memory with nk_buffer_realloc. I felt tempted to fix it, or to add some kind of assert, but the change would be unrelated to this PR, so I kept the same thing as it was before.

EDIT: Actually, after thinking about this one again: I should just put an assert here, just to be sure it doesn't corrupt any memory.

EDIT: Fixed in: https://github.com/Immediate-Mode-UI/Nuklear/compare/efefccd7db7ad5c25f8c73b27f07c5d0c57556a1..c43b6ab8a9a15c016152ddf2ca0e68c458ec7bc6

src/nuklear_buffer.c
```diff
@@ -11,7 +11,16 @@ NK_LIB void*
...
+    /*
+     * Due to historical reasons, Nuklear always calls alloc(user,old,size)
+     * with the "old" pointer always being NULL. For full explanation, see:
+     * https://github.com/Immediate-Mode-UI/Nuklear/issues/768
+     *
+     * Note that this limitation is only meant for Nuklear's internal code.
+     * If you're wondering what to pass into nk_allocator's "alloc" field,
+     * you can still pass realloc there and it will work just fine.
+     */
+    NK_ASSERT(!old && "Nuklear's internal code must never reallocate existing memory !");
```
RobLoach: When upgrading to this new version, will existing implementations that pass in `old`, even though it's ignored, end up getting failed asserts?

sleeptightAnsiC: Passing anything else than `NULL` to `old` never worked in this function

`nk_malloc` is internal, if any app/backend directly reused it, that was already a bug on its own. If anyone passed non-NULL `old` here, it would result in mem leak and corruption (and crash).

This should never fail for things that already worked in the past, and even if it does, this would not be a false-positive (if someone hits this assert they had a bug in their code already).

@RobLoach Does this make sense? You can find my reasoning about it under https://github.com/Immediate-Mode-UI/Nuklear/issues/768

EDIT: Resolving. We've talked about this in Discord DMs

sleeptightAnsiC:
Changing to DRAFT. I feel I should at least fix https://github.com/Immediate-Mode-UI/Nuklear/pull/878#discussion_r2732124657 before this one gets merged.

Also, I should reword the comment in `nk_malloc` a bit, and add some comment about lack of realloc in `nk_buffer_realloc`. Without this, people might get confused in the future.

EDIT: fixed in https://github.com/Immediate-Mode-UI/Nuklear/compare/efefccd7db7ad5c25f8c73b27f07c5d0c57556a1..c43b6ab8a9a15c016152ddf2ca0e68c458ec7bc6

sleeptightAnsiC:
@RobLoach any chance for getting this reviewed again?

Fixed all the points from my previous comments, and removed the changelog entry (because it's about to be removed in https://github.com/Immediate-Mode-UI/Nuklear/pull/909) I've decided to patch every suspicious line in that function. If `if (temp != b->memory.ptr) {` branch went unnoticed for so long, I don't believe I should really trust the existing code there.

src/nuklear_buffer.c
```diff
@@ -104,15 +113,15 @@ nk_buffer_realloc(struct nk_buffer *b, nk_size capacity, nk_size *size)
...
+    /* HACK: this simulates realloc with malloc+memcpy+free
+     * for backwards compatibility reasons, see the note in nk_malloc */
+    temp = b->pool.alloc(b->pool.userdata, 0, capacity);
+    if (!temp) return 0;
     *size = capacity;
-    if (temp != b->memory.ptr) {
-        NK_MEMCPY(temp, b->memory.ptr, buffer_size);
-        b->pool.free(b->pool.userdata, b->memory.ptr);
-    }
+    NK_ASSERT(buffer_size <= capacity && "FIXME: this code never supported shrinking");
```
sleeptightAnsiC: Should I change this condition to `capacity >= buffer_size` ?
I feel right now it reads [like if master Yoda wrote it](https://en.wikipedia.org/wiki/Yoda_conditions), hah

EDIT: fixed in https://github.com/Immediate-Mode-UI/Nuklear/compare/c43b6ab8a9a15c016152ddf2ca0e68c458ec7bc6..d02f6cb9bdb69145535ee4ad6a22ee9d8a506bcc

src/nuklear_buffer.c
```diff
@@ -104,15 +113,15 @@ nk_buffer_realloc(struct nk_buffer *b, nk_size capacity, nk_size *size)
...
     buffer_size = b->memory.size;
-    temp = b->pool.alloc(b->pool.userdata, b->memory.ptr, capacity);
-    NK_ASSERT(temp);
-    if (!temp) return 0;
+    NK_ASSERT(capacity >= buffer_size && "shrinking was never supported here");
 
+    /* HACK: this simulates realloc with malloc+memcpy+free
+     * for backwards compatibility reasons, see the note in nk_malloc */
+    temp = b->pool.alloc(b->pool.userdata, 0, capacity);
+    if (!temp) return 0;
```
RobLoach, line 122: The original `NK_ASSERT(temp)` is removed here.  Could likely help diagnose if .alloc() worked, rather than a `return 0`?
``` suggestion
    NK_ASSERT(temp);
    if (!temp) return 0;
```

sleeptightAnsiC: Here's comment about this exact assert: https://github.com/Immediate-Mode-UI/Nuklear/pull/878#discussion_r2732101464

This was intentionally removed, but I'm fine with bringing it back.

@RobLoach Just to confirm we understand each other: Shall I revert it?

sleeptightAnsiC:
oops... Seems like you pushed here by mistake?
There is a missmatch with the header. Gimme sec, I'll rebase it.

EDIT: fixed in https://github.com/Immediate-Mode-UI/Nuklear/compare/a6daf69214167c923e8ca54b73cb706eee7700b5..e8a129f5d1e930e54de6a4ff813eb8960cea40b4

PR #879, opened 2026-01-29
Fixing the demo to not provoque ctrl actions with modded keyboards or mobile virtual keyboards

belug23:
While using a modded keyboard with macros on the characters pressing the keys affected by the ctrl modifier without ctrl down, would still trigger it, it was the same issue using the Android virtual keyboard.

Here's an example of what was happening, pressing `a` would erase all text and write the letter `a` instead, pressing `b` would write the letter `b` at the start of the text area and set the cursor in between the first and second character, pressing `e` would add the letter `e` at the end of the text area.

After some code investigation, I found out that calling `nk_input_key` would always increase `ctx->input.keyboard.keys[key].clicked` even if `ctrl` wasn't push, then if it was released while the clicked counter was still equals to 1, it would increase it to 2 triggering the `nk_input_key_pressed` due to the validation of `if ((k->down && k->clicked) || (!k->down && k->clicked >= 2))`.

Only calling `nk_input_key` when `ctrl` is down fixes the problem.

### DIFF
--- a/demo/sdl3_renderer/nuklear_sdl3_renderer.h
+++ b/demo/sdl3_renderer/nuklear_sdl3_renderer.h
@@ -337,6 +337,7 @@ nk_sdl_clipboard_paste(nk_handle usr, struct nk_text_edit *edit)
     }
     SDL_free(text);
 }
+
 NK_INTERN void
 nk_sdl_clipboard_copy(nk_handle usr, const char *text, int len)
 {
--- a/demo/sdl3_renderer/nuklear_sdl3_renderer.h
+++ b/demo/sdl3_renderer/nuklear_sdl3_renderer.h
@@ -471,42 +471,42 @@ nk_sdl_handle_event(struct nk_context* ctx, SDL_Event *evt)
                     case SDL_SCANCODE_PAGEDOWN:  nk_input_key(ctx, NK_KEY_SCROLL_DOWN, down); break;
                     case SDL_SCANCODE_PAGEUP:    nk_input_key(ctx, NK_KEY_SCROLL_UP, down); break;
                     case SDL_SCANCODE_A:
-                        if (ctrl_down){
+                        if ((down && ctrl_down) ^ nk_input_is_key_down(&ctx->input, NK_KEY_TEXT_SELECT_ALL)){
                             nk_input_key(ctx, NK_KEY_TEXT_SELECT_ALL, down);
                         }
                         break;
                     case SDL_SCANCODE_Z:
-                        if (ctrl_down){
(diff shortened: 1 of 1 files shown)


sleeptightAnsiC (changes requested):
I think I do understand the problem, but this does not seem to be the correct solution.

Ater changing the states of those keys to DOWN, how do you make them UP again? With your patch it won't happen after releasing CTRL (?) or maybe the library itself resets them somewhere? this just doesn't seem right.

Note that when I was submitting `demo/sdl3_renderer`, I reused the code from other demos, so this issue is most likely also happening elsewhere. Have you tested other demos?

EDIT: read my comment down below.

~~I might be missing a bigger picture here, but after taking a look at how Nuklear deals with this thing internally, it seems that [nk_textedit_key](https://github.com/Immediate-Mode-UI/Nuklear/blob/b5f67b88889e77acde49622747fd73888488db73/src/nuklear_text_editor.c#L416-L721) should be handling all MOD/SHIFT/CTRL states on its own. The backend should passthrough all those _"special keys"_ without checking weather or not the MOD/SHIFT/CTRL is being held.~~

sleeptightAnsiC:
@belug23 Can you open a separate Issue for this? I do agree we have a bug here, just not with the proposed fix.

It just "_happens to work"_ with this patch, but it looks like the library and the backend are both fighting over who should be handling this, and in the long run it will most likely break again.

sleeptightAnsiC:
(Our conversation moved to https://github.com/Immediate-Mode-UI/Nuklear/issues/882 I'm only responding here because of the referenced comment)

> I might be missing a bigger picture here, but after taking a look at how Nuklear deals with this thing internally, it seems that [nk_textedit_key](https://github.com/Immediate-Mode-UI/Nuklear/blob/b5f67b88889e77acde49622747fd73888488db73/src/nuklear_text_editor.c#L416-L721) should be handling all MOD/SHIFT/CTRL states on its own. The backend should passthrough all those _"special keys"_ without checking weather or not the MOD/SHIFT/CTRL is being held.

Just want to add that my understanding wasn't really correct. That function handles the state of SHIFT, but nothing else. It's just a confusing design of Nuklear, that it implements all SHIFT-related behavior, but forces the backend to implement everything else. Strange...

belug23:
As discussed in #882 I've updated the code to handle the release of the special functions.

sleeptightAnsiC (commented):
Just note that I still don't consider this to be exactly correct, and **it should NOT be merged yet**. There are many nuances mentioned about this issue in https://github.com/Immediate-Mode-UI/Nuklear/issues/882 We decided to update this PR in order to have something to work with.

RobLoach (changes requested):
Thanks for the review. This logic always catches me.

PR #881, opened 2026-01-30
Add tooltip positioning

rswinkle:
It's been bothering me for a while that the tooltip text is always partially covered by the cursor because we position the top left of the tooltip at the cursor position.  This is my first look at fixing that.

This is not how I would really want to do it. Ideally I would break even more existing code by changing nk_tooltip() take the enum and then have nk_tooltip_begin take a nk_vec2 offset.  The position calculation would then be in nk_tooltip() and the user could always call nk_tooltip_begin() directly if they wanted it somewhere arbitrary not covered by the enum options.

Thoughts?

### DIFF
--- a/demo/common/overview.c
+++ b/demo/common/overview.c
@@ -844,7 +844,7 @@ overview(struct nk_context *ctx)
             bounds = nk_widget_bounds(ctx);
             nk_label(ctx, "Hover me for tooltip", NK_TEXT_LEFT);
             if (nk_input_is_mouse_hovering_rect(in, bounds))
-                nk_tooltip(ctx, "This is a tooltip");
+                nk_tooltip_positioned(ctx, "This is a tooltip", NK_BOTTOM_LEFT);
 
             nk_tree_pop(ctx);
         }
--- a/demo/common/overview.c
+++ b/demo/common/overview.c
@@ -842,9 +842,28 @@ overview(struct nk_context *ctx)
             /* tooltip */
             nk_layout_row_static(ctx, 30, 150, 1);
             bounds = nk_widget_bounds(ctx);
-            nk_label(ctx, "Hover me for tooltip", NK_TEXT_LEFT);
-            if (nk_input_is_mouse_hovering_rect(in, bounds))
-                nk_tooltip_positioned(ctx, "This is a tooltip", NK_BOTTOM_LEFT);
+            nk_label(ctx, "Hover for tooltip", NK_TEXT_LEFT);
+            if (nk_input_is_mouse_hovering_rect(in, bounds)) {
+                nk_tooltip(ctx, "This is a default tooltip");
+            }
+            bounds = nk_widget_bounds(ctx);
+            nk_label(ctx, "Chrome-like tooltip", NK_TEXT_LEFT);
+            if (nk_input_is_mouse_hovering_rect(in, bounds)) {
+                struct nk_vec2 offset = { 8, 8 };
+                nk_tooltip_pos_offset(ctx, "Chrome offsets just a bit", NK_TOP_LEFT, offset);
+            }
+            bounds = nk_widget_bounds(ctx);
+            nk_label(ctx, "Gnome-like tooltip", NK_TEXT_LEFT);
+            if (nk_input_is_mouse_hovering_rect(in, bounds)) {
+                struct nk_vec2 offset = { 0, -15 };
+                nk_tooltip_pos_offset(ctx, "Gnome centers with greater y offset", NK_BOTTOM_MIDDLE, offset);
+            }
+            bounds = nk_widget_bounds(ctx);
+            nk_label(ctx, "Bottom left tooltip", NK_TEXT_LEFT);
+            if (nk_input_is_mouse_hovering_rect(in, bounds)) {
+                struct nk_vec2 offset = { 0, 0 };
+                nk_tooltip_pos_offset(ctx, "Bottom left positioning", NK_BOTTOM_LEFT, offset);
+            }
 
             nk_tree_pop(ctx);
         }
--- a/demo/common/overview.c
+++ b/demo/common/overview.c
@@ -850,13 +850,13 @@ overview(struct nk_context *ctx)
             nk_label(ctx, "Hover for Gnome-like tooltip", NK_TEXT_LEFT);
             if (nk_input_is_mouse_hovering_rect(in, bounds)) {
                 struct nk_vec2 offset = { 0, -15 };
-                nk_tooltip_pos_offset(ctx, "Gnome bottom centers plus a -y offset", NK_BOTTOM_CENTER, offset);
+                nk_tooltip_offset(ctx, "Gnome bottom centers plus a -y offset", NK_BOTTOM_CENTER, offset);
             }
             bounds = nk_widget_bounds(ctx);
             nk_label(ctx, "Hover for a bottom left tooltip", NK_TEXT_LEFT);
             if (nk_input_is_mouse_hovering_rect(in, bounds)) {
                 struct nk_vec2 offset = { 0, 0 };
-                nk_tooltip_pos_offset(ctx, "Bottom left positioning", NK_BOTTOM_LEFT, offset);
+                nk_tooltip_offset(ctx, "Bottom left positioning", NK_BOTTOM_LEFT, offset);
             }
             bounds = nk_widget_bounds(ctx);
             nk_label(ctx, "Hover for MAGIC!", NK_TEXT_LEFT);
@@ -866,7 +866,7 @@ overview(struct nk_context *ctx)
                 struct nk_vec2 offset;
                 offset.x = radius * cos(accum_time_seconds * speed);
                 offset.y = radius * sin(accum_time_seconds * speed);
-                nk_tooltip_pos_offset(ctx, "WOW!", NK_MIDDLE_CENTER, offset);
+                nk_tooltip_offset(ctx, "WOW!", NK_MIDDLE_CENTER, offset);
                 accum_time_seconds += (double)(ctx->delta_time_seconds);
             }
 
@@ -902,7 +902,7 @@ overview(struct nk_context *ctx)
                 bounds = nk_widget_bounds(ctx);
                 nk_label(ctx, "Hover for custom tooltip (you can customize it below)", NK_TEXT_LEFT);
                 if (nk_input_is_mouse_hovering_rect(in, bounds)) {
-                    nk_tooltip_pos_offset(ctx, text_buf, cur_pos, offset);
+                    nk_tooltip_offset(ctx, text_buf, cur_pos, offset);
                 }
                 nk_layout_row_dynamic(ctx, 1, 1);
                 nk_rule_horizontal(ctx, nk_white, nk_true);
--- a/demo/common/overview.c
+++ b/demo/common/overview.c
@@ -839,30 +839,86 @@ overview(struct nk_context *ctx)
                 } else popup_active = nk_false;
             }
 
-            /* tooltip */
-            nk_layout_row_static(ctx, 30, 150, 1);
+            /* tooltips */
+            nk_layout_row_static(ctx, 30, 400, 1);
             bounds = nk_widget_bounds(ctx);
-            nk_label(ctx, "Hover for tooltip", NK_TEXT_LEFT);
+            nk_label(ctx, "Hover for default tooltip", NK_TEXT_LEFT);
             if (nk_input_is_mouse_hovering_rect(in, bounds)) {
                 nk_tooltip(ctx, "This is a default tooltip");
             }
             bounds = nk_widget_bounds(ctx);
-            nk_label(ctx, "Chrome-like tooltip", NK_TEXT_LEFT);
-            if (nk_input_is_mouse_hovering_rect(in, bounds)) {
-                struct nk_vec2 offset = { 8, 8 };
-                nk_tooltip_pos_offset(ctx, "Chrome offsets just a bit", NK_TOP_LEFT, offset);
-            }
-            bounds = nk_widget_bounds(ctx);
-            nk_label(ctx, "Gnome-like tooltip", NK_TEXT_LEFT);
+            nk_label(ctx, "Hover for Gnome-like tooltip", NK_TEXT_LEFT);
             if (nk_input_is_mouse_hovering_rect(in, bounds)) {
                 struct nk_vec2 offset = { 0, -15 };
-                nk_tooltip_pos_offset(ctx, "Gnome centers with greater y offset", NK_BOTTOM_MIDDLE, offset);
+                nk_tooltip_pos_offset(ctx, "Gnome bottom centers plus a -y offset", NK_BOTTOM_CENTER, offset);
             }
             bounds = nk_widget_bounds(ctx);
-            nk_label(ctx, "Bottom left tooltip", NK_TEXT_LEFT);
+            nk_label(ctx, "Hover for a bottom left tooltip", NK_TEXT_LEFT);
             if (nk_input_is_mouse_hovering_rect(in, bounds)) {
                 struct nk_vec2 offset = { 0, 0 };
                 nk_tooltip_pos_offset(ctx, "Bottom left positioning", NK_BOTTOM_LEFT, offset);
+            }
+            bounds = nk_widget_bounds(ctx);
+            nk_label(ctx, "Hover for MAGIC!", NK_TEXT_LEFT);
+            if (nk_input_is_mouse_hovering_rect(in, bounds)) {
+                static double accum_time_seconds = 0.0;
+                const double speed = 3.0, radius = 50.0;
+                struct nk_vec2 offset;
+                offset.x = radius * cos(accum_time_seconds * speed);
+                offset.y = radius * sin(accum_time_seconds * speed);
+                nk_tooltip_pos_offset(ctx, "WOW!", NK_MIDDLE_CENTER, offset);
+                accum_time_seconds += (double)(ctx->delta_time_seconds);
+            }
+
+            /* editor for custom tooltip */
+            {
+                static char text_buf[64] = {0};
+                static int text_len = 0;
+                static int text_initialized = 0;
+                static struct nk_vec2 offset = {0};
+                static const char* tooltip_positions[] =
+                {
+                    "TOP_LEFT",
+                    "TOP_CENTER",
+                    "TOP_RIGHT",
+
+                    "MIDDLE_LEFT",
+                    "MIDDLE_CENTER",
+                    "MIDDLE_RIGHT",
+
+                    "BOTTOM_LEFT",
+                    "BOTTOM_CENTER",
+                    "BOTTOM_RIGHT"
+                };
+                static int cur_pos = NK_TOP_LEFT;
+
+                if (!text_initialized) {
+                    const char text_default[] = "you can customize this!";
+                    NK_ASSERT(sizeof(text_default) < sizeof(text_buf));
+                    memcpy(text_buf, text_default, sizeof(text_default));
+                    text_len = sizeof(text_default) - 1;
+                    text_initialized = 1;
+                }
+                bounds = nk_widget_bounds(ctx);
+                nk_label(ctx, "Hover for custom tooltip (you can customize it below)", NK_TEXT_LEFT);
+                if (nk_input_is_mouse_hovering_rect(in, bounds)) {
+                    nk_tooltip_pos_offset(ctx, text_buf, cur_pos, offset);
+                }
+                nk_layout_row_dynamic(ctx, 1, 1);
+                nk_rule_horizontal(ctx, nk_white, nk_true);
+                nk_layout_row_dynamic(ctx, 30, 2);
+                nk_label(ctx, "custom tooltip text:", NK_TEXT_LEFT);
+                nk_edit_string(ctx, NK_EDIT_FIELD, text_buf, &text_len, sizeof(text_buf), nk_filter_default);
+                text_buf[text_len] = '\0';  /* TODO: why nk_edit_string is NOT setting this on its own? */
+                nk_layout_row_dynamic(ctx, 30, 1);
+                cur_pos = nk_combo(ctx, tooltip_positions, NK_LEN(tooltip_positions), cur_pos, 25, nk_vec2(200, 200));
+
+
+                nk_layout_row_dynamic(ctx, 30, 2);
+                nk_label(ctx, "custom tooltip offset", NK_TEXT_LEFT);
+                nk_property_float(ctx, "x", -100.0f, &offset.x, 100.0f, 5.0f, 0.5f);
+                nk_label(ctx, "custom tooltip offset", NK_TEXT_LEFT);
+                nk_property_float(ctx, "y", -100.0f, &offset.y, 100.0f, 5.0f, 0.5f);
             }
 
             nk_tree_pop(ctx);
--- a/src/nuklear.h
+++ b/src/nuklear.h
@@ -292,6 +292,7 @@ enum nk_color_format    {NK_RGB, NK_RGBA};
 enum nk_popup_type      {NK_POPUP_STATIC, NK_POPUP_DYNAMIC};
 enum nk_layout_format   {NK_DYNAMIC, NK_STATIC};
 enum nk_tree_type       {NK_TREE_NODE, NK_TREE_TAB};
+enum nk_tooltip_pos     {NK_TOP_LEFT, NK_BOTTOM_LEFT, NK_TOP_RIGHT, NK_BOTTOM_RIGHT};
 
 typedef void*(*nk_plugin_alloc)(nk_handle, void *old, nk_size);
 typedef void (*nk_plugin_free)(nk_handle, void *old);
@@ -3604,11 +3605,12 @@ NK_API void nk_contextual_end(struct nk_context*);
  *
  * ============================================================================= */
 NK_API void nk_tooltip(struct nk_context*, const char*);
+NK_API void nk_tooltip_positioned(struct nk_context*, const char*, enum nk_tooltip_pos);
 #ifdef NK_INCLUDE_STANDARD_VARARGS
 NK_API void nk_tooltipf(struct nk_context*, NK_PRINTF_FORMAT_STRING const char*, ...) NK_PRINTF_VARARG_FUNC(2);
 NK_API void nk_tooltipfv(struct nk_context*, NK_PRINTF_FORMAT_STRING const char*, va_list) NK_PRINTF_VALIST_FUNC(2);
 #endif
-NK_API nk_bool nk_tooltip_begin(struct nk_context*, float width);
+NK_API nk_bool nk_tooltip_begin(struct nk_context*, float width, enum nk_tooltip_pos);
 NK_API void nk_tooltip_end(struct nk_context*);
 /* =============================================================================
  *
--- a/src/nuklear.h
+++ b/src/nuklear.h
@@ -292,7 +292,7 @@ enum nk_color_format    {NK_RGB, NK_RGBA};
 enum nk_popup_type      {NK_POPUP_STATIC, NK_POPUP_DYNAMIC};
 enum nk_layout_format   {NK_DYNAMIC, NK_STATIC};
 enum nk_tree_type       {NK_TREE_NODE, NK_TREE_TAB};
-enum nk_tooltip_pos     {NK_TOP_LEFT, NK_BOTTOM_LEFT, NK_TOP_RIGHT, NK_BOTTOM_RIGHT};
+enum nk_tooltip_pos     {NK_TOP_LEFT, NK_BOTTOM_LEFT, NK_TOP_RIGHT, NK_BOTTOM_RIGHT, NK_TOP_MIDDLE, NK_BOTTOM_MIDDLE};
 
 typedef void*(*nk_plugin_alloc)(nk_handle, void *old, nk_size);
 typedef void (*nk_plugin_free)(nk_handle, void *old);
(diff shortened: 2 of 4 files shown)


src/nuklear.h
```diff
@@ -283,6 +283,7 @@ enum nk_color_format    {NK_RGB, NK_RGBA};
 enum nk_popup_type      {NK_POPUP_STATIC, NK_POPUP_DYNAMIC};
 enum nk_layout_format   {NK_DYNAMIC, NK_STATIC};
 enum nk_tree_type       {NK_TREE_NODE, NK_TREE_TAB};
+enum nk_tooltip_pos     {NK_TOP_LEFT, NK_BOTTOM_LEFT, NK_TOP_RIGHT, NK_BOTTOM_RIGHT};
```
sleeptightAnsiC: ~Do you think this abstraction will be enough?~ I guess, you don't like it neither, according to your initial message.

I have a feeling that sooner or later someone will need to position the cursor at the edges of tooltip (instead of the corners) and we will have to redo/extend this thing again. Same with how far it should be from the cursor.

It also would be really nice for the tooltip to stay in the available screen space. Right now it just draws outside so you can't see its contents [screenshot below], however, I'm not sure if this problem is solvable right now (like does Nuklear keep the OS Window size somewhere?)

Image

src/nuklear.h
```diff
@@ -3595,11 +3596,12 @@ NK_API void nk_contextual_end(struct nk_context*);
  *
  * ============================================================================= */
 NK_API void nk_tooltip(struct nk_context*, const char*);
+NK_API void nk_tooltip_positioned(struct nk_context*, const char*, enum nk_tooltip_pos);
 #ifdef NK_INCLUDE_STANDARD_VARARGS
 NK_API void nk_tooltipf(struct nk_context*, NK_PRINTF_FORMAT_STRING const char*, ...) NK_PRINTF_VARARG_FUNC(2);
 NK_API void nk_tooltipfv(struct nk_context*, NK_PRINTF_FORMAT_STRING const char*, va_list) NK_PRINTF_VALIST_FUNC(2);
 #endif
-NK_API nk_bool nk_tooltip_begin(struct nk_context*, float width);
+NK_API nk_bool nk_tooltip_begin(struct nk_context*, float width, enum nk_tooltip_pos);
```
sleeptightAnsiC: I'm against breaking public API.

Can't this new function simply be `nk_tooltip_begin_positioned()` ?
You could call the new one from the old one, and keep the old signature unchanged.

I know that a lot of people are statically linking, but in this case it would break the existing code and existing bindings.

sleeptightAnsiC (changes requested):
EDIT: Hiding this. We already talked through all of this.

~~Just wrote few of my thought here and there. In general, I'm against breaking the old API and that's my biggest problem with this PR right now. I do agree that the current behavior of tooltip is pretty bad though.~~ EDIT: this was later fixed

@rswinkle btw, would you mind taking a look at few recent PRs. These are awaiting approval from other contributor and Rob would merge them otherwise. (Examples: https://github.com/Immediate-Mode-UI/Nuklear/pull/841 , https://github.com/Immediate-Mode-UI/Nuklear/pull/862 )

src/nuklear.h
rswinkle: I could but it just feels like a lot of duplication. And I don't really understand the resistance to breaking the API here because as you say, not only are most statically linking, but even those who aren't are not depending on a system installed dynamic library, they're handling it internally/shipping with it anyway so it still doesn't affect them. Same with bindings. If we add new public functions, the bindings are out out of date and have to be updated anyway, and people using old versions are unaffected.

Am I missing something there? I'm fine to try to avoid "breaking" it if people insist but I just don't really see the point for a single header library like Nuklear that's designed to be vendored one way or another.

src/nuklear.h
rswinkle: Yes, I do think the providing an offset is the only reasonable final form even if the enums are kept somehow as a shortcut/convenience for a simpler API. Again that grows the API even more, especially if we insist on not breaking the exsting API.

I've thought about the offscreen issue too. I'm sure there's a way but whether it's pretty or backend specific or not is the question. I'll look into it after I finish the rest of it.

rswinkle:
I can take a look at those two other PRs when I get a chance. Sorry I must have missed you tagging me back in November or just forgotten about it since that was a busy time.

src/nuklear.h
sleeptightAnsiC: (I know that you force-pushed and have seen the new changes, just want to continue our discussion here)

How about, instead of specifying all possible states up front in the enum (TOP, LEFT, TOP_LEFT, TOP_RIGHT, and so on...), we just use a bitmask flags? So in case you want the tooltip aligned to the edge you would use `LEFT`, and for the corner it would be `TOP|LEFT`.

And later when we attempt to fix the offscreen issue, we would just add a new bitmask flag (e.g. OFFSCREEN_WRAP). This would even allow us to make more changes later by simply adding a new flag. IMO bitmasks are quite nice to work with when dealing with problems that we are facing here.

Personally, I think having two calls: default one without arguments, and custom one with bitmask and x/y offsets - would be the sweat spot. We will try to enforce the best possible behavior in the default call, and if someone doesn't like our defaults, the other call should be enough. If this still wouldn't be enough, that will most likely be the case for custom widget (we can't please everyone anyway)

Btw, I'm planning to fork this branch and implement the ideas that I described above (not as PR, I will give you a raw diff or a link to another branch later). Just let me know what do you think about this, and I will attemt to cook something good, either today or in few days :)

src/nuklear.h
sleeptightAnsiC: I see that you fixed it with recent push so fell free to resolve this thread. Thanks!

About why breaking the API is _"bad"_, it's just a matter of _"don't do unto others what you don't want done unto you"_. In this case, you were making those breaking changes, so you were in charge about how and when it will happen, but there are many who simply don't want do deal with this for one reason or another. I'm pretty sure you've also been in the same situation.

It really _sucks_ that someone have to patch their code just in order to update a library, and the argument that _"the most are vendoring and statically linking"_ is not correct, or at least not for me IMO. Statically linking makes it easier to deal with, but not seamless. You would still have to patch every place in the code where said call was used. The fact that we have to vendor it is a symptom, not solution. After updating the library, you should only test it for regression and move on. Having to deal with incompatible API is unnecessary friction and always introduces some problems.

For bindings this is even worse. Yes, people making those binding are fine with regenerating them, but in many cases the API breakage goes unnoticed (in case of automatic bindings) and only breaks someone's code down the road. That "someone" goes to the bindings' author and complains. In practice, this almost always ends up in fork where author implements their own compatibility patches and this creates friction.

We really have no technical reason to break the API here, and even if we had any, I would always advice against it. I hope this clarifies my point. If you still disagree or don't understand something, feel free to ask.

src/nuklear.h
rswinkle: Using a bitmask occurred to me but I rejected it because I feel it would be more unwieldy, less flexible and less safe. Less performant too since you can't use a switch.

For example, while the number of characters is only slightly more for a call, it is far easier to make a mistake/typo with bitmasks and you can do illogical things with bitmasks that you can't do with simple enums (for example TOP | BOTTOM or TOP|MIDDLE|LEFT). Also I immediately thought about how to handle defaults and the different ways of doing it and decided that doesn't really help.* Even with my current version I could simplify especially if I didn't care about compiler warnings but compared to the different ways you could do bitmasks and what the code would/could look like I think my code is already much better/simpler.

> And later when we attempt to fix the offscreen issue, we would just add a new bitmask flag (e.g. OFFSCREEN_WRAP). This would even allow us to make more changes later by simply adding a new flag. IMO bitmasks are quite nice to work with when dealing with problems that we are facing here.

I looked into this just a bit and it doesn't seem like the window dimensions are stored anywhere in Nuklear. I feel like the solution to that problem should be its own PR and it has to be added to the context and updated in the event handlers which means updating *all* the backends. Or at the very least the user has to optionally update it when they handle window resizes.

Also OFFSCREEN_WRAP implies wrapping which I don't think is the right approach. Wrapping doesn't help if you're right up against the edge on the same side the tooltip is supposed to be. We really need intelligent positioning (ie check if the cursor is too close to an edge and if so we ensure that the positioning is such that the entire tooltip is visilible which obviously is always possible unless the entire window is so small it can't fit it at all. The problem is I designing this and how it would interact with the user positioning if at all.

>Personally, I think having two calls: default one without arguments, and custom one with bitmask and x/y offsets - would be the sweat spot.

I'm not sure I follow? You want a bitmask/position *and* offset parameters in the same call?  How does that work? It's either redundant or contradictory. I could see doing the offsets as simple int x,y parameters though instead of an nk_vec2 but I figured the latter fit more with nuklear's style.

>Btw, I'm planning to fork this branch and implement the ideas that I described above 

Have at it, I'm curious to see how you approach it, and maybe I'm just not understanding what you're thinking.


*For example normally 0 for a bitmask would be the default so that would might be equivalent to TOP | LEFT but of course they need actual values or you wouldn't be able to use them and I also thought about "overloading" the bits so both 0 and 1 for a bit mean for instance TOP and BOTTOM so you'd really only specify BOTTOM and/or RIGHT because leaving them 0 implies the other ... but then how do you handle MIDDLE because I really think that should be an option too?  Like I said I just think enums make things simpler and more straight forward.

sleeptightAnsiC: Yea, I'll try to implement it soon (with playable demo maybe). I already have some implementation details in mind, which I think would be really confusing to explain here, so I get why you disagree on the overall idea.

When it comes to whether it will be more/less flexible/safe/performant, well you will see for yourself, but I don't think it will be much of a difference. You can already break this code with wrong input (kinda). It does fallback to sane value (00 in your case) and I had something similar in mind. I don't believe you will see any performance difference really (compiler in O2/O3 will probably already turn it into magic because it all appears in one function scope).

That thing I suggested with OFFSCREEN_WRAP was just a wild guess. I agree we should leave it for later.

> > Personally, I think having two calls: default one without arguments, and custom one with bitmask and x/y offsets - would be the sweat spot.
> 
> I'm not sure I follow? You want a bitmask/position _and_ offset parameters in the same call? How does that work? It's either redundant or contradictory.

You newer know how big the tooltip will end up being, so you specify what edge/corner point it should start at with bitmask, and then the offset to that point with vec2. So for example: `tooltip(LEFT|UP, vec2(0, 0))` would get you the same behavior as current master branch, and `tooltip(DOWN, vec2(0, -10))` would spawn it 10 pixels above your cursor, and for the exact position you could use `tooltip(0, vec2(X, Y))` (but this last one probably wouldn't make much sense as you don't know the tooltip size anyway)

src/nuklear.h
rswinkle: > About why breaking the API is _"bad"_, it's just a matter of _"don't do unto others what you don't want done unto you"_. In this case, you were making those breaking changes, so you were in charge about how and when it will happen, but there are many who simply don't want do deal with this for one reason or another. I'm pretty sure you've also been in the same situation.

Not really. Perhaps it's just the nature of the type of development I do but it has never been an issue for me.

> It really _sucks_ that someone have to patch their code just in order to update a library, and the argument that _"the most are vendoring and statically linking"_ is not correct, or at least not for me IMO. Statically linking makes it easier to deal with, but not seamless. You would still have to patch every place in the code where said call was used. The fact that we have to vendor it is a symptom, not solution. After updating the library, you should only test it for regression and move on. Having to deal with incompatible API is unnecessary friction and always introduces some problems.

I think we have some miscommunication here. First, as I said, vendoring is a superset of statically linking; ie if you compile it yourself, even if you compile it to a dynamic library and have to ship the library with your executable, that's still vendoring. You completely control the version you use. If I want to update a library, I'm already having to copy in the new version and recompile, whether it breaks API or not. And if I'm doing it for new features which is the usual case, then I'm going to be making changes to the code anyway, to make use of those features. Even if I'm not, the fact that I'm bothering to update the libraries means it's still in active development and making changes is kind of the the definition of active development so I don't really see much difference.

The only time where I really understand the "No breaking API" mantra is for system APIs and system installed libraries. For instance the Linux kernel API, POSIX/syscalls etc. (I completely agree with Torvalds "We do not break userspace") or libc or the GTK or QT libraries installed through the package manager. Even in the latter case, they still can break API they just do it very carefully and use versioning correctly and allow multiple versions to exist on the system simultaneously.

> For bindings this is even worse. Yes, people making those binding are fine with regenerating them, but in many cases the API breakage goes unnoticed (in case of automatic bindings) and only breaks someone's code down the road. That "someone" goes to the bindings' author and complains. In practice, this almost always ends up in fork where author implements their own compatibility patches and this creates friction.

This is really a continuation of above paragraph but thought I'd respond here because it applies. Versioning, especially semantic versioning, exists for a reason. And when something is vendored, the developer has the choice whether to update or not and the change in version (and whether it's breaking) is part of that decision. In most cases, there's no reason to update, especially not for every version change. Unless there's a serious security issue which applies to their software with that library (unlikely), serious performance improvements that would make a noticeable difference (also rare), or there are new features in the new version that they want to take advantage of (which would again also require changing the code to use), there's no real reason to update. And if one or more of those reasons do apply, then having to fix a few breaking changes is a reasonable price to pay for the improvements and not a big deal in the scheme of things (especially for larger more complicated libraries; having to make changes might feel more significant when "updating the library" consists of downloading/copying a single file for a single file lib and it doesn't affect your build process like having a giant external dependency like say SDL does, but it's still not much).

> We really have no technical reason to break the API here, and even if we had any, I would always advice against it. I hope this clarifies my point. If you still disagree or don't understand something, feel free to ask.

Overall, my opinion is that fear of breaking changes more often holds libraries back and does more harm than good for 95% of cases, just causing bloat and stagnancy and awkward APIs.

SQLite is a good exception than proves the rule. It has a couple of [quirks](https://www.sqlite.org/quirks.html) that go back to bugs/decisions at the beginning but it blew up so much/quickly and was so useful to so many that they decided to leave things the way they are and work around things where possible even across major version updates. I still am not 100% on that but I understand it in their case and they did such a good job in their initial/early design and in general that there aren't many real issues despite having no major updates for 20+ years.

src/nuklear.h
rswinkle: >You newer know how big the tooltip will end up being, so you specify what edge/corner point it should start at with bitmask, and then the offset to that point with vec2. So for example: tooltip(LEFT|UP, vec2(0, 0)) would get you the same behavior as current master branch, and tooltip(DOWN, vec2(0, -10)) would spawn it 10 pixels above your cursor, and for the exact position you could use tooltip(0, vec2(X, Y)) (but this last one probably wouldn't make much sense as you don't know the tooltip size anyway)

Hmmm. I guess I can see what you're going for, if by "exact position" you mean relative to current top/left behavior, but that's why I mostly considered the offset call to be an internal call anyway. Though it's still useful for some behaviors, like some tooltips spawn just a bit below and to the right of the cursor to avoid the exact problem that made me create this PR. You don't need to know the size of the tooltip to get it out from behind a standard cursor but still be roughly top/left (chromium based browsers do this).  I do think you should add MIDDLE as some applications (GNOME apps for instance) center tooltips.

src/nuklear.h
sleeptightAnsiC: I think our disagreement is becoming _"political"_ rather than _"technical"_. I'm against breaking changes at all cost, and you're fine with breakage, and I don't believe either of us will convince one another. Apologies for not responding to your points. I did fully read your comment, I just don't feel like we should continue discussing this, especially since this is already solved in this PR

Just for reference, changing this one signature would break all of this publicly available code: https://github.com/search?q="nk_tooltip_begin("+-path%3Anuklear.h+-path%3Anuklear_tooltip.c&type=code
Many of those are bindings indeed, but there is quite a lot of hand-written code too. For me this is just unjustified, no matter the argument. EDIT: btw, we can probably read some of those in order to figure out the pattern that people are trying to use tooltips with.

rswinkle: I wouldn't call it political, more philosophical but system/API design is still technical so what's happening is that technical decisions are made based on philosophical beliefs.  My view boils down to:

If it breaks without the developer doing/changing anything, it's the library/system's fault. If it breaks when the developer changes something (ie updates the library between major versions), that's not bad and in fact expected. Giving people the *option* to trade a little effort for a better library will never seem like a bad idea to me.

I think more people are coming to my view in some ways in the last decade. I read a post somewhere in the last year or two by someone explaining that they purposely keeps their projects/libraries at major version 0 so they can keep breaking things indefinitely and avoid dealing with the false idea that major version numbers have to represent some large radical change and that high major numbers feel wrong and "scary" to users so it should rarely change. Many developers responded positively to that idea for the exact reasons I've discussed.

>Just for reference, changing this one signature would break all of this publicly available code:
Only if they update the Nuklear, which most will never do. Also most of those look like bindings/backend code not application code. I think most people just use nk_tooltip() not nk_tooltip_begin().

Nuklear hasn't had a major update in 8 years. Who knows how many sub-optimal decisions have been made in that time in the name of not breaking the API. Nuklear is very very far from being a widely used library in the scheme of things.

EDIT: typos

rswinkle: PS. Honestly I like Nuklear but I've thought about seriously forking it myself for a couple years now because I don't want to worry about making breaking changes (API or graphically/behaviorally) and half the reason I don't is that I'm just waiting for [ARC](https://www.youtube.com/watch?v=UULQyRKc6sM) to come out so I can see whether I should just switch to that instead as my goto C imgui.

RobLoach:
This has been an annoyance for me forever. Never thought of fixing it haha. Nice work.

EDIT: Regarding breaking changes, since it seems like one parameter, it's not that big of a breaking change. Yes, a breaking change no less, but easily caught.

rswinkle:
> This has been an annoyance for me forever. Never thought of fixing it haha. Nice work.

Thanks

> EDIT: Regarding breaking changes, since it seems like one parameter, it's not that big of a breaking change. Yes, a breaking change no less, but easily caught.

Well I just pushed a better variant based on @sleeptightAnsiC's idea combining the parameters. I'm not breaking the API currently and with the simplification I don't feel as much of a need to.  Really all I would do now is get rid of `nk_tooltip_begin()` entirely which probably isn't worth it.

If there aren't any issues with the design now, I can add the `nk_tooltipf_pos_offset()` variant as well and call it good.

RobLoach (approved):
Looks great to me, but will let others give another +1. Thjnaks for the clean up.

demo/common/overview.c
```diff
@@ -844,7 +844,7 @@ overview(struct nk_context *ctx)
             bounds = nk_widget_bounds(ctx);
             nk_label(ctx, "Hover me for tooltip", NK_TEXT_LEFT);
             if (nk_input_is_mouse_hovering_rect(in, bounds))
-                nk_tooltip(ctx, "This is a tooltip");
+                nk_tooltip_positioned(ctx, "This is a tooltip", NK_BOTTOM_LEFT);
```
sleeptightAnsiC: Seems like you forgot to change this one during latest push?

CI is failing with:
```
cc main.c -std=c89 -Wall -Wextra -pedantic -Wno-unused-function -o bin/demo -lallegro -lallegro_main -lallegro_image -lallegro_font -lallegro_ttf -lallegro_primitives -lm
In file included from main.c:67:
../../demo/common/overview.c: In function ‘overview’:
../../demo/common/overview.c:847:17: warning: implicit declaration of function ‘nk_tooltip_positioned’; did you mean ‘nk_tooltip_pos_offset’? [-Wimplicit-function-declaration]
  847 |                 nk_tooltip_positioned(ctx, "This is a tooltip", NK_BOTTOM_LEFT);
      |                 ^~~~~~~~~~~~~~~~~~~~~
      |                 nk_tooltip_pos_offset
/usr/bin/ld: /tmp/cckClT8R.o: in function `overview':
main.c:(.text+0x490dc): undefined reference to `nk_tooltip_positioned'
collect2: error: ld returned 1 exit status
```

sleeptightAnsiC:
> Well I just pushed a better variant based on @sleeptightAnsiC's idea combining the parameters. I'm not breaking the API currently and with the simplification I don't feel as much of a need to.

Thanks for this one. I hope I will create my alternative version with the bitmask soon. Sorry for not doing it already.

> Really all I would do now is get rid of nk_tooltip_begin() entirely which probably isn't worth it.

I'm pretty sure that this call is beneficial for people who want to have good enough tooltips with other things than pain text (e.g. image or plotting). Could we please keep this call in the future?

sleeptightAnsiC:
EDIT: https://github.com/Immediate-Mode-UI/Nuklear/pull/900

@RobLoach Just a reminder in case you didn't read the whole conversation: 

> Btw, I'm planning to fork this branch and implement the ideas that I described above
>
> > Have at it, I'm curious to see how you approach it, and maybe I'm just not understanding what you're thinking.
> >
> > > Well I just pushed a better variant based on sleeptightAnsiC's idea combining the parameters.
> > >
> > > > Thanks for this one. I hope I will create my alternative version with the bitmask soon. Sorry for not doing it already.

### Please, wait with merging until I submit that other version so we can discuss it first.

RobLoach (changes requested):
Sounds good, I'll mark this as such till then.

demo/common/overview.c
rswinkle: Oops, must have accidentally committed a change to it in an earlier commit and then remembered to leave it out in the latest. I'll put it back for now but we may want to add more tooltip examples to demonstrate the different uses/behavior eventually.

EDIT: Actually, I might as well just add the examples myself.  Easier for everyone to test it.

src/nuklear.h
RobLoach: @rswinkle I honestly believe we should reconsider the strict stance on breaking changes. It hinders innovation, and blocks new contributors. Made an issue to discuss that: https://github.com/Immediate-Mode-UI/Nuklear/issues/887

ARC looks pretty interesting. Curious to check it out once they open source it. [Clay](https://github.com/nicbarker/clay) looks nice too. I just love the simplicity of Nuklear though.

rswinkle: > @rswinkle I honestly believe we should reconsider the strict stance on breaking changes. It hinders innovation, and blocks new contributors. Made an issue to discuss that: #887

Will definitely chime in there.

> ARC looks pretty interesting. Curious to check it out once they open source it. [Clay](https://github.com/nicbarker/clay) looks nice too. I just love the simplicity of Nuklear though.

Yeah, it seems pretty nice, fast and small (4K LOC I think he mentioned?). Hopefully it comes out soon.

I've been meaning to try Clay since it came out but I haven't partly because I'm already using Nuklear for my active projects and I'm invested in it (I *really* don't want to have to rewrite a file browser GUI), and also I'm not 100% sold on Clay's declarative syntax but I would need to give it a real try to know for sure.

src/nuklear.h
```diff
@@ -283,6 +283,7 @@ enum nk_color_format    {NK_RGB, NK_RGBA};
 enum nk_popup_type      {NK_POPUP_STATIC, NK_POPUP_DYNAMIC};
 enum nk_layout_format   {NK_DYNAMIC, NK_STATIC};
 enum nk_tree_type       {NK_TREE_NODE, NK_TREE_TAB};
+enum nk_tooltip_pos     {NK_TOP_LEFT, NK_BOTTOM_LEFT, NK_TOP_RIGHT, NK_BOTTOM_RIGHT, NK_TOP_MIDDLE, NK_BOTTOM_MIDDLE};
```
sleeptightAnsiC: Note: Those symbols will probably need a prefix. Something like `NK_TOP_LEFT` feels too generic and ambiguous for tooltip-specifc enum to reserve, right? I planned to use `NK_TOOLTIP_*` prefix for my alternative implementation. Just wanted to mention it here in case you may have some thoughts about it.

(sorry, not sure how I missed this one the first time I reviewed it)

sleeptightAnsiC (changes requested):

src/nuklear.h
rswinkle: I was going to say I think that's kind of long and we might want to use these elsewhere, then I looked and saw we already have these:

```
enum nk_widget_align {
    NK_WIDGET_ALIGN_LEFT        = 0x01,
    NK_WIDGET_ALIGN_CENTERED    = 0x02,
    NK_WIDGET_ALIGN_RIGHT       = 0x04,
    NK_WIDGET_ALIGN_TOP         = 0x08,
    NK_WIDGET_ALIGN_MIDDLE      = 0x10,
    NK_WIDGET_ALIGN_BOTTOM      = 0x20
};
enum nk_widget_alignment {
    NK_WIDGET_LEFT        = NK_WIDGET_ALIGN_MIDDLE|NK_WIDGET_ALIGN_LEFT,
    NK_WIDGET_CENTERED    = NK_WIDGET_ALIGN_MIDDLE|NK_WIDGET_ALIGN_CENTERED,
    NK_WIDGET_RIGHT       = NK_WIDGET_ALIGN_MIDDLE|NK_WIDGET_ALIGN_RIGHT
};

enum nk_text_align {
    NK_TEXT_ALIGN_LEFT        = 0x01,
    NK_TEXT_ALIGN_CENTERED    = 0x02,
    NK_TEXT_ALIGN_RIGHT       = 0x04,
    NK_TEXT_ALIGN_TOP         = 0x08,
    NK_TEXT_ALIGN_MIDDLE      = 0x10,
    NK_TEXT_ALIGN_BOTTOM      = 0x20
};
enum nk_text_alignment {
    NK_TEXT_LEFT        = NK_TEXT_ALIGN_MIDDLE|NK_TEXT_ALIGN_LEFT,
    NK_TEXT_CENTERED    = NK_TEXT_ALIGN_MIDDLE|NK_TEXT_ALIGN_CENTERED,
    NK_TEXT_RIGHT       = NK_TEXT_ALIGN_MIDDLE|NK_TEXT_ALIGN_RIGHT
};

// Also this, used for drawing triangles and indicating knob zero direction. Note the lack of prefix
enum nk_heading         {NK_UP, NK_RIGHT, NK_DOWN, NK_LEFT};

```

Which gives me another perfect example of something that I would immediately change, breaking the API once again because deleting code an increasing simplicity is worth it. There's no reason to have 2 identical sets of constants used for the same exact purpose, let alone 3 or more if you're adding another one.

We should just have one set used for alignment everywhere, whether it's a widget, text, or tooltips or anything else. And look we can even compromise on the bitmask vs enum thing:

```
enum nk_align {
    NK_ALIGN_LEFT        = 0x01,
    NK_ALIGN_CENTERED    = 0x02,
    NK_ALIGN_RIGHT       = 0x04,
    NK_ALIGN_TOP         = 0x08,
    NK_ALIGN_MIDDLE      = 0x10,
    NK_ALIGN_BOTTOM      = 0x20
};
enum nk_alignment {
    NK_TOP_LEFT         = NK_ALIGN_TOP|NK_ALIGN_LEFT,
    NK_TOP_CENTERED     = NK_ALIGN_TOP|NK_ALIGN_CENTERED,
    NK_TOP_RIGHT        = NK_ALIGN_TOP|NK_ALIGN_RIGHT,

    NK_MIDDLE_LEFT      = NK_ALIGN_MIDDLE|NK_ALIGN_LEFT,
    NK_MIDDLE_CENTERED  = NK_ALIGN_MIDDLE|NK_ALIGN_CENTERED,
    NK_MIDDLE_RIGHT     = NK_ALIGN_MIDDLE|NK_ALIGN_RIGHT

    NK_BOTTOM_LEFT      = NK_ALIGN_BOTTOM|NK_ALIGN_LEFT,
    NK_BOTTOM_CENTERED  = NK_ALIGN_BOTTOM|NK_ALIGN_CENTERED,
    NK_BOTTOM_RIGHT     = NK_ALIGN_BOTTOM|NK_ALIGN_RIGHT
};
```

I still don't really see the point of the bitmasks because no one would ever choose to type them manually both because it's longer and more error prone, and any sensible combination will have a single enum shortcut but hey, at least they're there if you want them and sometime in the future there might be a decent reason to use them directly. And my enum code wouldn't even have to be changed for now (because handling the bitmasks directly is still an ugly pain compared to my switch).

I'm just going to wait till I see your version before I do anything else here.

RobLoach (approved):
This is working for me. I really like that I'm able to read the tooltips now 😆

rswinkle:
> This is working for me. I really like that I'm able to read the tooltips now 😆

Thanks, me too, and it is long overdue. Honestly I think it's fine as is but if you want to chime in on the duplicate enum/bitmask thing? Again it would be nice to just get rid of the other two sets and just have a single set for alignment purposes, and just go to 5.0.0, along with many other things. I still need to reply to your Meta issue but sleeptight's wall of text has been intimidating, ha.

Also I've been working on a project using Nuklear and while I already had a backlog of minor issues/bugs/wish list items, this has only reminded me of several of them and added more. Hopefully I'll start to make some issues/PRs in the near future and am more focused on polish.

sleeptightAnsiC:
EDIT: https://github.com/Immediate-Mode-UI/Nuklear/pull/900

Just a reminder that I remember about this one, and still planning to submit an alternative.

Apologies for this taking so long. I had few higher priorities in past week. I hope I will deal with this soon.

(and also sorry for that wall of text under META issue, hah)

RobLoach:
Alternative? Can you build upon what's here? Anything before we merge?

sleeptightAnsiC:
EDIT: https://github.com/Immediate-Mode-UI/Nuklear/pull/900

Yea, an alternative. I though you did read https://github.com/Immediate-Mode-UI/Nuklear/pull/881#issuecomment-3840426388 ? Please wait with merging.

All my current plans for this PR are:
- apply some reasonable offset for the default tooltip (this is still broken for the default one) 
- try with bitmask instead of enum (just as we talked about) 
- use prefix for tooltip related symbols
- some todos from code that I wanted to investigate
- ~~report new task about dealing with offscreen behavior (this is not only affecting tooltips)~~ EDIT: https://github.com/Immediate-Mode-UI/Nuklear/issues/899
- and probably something more that I don't remeber right now

src/nuklear_tooltip.c
```diff
@@ -55,6 +91,43 @@ nk_tooltip_end(struct nk_context *ctx)
     nk_popup_close(ctx);
     nk_popup_end(ctx);
 }
+
+NK_API void
+nk_tooltip_pos_offset(struct nk_context *ctx, const char *text, enum nk_tooltip_pos position, struct nk_vec2 offset)
+{
```
sleeptightAnsiC, line 110: EDIT: This was fixed in https://github.com/Immediate-Mode-UI/Nuklear/pull/881/changes/a825bda2503a57fe82eba1725e5b655f83692646

seems like you have copied all this code from `nk_tooltip`, but forgot to delete it there? Right now, there are two big duplicated code chunks in those two places (this function and the next one). Anyway, I fixed this in my alternative version.

sleeptightAnsiC:
V2 https://github.com/Immediate-Mode-UI/Nuklear/pull/900

Since there was quite a lot of changes, pasting a diff wouldn't make sense, I just submitted another DRAFT. We can keep talking here, if you prefer to keep all the info in one place.

rswinkle:
The version with styling #911 was already merged so I'm closing this.

Issue #812, opened 2025-05-29
`nk_style_slider` and `nk_style_scrollbar`'s `show_buttons` member typing is incorrect


Emily-TTG:
`nk_style_slider` and `nk_style_scrollbar`'s `show_buttons` member is `int` instead of `nk_bool`. This causes an issue with `nk_checkbox_label` and the like where the passed pointer can cause a mismatch if `NK_INCLUDE_STANDARD_BOOL` is set.

This manifests in the configurator demo in both structures' respective functions.

sleeptightAnsiC:
Possibly related to https://github.com/Immediate-Mode-UI/Nuklear/issues/849
Probably caused by https://github.com/Immediate-Mode-UI/Nuklear/pull/185

PavelSharp:
Hello, @Emily-TTG. Thanks a lot for reporting this. 

The underlying issue was introduced after PR #185 was merged.  Besides that, it results in undefined behavior in `style_configurator.c` whenever `nk_bool` is redefined to a type other than `int`.

https://github.com/Immediate-Mode-UI/Nuklear/blob/9afb3dd6d14fba0c5e62f08809b494b6d580b9b1/demo/common/style_configurator.c#L302
**as well as**
https://github.com/Immediate-Mode-UI/Nuklear/blob/9afb3dd6d14fba0c5e62f08809b494b6d580b9b1/demo/common/style_configurator.c#L383

sleeptightAnsiC:
> This causes an issue with nk_checkbox_label and the like where the passed pointer can cause a mismatch

@Emily-TTG you described this like if there were other symbols than `nk_checkbox_label` that could cause it. Was this simply a wild guess, or do you remember any of those, maybe?

<details><summary> If not, then I have a very long and complicated explanation, that proofs it only affects the two places from Pavel's comment above. (clicking this is dangerous...)</summary>
<p>

(I was figuring this out as I was typing)

1. Find all struct member with `int` ? I did it with AWK script, not sure if correctly though:

```sh
~/Nuklear (master) $ awk '                                                                                                                                                                                                                                     
function _found()          { print FILENAME ":" FNR ":" $0 }                                                                                                                                                                                                   
function _err()            { _found(); print " ^ this should never happen!" >/dev/null; exit(1) }                                                                                                                                                              
BEGIN                      { inside=0; }                                                                                                                                                                                                                       
[...]

Issue #849, opened 2025-11-04
suspicious static checks for sizeof(nk_bool)


sleeptightAnsiC:
If you try to provide your own `NK_BOOL` - either by defining it to C99 `_Bool`, C++/C23 `bool`, or some small integer type like `int8_t` - there will be two static asserts that fail and prevent you to do so.

repro:
```sh
$ git clone https://github.com/Immediate-Mode-UI/Nuklear.git
$ cd Nuklear/
$ cat <<'EOF' >testcase.c
    #define NK_BOOL _Bool                                                                                                                                                                                                                                      
    #define NK_IMPLEMENTATION                                                                                                                                                                                                                                  
    #include "./nuklear.h"                                                                                                                                                                                                                                     
    int main(void) { return 0; }                                                                                                                                                                                                                               
EOF
$ cc testcase.c
```

<details><summary>result (CLICK ME)</summary>
<p>

```
In file included from testcase.c:3:
./nuklear.h:302:61: error: size of array ‘_dummy_array449’ is negative
  302 |   #define NK_STATIC_ASSERT(exp) typedef char NK_UNIQUE_NAME(_dummy_array)[(exp)?1:-1]
      |                                                             ^~~~~~~~~~~~
./nuklear.h:291:46: note: in definition of macro ‘NK_STRING_JOIN_IMMEDIATE’
  291 | #define NK_STRING_JOIN_IMMEDIATE(arg1, arg2) arg1 ## arg2
      |                                              ^~~~
./nuklear.h:293:36: note: in expansion of macro ‘NK_STRING_JOIN_DELAY’
  293 | #define NK_STRING_JOIN(arg1, arg2) NK_STRING_JOIN_DELAY(arg1, arg2)
      |                                    ^~~~~~~~~~~~~~~~~~~~
./nuklear.h:298:32: note: in expansion of macro ‘NK_STRING_JOIN’
  298 |   #define NK_UNIQUE_NAME(name) NK_STRING_JOIN(name,__LINE__)
      |                                ^~~~~~~~~~~~~~
./nuklear.h:302:46: note: in expansion of macro ‘NK_UNIQUE_NAME’
  302 |   #define NK_STATIC_ASSERT(exp) typedef char NK_UNIQUE_NAME(_dummy_array)[(exp)?1:-1]
      |                                              ^~~~~~~~~~~~~~
./nuklear.h:449:1: note: in expansion of macro ‘NK_STATIC_ASSERT’
[...]

PR #883, opened 2026-01-31
fix(style_configurator): fix unsafe int to nk_bool pointer casts

sleeptightAnsiC:
Fixes: https://github.com/Immediate-Mode-UI/Nuklear/issues/812

For more detailed explanation see the commit message or the linked issue. Also, if you're wondering if those are the only places affected by this issue, [I really think there is nothing else left :)](https://github.com/Immediate-Mode-UI/Nuklear/issues/812#issuecomment-3829232745)

### DIFF
--- a/demo/common/style_configurator.c
+++ b/demo/common/style_configurator.c
@@ -275,6 +275,7 @@ style_slider(struct nk_context* ctx, struct nk_style_slider* out_style)
 {
 	struct nk_style_slider slider = *out_style;
 	struct nk_style_button* dups[1];
+	nk_bool show_buttons_b;
 
 	nk_layout_row_dynamic(ctx, 30, 2);
 
@@ -299,7 +300,9 @@ style_slider(struct nk_context* ctx, struct nk_style_slider* out_style)
 	nk_property_float(ctx, "#Rounding:", -100.0f, &slider.rounding, 100.0f, 1,0.5f);
 
 	nk_layout_row_dynamic(ctx, 30, 1);
-	nk_checkbox_label(ctx, "Show Buttons", &slider.show_buttons);
+	show_buttons_b = (nk_bool)slider.show_buttons;
+	nk_checkbox_label(ctx, "Show Buttons", &show_buttons_b);
+	slider.show_buttons = (int)show_buttons_b;
 
 	if (slider.show_buttons) {
 		nk_layout_row_dynamic(ctx, 30, 2);
@@ -355,6 +358,7 @@ style_scrollbars(struct nk_context* ctx, struct nk_style_scrollbar* out_style, s
 {
 	struct nk_style_scrollbar scroll = *out_style;
 	struct nk_style_button* dups[3];
+	nk_bool show_buttons_b;
 
 	nk_layout_row_dynamic(ctx, 30, 2);
 
@@ -380,7 +384,9 @@ style_scrollbars(struct nk_context* ctx, struct nk_style_scrollbar* out_style, s
 
 	/* TODO what is wrong with scrollbar buttons?  Also look into controlling the total width (and height) of scrollbars */
 	nk_layout_row_dynamic(ctx, 30, 1);
-	nk_checkbox_label(ctx, "Show Buttons", &scroll.show_buttons);
+	show_buttons_b = (nk_bool)scroll.show_buttons;
+	nk_checkbox_label(ctx, "Show Buttons", &show_buttons_b);
+	scroll.show_buttons = (int)show_buttons_b;
 
 	if (scroll.show_buttons) {
 		nk_layout_row_dynamic(ctx, 30, 2);


PavelSharp:
In fact, fixing this bug is not a trivial task. There are two possible strategies:

1. Preserve the existing API.
In this case, priority is given to API stability, and existing user code is not affected. It may remain either working or broken. The main drawback is that users will intuitively expect `show_buttons` to be a boolean type, which can still lead to potential UB in similar cases.

2. Change the field type in `nk_style_slider` and `nk_style_scrollbar` to `nk_bool`.
This would break the existing API and user code. However, I believe this is a worthwhile long-term investment, since user expectations would better match reality by reducing surprises and making the library more predictable.

sleeptightAnsiC:
> 1\. Preserve the existing API.

That wasn't really my intent.

I just wanted to fix this thing and move on. I only did it because I was investigating other issue related to nk_bool (https://github.com/Immediate-Mode-UI/Nuklear/issues/849) and found out this is the safest approach. This thing was working fine in the past and https://github.com/Immediate-Mode-UI/Nuklear/pull/185 broke it only because our CI can't really catch this (but maybe it should...) I tried to find other possible places affected by this and found none (explained in https://github.com/Immediate-Mode-UI/Nuklear/issues/812#issuecomment-3829232745 )

> The main drawback is that users will intuitively expect show_buttons to be a boolean type, which can still lead to potential UB in similar cases.

These days, if you find yourself writing a code with similar defect (aka passing int* into bool*), a compiler with half-decent warnings enabled should catch this. In fact, GCC 15 does this by default and reports as error. Proof here: https://godbolt.org/z/4Pjncdjqn And if your compiler cannot do this, well... you're screwed anyway, because those kinds of issues can happen all over the C code.

> 2\. Change the field type in `nk_style_slider` and `nk_style_scrollbar` to `nk_bool`.

I would rather not do this.

Not only there are way more int fields with bool semantics like these two, there is simply no other use of nk_bool type anywhere in the style structs (EDIT: I think I found one somewhere). If you want these two changed, you should probably also change other similar fields elsewhere, and then go through each line that uses them. This in return will create even bigger mess and more bugs.

sleeptightAnsiC:
> 2\. Change the field type in `nk_style_slider` and `nk_style_scrollbar` to `nk_bool`.
>     This would break the existing API and user code. However, I believe this is a worthwhile long-term investment
>
> > there is simply no other use of nk_bool anywhere in the style structs. If you want these two changed, you should probably also change other similar fields elsewhere

Btw, I think I remember some people talking about serializing those structs. So there is some benefit of keeping them the same, and adding nk_bool would definitely break this use case. You can't really serialize bool safely, although serializing raw signed integer is not a good idea neither, it will still probably work on Win32 platforms. That might be why they haven't changed it back in https://github.com/Immediate-Mode-UI/Nuklear/pull/185 and why those were never changed from `int` to something like `nk_int` (or `nk_int32`). Changing any types in those structs feels like playing Russian Roulette. Don't do this, please...

demo/common/style_configurator.c
```diff
@@ -299,7 +300,9 @@ style_slider(struct nk_context* ctx, struct nk_style_slider* out_style)
 	nk_property_float(ctx, "#Rounding:", -100.0f, &slider.rounding, 100.0f, 1,0.5f);
 
 	nk_layout_row_dynamic(ctx, 30, 1);
-	nk_checkbox_label(ctx, "Show Buttons", &slider.show_buttons);
+	show_buttons_b = slider.show_buttons;
+	nk_checkbox_label(ctx, "Show Buttons", &show_buttons_b);
+	slider.show_buttons = show_buttons_b;
```
RobLoach: Maybe being explicit about the type change? Does this break compilation?
``` suggestion
	slider.show_buttons = (int)show_buttons_b;
```

demo/common/style_configurator.c
```diff
@@ -380,7 +384,9 @@ style_scrollbars(struct nk_context* ctx, struct nk_style_scrollbar* out_style, s
 
 	/* TODO what is wrong with scrollbar buttons?  Also look into controlling the total width (and height) of scrollbars */
 	nk_layout_row_dynamic(ctx, 30, 1);
-	nk_checkbox_label(ctx, "Show Buttons", &scroll.show_buttons);
+	show_buttons_b = scroll.show_buttons;
+	nk_checkbox_label(ctx, "Show Buttons", &show_buttons_b);
+	scroll.show_buttons = show_buttons_b;
```
RobLoach:
``` suggestion
	scroll.show_buttons = (int)show_buttons_b;
```

RobLoach:
Overall, I think this makes sense. Making `show_buttons` a `nk_bool` would also make sense to me, despite potentially breaking things :shrug:

sleeptightAnsiC:
@RobLoach latest force-push makes those casts explicit, just like you requested.

PavelSharp:
@RobLoach As you can see, sleeptightAnsiC has already provided several arguments.
However, I guess the discussion above does not give enough attention to the fact that #185 introduced an inconsistency (defect). 

At the moment, Nuklear contains several structures that use `nk_bool` as a data member, among them the style structure `struct nk_style_chart`, where `show_markers` is defined as `nk_bool`:
https://github.com/Immediate-Mode-UI/Nuklear/blob/25f33c8780a8b41668447f83e7d154e735cf424f/src/nuklear.h#L5228

But in `nk_style_slider` and `nk_style_scrollbar` this was apparently overlooked:
https://github.com/Immediate-Mode-UI/Nuklear/blob/25f33c8780a8b41668447f83e7d154e735cf424f/src/nuklear.h#L5038
https://github.com/Immediate-Mode-UI/Nuklear/blob/25f33c8780a8b41668447f83e7d154e735cf424f/src/nuklear.h#L5134

Regarding API/ABI breakage, I believe the impact would be limited, since Nuklear defines `nk_bool` as `int` by default. This would only affect cases where `nk_bool` was explicitly [redefined by the user](https://github.com/Immediate-Mode-UI/Nuklear/blob/25f33c8780a8b41668447f83e7d154e735cf424f/src/nuklear.h#L187-L194).

sleeptightAnsiC:
> I guess the discussion above does not give enough attention to the fact that #185 introduced an inconsistency (defect).
> At the moment, Nuklear contains several structures that `nk_bool` uses as a data member, among them the style structure `struct nk_style_chart`, where `show_markers` is defined as `nk_bool`:
> But in `nk_style_slider` and `nk_style_scrollbar` this was apparently overlooked:

I believe, there are more issues like this one, not only in styling structs. I personally don't want to touch this (for all the reasons mentioned before). This can be fixed as another Issue/PR, but not this one.

PavelSharp:
> I believe, there are more issues like this one, not only in styling structs.

I may missing something, but I’m not entirely sure why this issue is being generalized to such a broad scope in the discussion.
As shown above, these two fields appear to be a concrete API defect, and fixing them would already improve API consistency. However, by broadening the problem to all similar cases, the conclusion naturally becomes that fixing it would be complex and risky. To me, this feels like a conceptual trap.

sleeptightAnsiC:
> these two fields appear to be a concrete API defect, and fixing them would already improve API consistency. However, by broadening the problem to all similar cases, the conclusion naturally becomes that fixing it would be complex and risky. To me, this feels like a conceptual trap.

Those were the only ones that we had the code coverage to catch. If someone can write a new faulty code that casts int* to bool* , then this can be caused by literally anything. The fact that it happened here wasn't caused by someone being confused by the API, but because https://github.com/Immediate-Mode-UI/Nuklear/pull/185 forgot to change this code path. That's all.

> I may missing something, but I’m not entirely sure why this issue is being generalized to such a broad scope in the discussion.

Yes, you do:
https://github.com/Immediate-Mode-UI/Nuklear/blob/25f33c8780a8b41668447f83e7d154e735cf424f/src/nuklear.h#L5359
https://github.com/Immediate-Mode-UI/Nuklear/blob/25f33c8780a8b41668447f83e7d154e735cf424f/src/nuklear.h#L4642
https://github.com/Immediate-Mode-UI/Nuklear/blob/25f33c8780a8b41668447f83e7d154e735cf424f/src/nuklear.h#L5741
(and probably way more)

@PavelSharp This PR is only about fixing those two specific code paths. Either take it or leave it. If you want the struct methods changed, go open a new PR/Issue for it. I'm not going to do this myself (you already know my opinion on this).

sleeptightAnsiC:
I just entered a new issue about all those int fields https://github.com/Immediate-Mode-UI/Nuklear/issues/891

The conversation above is irrelevant here. This PR only fixes the two existing code paths. Nothing else.

(also, if we would ever want to change those to nk_bool, the code in this PR will still be compatible with it)

sleeptightAnsiC:
@RobLoach Fell free to either merge or reject this one. I will not be messing with those structs here.

PR #884, opened 2026-01-31
Fix(build): replace incorrect static asserts for size(nk_bool)

sleeptightAnsiC:
Fixes: https://github.com/Immediate-Mode-UI/Nuklear/issues/849

See the linked issue and the commit message for more detailed explanation.

I was thinking about fixing this one for a while, trying to understand if there are any reasoning behind previous approach, but I really think it was overlooked and incorrectly implemented to begin with, back in https://github.com/Immediate-Mode-UI/Nuklear/pull/185

### DIFF
--- a/demo/sdl3_renderer/main.c
+++ b/demo/sdl3_renderer/main.c
@@ -58,11 +58,8 @@
     #define NK_POINTER_TYPE      uintptr_t
 #endif
 
-/* FIXME: We could also use the `bool` symbol provided by SDL,
- * but this is currently broken due to internal Nuklear issue, see:
- * https://github.com/Immediate-Mode-UI/Nuklear/issues/849
- * */
-#define NK_INCLUDE_STANDARD_BOOL
+/* We can reuse the `bool` symbol because SDL3 guarantees its existence */
+/*#define NK_INCLUDE_STANDARD_BOOL*/
 #ifndef NK_INCLUDE_STANDARD_BOOL
     #define NK_BOOL               bool
 #endif
--- a/src/nuklear.h
+++ b/src/nuklear.h
@@ -184,12 +184,13 @@ extern "C" {
   #endif
 #endif
 
+/**< could be any type with semantic of standard bool, either equal or smaller than int */
 #ifndef NK_BOOL
   #ifdef NK_INCLUDE_STANDARD_BOOL
     #include <stdbool.h>
     #define NK_BOOL bool
   #else
-    #define NK_BOOL int /**< could be char, use int for drop-in replacement backwards compatibility */
+    #define NK_BOOL int
   #endif
 #endif
 
@@ -220,11 +221,7 @@ NK_STATIC_ASSERT(sizeof(nk_flags) >= 4);
 NK_STATIC_ASSERT(sizeof(nk_rune) >= 4);
 NK_STATIC_ASSERT(sizeof(nk_size) >= sizeof(void*));
 NK_STATIC_ASSERT(sizeof(nk_ptr) >= sizeof(void*));
-#ifdef NK_INCLUDE_STANDARD_BOOL
-NK_STATIC_ASSERT(sizeof(nk_bool) == sizeof(bool));
-#else
-NK_STATIC_ASSERT(sizeof(nk_bool) >= 2);
-#endif
+NK_STATIC_ASSERT(sizeof(nk_bool) <= sizeof(int));
 
 /* ============================================================================
  *
@@ -270,6 +267,10 @@ struct nk_image {nk_handle handle; nk_ushort w, h; nk_ushort region[4];};
 struct nk_nine_slice {struct nk_image img; nk_ushort l, t, r, b;};
 struct nk_cursor {struct nk_image img; struct nk_vec2 size, offset;};
 struct nk_scroll {nk_uint x, y;};
+
+/* Make sure the semantic of nk_true/nk_false is compatible with nk_bool */
+NK_STATIC_ASSERT(!((nk_bool)0) == !(nk_false));
+NK_STATIC_ASSERT(!((nk_bool)1) == !(nk_true));
 
 enum nk_heading         {NK_UP, NK_RIGHT, NK_DOWN, NK_LEFT};
 enum nk_button_behavior {NK_BUTTON_DEFAULT, NK_BUTTON_REPEATER};
@@ -1925,7 +1926,7 @@ NK_API void nk_window_show_if(struct nk_context *ctx, const char *name, enum nk_
  * # # nk_window_show_if
  * Line for visual separation. Draws a line with thickness determined by the current row height.
  * ```c
- * void nk_rule_horizontal(struct nk_context *ctx, struct nk_color color, NK_BOOL rounding)
+ * void nk_rule_horizontal(struct nk_context *ctx, struct nk_color color, nk_bool rounding)
  * ```
  *
  * Parameter       | Description
--- a/src/nuklear_internal.h
+++ b/src/nuklear_internal.h
@@ -71,11 +71,7 @@ NK_STATIC_ASSERT(sizeof(nk_short) == 2);
 NK_STATIC_ASSERT(sizeof(nk_uint) == 4);
 NK_STATIC_ASSERT(sizeof(nk_int) == 4);
 NK_STATIC_ASSERT(sizeof(nk_byte) == 1);
-#ifdef NK_INCLUDE_STANDARD_BOOL
-NK_STATIC_ASSERT(sizeof(nk_bool) == sizeof(bool));
-#else
-NK_STATIC_ASSERT(sizeof(nk_bool) == 4);
-#endif
+NK_STATIC_ASSERT(sizeof(nk_bool) <= sizeof(int));
 
 NK_GLOBAL const struct nk_rect nk_null_rect = {-8192.0f, -8192.0f, 16384, 16384};
 #define NK_FLOAT_PRECISION 0.00000000000001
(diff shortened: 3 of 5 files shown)


src/nuklear.h
sleeptightAnsiC: I think I should also update this comment (and documentation?) because it may suggest that you can only use char and int, which is not true. The library itself can set it to stdbool which is already a well tested path.

EDIT: fixed with latest force push.

src/nuklear_internal.h
sleeptightAnsiC, line 64: (unrelated to this PR)

I wonder why do we need these duplicated checks here? The ones from `src/nuklear.h` do the very same thing, and all src/*.c files are including both of those headers, so I'm not sure what is the benefit?

This thing can only really harm us, just like in case of https://github.com/Immediate-Mode-UI/Nuklear/pull/185 where the `sizeof(nk_bool) == ...` checks were different between each file, and went unnoticed for so long.

~~EDIT: TODO: make sure to report this thing as separate Issue after _this_ PR gets merged.~~

EDIT: moved to https://github.com/Immediate-Mode-UI/Nuklear/issues/889

demo/sdl3_renderer/main.c
```diff
@@ -58,11 +58,8 @@
     #define NK_POINTER_TYPE      uintptr_t
 #endif
 
-/* FIXME: We could also use the `bool` symbol provided by SDL,
- * but this is currently broken due to internal Nuklear issue, see:
- * https://github.com/Immediate-Mode-UI/Nuklear/issues/849
- * */
-#define NK_INCLUDE_STANDARD_BOOL
+/* We can reuse the `bool` symbol because SDL3 guarantees its existence */
+/*#define NK_INCLUDE_STANDARD_BOOL*/
```
RobLoach, line 62: Could we safely set this in the sdl3 renderer header itself?

sleeptightAnsiC: No, because this is up to developer to setup, and you also need those to propagate to the Nuklear header itself, which is just impossible (it would create circular dependency, chicken'n'egg problem)

However, I had some idea about reusing this config in easier way. I could maybe submit another PR for this, if you're interested ?

RobLoach: Anything to make use easier 👍

sleeptightAnsiC: I just reported a task for this https://github.com/Immediate-Mode-UI/Nuklear/issues/886 I will submit it as PR later, if I see someone's approval under it.

This problem is not related to _this_ specific PR, so I'm going to resolve _this_ thread as soon as I notice you responding there (or here).

src/nuklear.h
```diff
@@ -220,11 +220,7 @@ NK_STATIC_ASSERT(sizeof(nk_flags) >= 4);
 NK_STATIC_ASSERT(sizeof(nk_rune) >= 4);
 NK_STATIC_ASSERT(sizeof(nk_size) >= sizeof(void*));
 NK_STATIC_ASSERT(sizeof(nk_ptr) >= sizeof(void*));
-#ifdef NK_INCLUDE_STANDARD_BOOL
-NK_STATIC_ASSERT(sizeof(nk_bool) == sizeof(bool));
-#else
-NK_STATIC_ASSERT(sizeof(nk_bool) >= 2);
-#endif
+NK_STATIC_ASSERT(sizeof(nk_bool) <= sizeof(int));
```
PavelSharp, line 224: Could you clarify the purpose of this check? As I remember, the C99 standard allows `_Bool` to have any size(In theory), so I’m curious what scenarios this assertion is intended to cover.

sleeptightAnsiC: `nk_bool` was historically an `int` in this library
(in practice, it was signed 32bit integer, because most of people used it on x86_64)
so if you provide any bigger type than `int` it will most likely not work
because code may do something strange that it never expected in the first place

If you're on platform where `sizeof(_Bool) > sizeof(int)` then I would say this check is justified to fail, because there would be something really wrong going on...

This library was never tested with anything bigger. We can't safely allow it. And why would you want a bigger boolean anyway (?)

PavelSharp: The [current static assertion system](https://github.com/Immediate-Mode-UI/Nuklear/blob/25f33c8780a8b41668447f83e7d154e735cf424f/src/nuklear_internal.h#L74-L78) requires the size of `nk_bool` to be either the same as `_Bool` or equal to 4.

However, this PR introduces new rules that are not compatible with the C standard (since the size of `_Bool` is arbitrary) and are not compatible with the current rules. Because of that, it feels a bit unnatural and unexpected.

It would be right to explain why this new check is needed. If it’s meant to prevent real bugs, showing at least one concrete example would really help. Otherwise, it’s hard to see a strong justification for it.

sleeptightAnsiC: Those current checks are totally bonkers. The one from [src/nuklear.h](https://github.com/Immediate-Mode-UI/Nuklear/blob/25f33c8780a8b41668447f83e7d154e735cf424f/src/nuklear.h#L226) says `sizeof(nk_bool) >= 2`

I can change my patch to `sizeof(nk_bool) <= 4`, if you're so worried about it, but you have to have in mind that previous ones were just totally wrong. (Those never worked correctly since https://github.com/Immediate-Mode-UI/Nuklear/pull/185 was merged)

On platforms where Nuklear is tested, the `sizeof(int)` is already exactly `4`, and if my change breaks someone's obscure hacks on some very rare platform, then I would say they had it coming, because the only thing I can imagine this breaking, is when someone used `long`, or something even more incorrect that they shouldn't.

> However, this PR introduces new rules that are not compatible with the C standard (since the size of _Bool is arbitrary) and are not compatible with the current rules. Because of that, it feels a bit unnatural and unexpected.

Please, show me at least one platform where `sizeof(_Bool) <= sizeof(int)` evaluates to `0` or stop wasting my time. I have no idea where this assumption comes from... How would Nuklear be even able to work on such platform?

PavelSharp: As you mentioned, the current static assert system are inconsistent, yeah, it's true.
So the strongest constraint to imply the `nk_bool` is expected to be either the same size as `bool` or 4 bytes.

Given that, this PR could have addressed the issue in a direct way, by using something like:
`NK_STATIC_ASSERT(sizeof(nk_bool) == sizeof(bool) || sizeof(nk_bool) == sizeof(int));`

Instead, this PR introduces nonequivalent check, without proper justification.

**Another fact**. Considering [this line](https://github.com/Immediate-Mode-UI/Nuklear/blob/25f33c8780a8b41668447f83e7d154e735cf424f/src/nuklear_internal.h#L75) and considering C standard (the size of `_Bool` as implementation-defined), it logically follows that any check restricting the size of `nk_bool` **is redundant**.  It's just not necessary.

sleeptightAnsiC: > Another fact. Considering [this line](https://github.com/Immediate-Mode-UI/Nuklear/blob/25f33c8780a8b41668447f83e7d154e735cf424f/src/nuklear_internal.h#L75) and considering C standard (the size of _Bool as implementation-defined), it logically follows that any check restricting the size of nk_bool is redundant. It's just not necessary.

Dude, how can you even consider this line to be correct and worth mentioning about? Are you really not seeing what the the preprocessor path leading to this check is? It literally checks nothing:

```c
#ifndef NK_BOOL
  #ifdef NK_INCLUDE_STANDARD_BOOL
    #include <stdbool.h>
    #define NK_BOOL bool
  #else
    // ...
  #endif
#endif

// ...

typedef NK_BOOL nk_bool;

// ...

#ifdef NK_INCLUDE_STANDARD_BOOL
NK_STATIC_ASSERT(sizeof(nk_bool) == sizeof(bool));
#else
//...
#endif

// then it repeats in src/nuklear_internal.h 
```

How can you consider this...

```c
    #include <stdbool.h>
    #define NK_BOOL bool
    typedef NK_BOOL nk_bool;
    NK_STATIC_ASSERT(sizeof(nk_bool) == sizeof(bool));
```
...to be anything meaningful? This check would never fail, and was never proving anything. Why do you think that someone writing this had any C rule in mind? and we should randomly consider said rule here? It just makes no sense at all.

### I will not be allowing arbitrary nk_bool, just because there is some ancient C standard rule, that under-specifies how big maximum size of _Bool can be. That library used `int` in the past, and allowing anything bigger is irrational.

sleeptightAnsiC: @PavelSharp Please, stop... I'm resolving this thread, and I won't be replaying to your comments under this PR. At this point, you're seriously getting hallucinated by whatever LLM you're using, and that's wasting my time and energy.

This is a classic AI-halucination-feedback-loop: it read the code that was meaningless to begin with, it told you some theoretical assumption based on it, then ignored all my points (literally none of my questions were answered), and went with the same wrong answer over and over again, just with a different wording each time. This is seriously wrong and abusive. I won't be tolerating it.

If any other contributor wishes to talk about this, I will be fine to continue.

If anyone ever finds a platform, where this check is not behaving as expected, then please let me know, and remember to include all the details about your environment.

EDIT: Changed this comment to add more context for people who really don't understand it. If you're using AI agent in order to reason about old buggy code, that was wrong back when it was implemented, the AI will always deceive you. The suggestion from https://github.com/Immediate-Mode-UI/Nuklear/pull/884#discussion_r2761014495 wouldn't even compile...

sleeptightAnsiC:
Rebased, fixed that doc related thing, and moved everything unrelated to different issues.

@RobLoach mind merging, or at least taking a look again?

RobLoach:
Merging this one. I don't think there would ever really be a case where bools are larger than ints, and it's likely good to know when that does happen. Thanks.

Issue #890, opened 2026-02-06
add CI for detecting a mismatch between ./src/* and ./nuklear.h


sleeptightAnsiC:
follow up from https://github.com/Immediate-Mode-UI/Nuklear/pull/845 ...
> > > > I think it's fine to avoid re-building it for now. The change will be brought in next time it's built.
> > > 
> > > 
> > > This is not okey. Other PRs may have some problems because of this, like merge conflicts, confusing diffs, and so on... I think we can prevent similar situations with CI
> > 
> > 
> > Rebuilt manually. CI rebuild would be nice.
> 
> I meant CI for detecting a mismatch, but I guess auto-rebuilding would also be nice. (though, I'm not sure how to implement the latter one)

...and from https://github.com/Immediate-Mode-UI/Nuklear/issues/887 ...

> > Build out the CI to build nuklear.h for us, so that git conflicts don't happen
>
> How about I create a similar script to [.github/ci_compile_sources.sh](https://github.com/Immediate-Mode-UI/Nuklear/blob/master/.github/ci_compile_sources.sh) that will look for incompatibilities between ./nuklear.h and ./src/* and will block PRs based on it? Still not automatic but should be an improvement.

I'll probably add a PR for this in future. (only for auto detection, NOT for auto rebuild)

My biggest problem with solving this right now is that [./src/paq.sh](https://github.com/Immediate-Mode-UI/Nuklear/blob/25f33c8780a8b41668447f83e7d154e735cf424f/src/paq.sh) and [./src/paq.bat](https://github.com/Immediate-Mode-UI/Nuklear/blob/25f33c8780a8b41668447f83e7d154e735cf424f/src/paq.bat) have hardcoded arguments and output, which makes it problematic to reuse inside of CI script, and I need to deal with it before doing anything.

sleeptightAnsiC:
Seems like Rob is cooking something here: https://github.com/Immediate-Mode-UI/Nuklear/pull/892 so I will wait with doing anything until the other one is done. Maybe we won't need this after all.

RobLoach:
Rather than erroring out on mismatch, I think it'll ease the workflow if we just build it directly on `master`.

### Questions

- [ ] Remove CHANGELOG?
- [ ] Version number?

sleeptightAnsiC:
> Rather than erroring out on mismatch, I think it'll ease the workflow if we just build it directly on master.

[...]

PR #892, opened 2026-02-07
GitHub Actions: Build and Push

RobLoach:
This adds a CI workflow to paq.sh and push the changes to master.

### DIFF
--- a/.github/workflows/build-and-push.yml
+++ b/.github/workflows/build-and-push.yml
@@ -27,7 +27,7 @@ jobs:
           git config user.name "github-actions[bot]"
           git config user.email "github-actions[bot]@users.noreply.github.com"
           git add .
-          git commit -m "Automated build: run ./src/paq.sh" || echo "No changes to commit"
+          git commit -m "Automated build" || echo "No changes to commit"
 
       - name: Push Changes
         uses: ad-m/github-push-action@v1.0.0


demo/x11/main.c
```diff
@@ -79,7 +79,7 @@ sleep_for(long t)
 /*#define INCLUDE_ALL */
 /*#define INCLUDE_STYLE */
 /*#define INCLUDE_CALCULATOR */
-/*#define INCLUDE_CANVAS */
+#define INCLUDE_CANVAS
```
RobLoach:
``` suggestion
/*#define INCLUDE_CANVAS */
```

.github/workflows/build-and-push.yml
```diff
@@ -0,0 +1,35 @@
...
+
+      - name: Commit Changes
+        run: |
+          git config user.name "github-actions[bot]"
+          git config user.email "github-actions[bot]@users.noreply.github.com"
+          git add .
+          git commit -m "Automated build: run ./src/paq.sh" || echo "No changes to commit"
+
+      - name: Push Changes
+        uses: ad-m/github-push-action@v0.6.0
```
RobLoach:
``` suggestion
        uses: ad-m/github-push-action@v1.0.0
```

.github/workflows/build-and-push.yml
```diff
@@ -0,0 +1,35 @@
...
+      - name: Run the Build Script
+        run: ./paq.sh
+        working-directory: src
+
+      - name: Commit Changes
+        run: |
+          git config user.name "github-actions[bot]"
+          git config user.email "github-actions[bot]@users.noreply.github.com"
+          git add .
+          git commit -m "Automated build: run ./src/paq.sh" || echo "No changes to commit"
```
RobLoach:
``` suggestion
          git commit -m "Automated build" || echo "No changes to commit"
```

PavelSharp:
It seems this PR enables automatic building of `Nuklear.h`. To prevent users from accidentally modifying this generated file, maybe we should consider adding it to `.gitignore`?  In that case, we would need to use [git add -f](https://git-scm.com/docs/git-add#Documentation/git-add.txt--f) in the workflow to commit it.
To my knowledge, GitHub Actions Workflows run in parallel. This means that `ccpp.yml` could fail while `build-and-push.yml` completes successfully.

RobLoach:
> It seems this PR enables automatic building of Nuklear.h. To prevent users from accidentally modifying this generated file, maybe we should consider adding it to .gitignore?

While this seems sound, many people just git clone nuklear, and then refernecee nuklear.h directly. Wouldn't want to break their workflow.

> To my knowledge, GitHub Actions Workflows run in parallel. This means that ccpp.yml could fail while build-and-push.yml completes successfully.

Good call. We'll have to make sure this runs after the tests.

sleeptightAnsiC:
Moving my comment from https://github.com/Immediate-Mode-UI/Nuklear/issues/890#issuecomment-3884545259 as it was really about this PR:

> > I think it'll ease the workflow if we just build it directly on master.
> 
> That would be really nice and the best solution. I simply don't know how to implement this correctly and have few concerns. If you know the answers, we can just abandon my idea and do the other one istead:
> * How do we know, if that CI is working correctly? In case of a bug, it would destroy the git history on master branch. Are there any other projects using this approach so we can yank their solution?
> * How do we ensure that people are not submitting changes to single-header-lib anymore? Seems like we will need another CI for this one, right?
> * What shall we do with all open PRs? Some of those are already following the workflow correctly. Would those break?

^ I think, I should implement https://github.com/Immediate-Mode-UI/Nuklear/issues/890 anyway, because even if we won't need it, you could still somehow reuse it here (?)

Any thoughts? @RobLoach

RobLoach:
Maybe just a test is a worth-while change in the mean time. Thanks.

Issue #888, opened 2026-02-05
[META::Emscripten] Live version integration strategy and progress report


PavelSharp:
### The first step in integrating Emscripten is to choose the backend. See the report below.

Emscripten Integration Report
-----------------------------------------
### Good News
1. ⭐⭐ Successfully built and tested `demo/sdl3_renderer`.
Everything works, including builds with `INCLUDE_ALL`. However, this requires either merging the small PR #883 or manually fixing it by adjusting field types.
2. ⭐ Successfully built and tested `demo/sdl_opengles2`. 
Everything works as expected. Untested with `INCLUDE_ALL`.
3.  ❌Successfully built `demo/sdl_renderer`. 
Launch is currently unsuccessful, since opening `index.html` results in an infinite loading loop. But I didn't take many attempts.

### Bad News
1.  Standard examples (`Nuklear/example/`) based on `GLEW/GLFW` do not build due to `glMapBuffer` and `glUnmapBuffer` calls.
    1.1  The issue is described here [issues/319](https://github.com/vurtun/nuklear/issues/319).
    1.2  There is an interesting PR #873 that introduces a `have_mapbuffer` flag, but it is always `true`. As a temporary workaround, its value could be set based on the `__EMSCRIPTEN__` macro.
    1.3 This PR will **only fix demos** including `demo/glfw_opengl3`, but the our [examples](https://github.com/Immediate-Mode-UI/Nuklear/tree/master/example) (based on `GLFW/OpenGL3` too) **remain unchanged**.
2.  If we choose `sdl3_renderer` as the Emscripten backend, we cannot include [demo/common/file_browser.c](https://github.com/Immediate-Mode-UI/Nuklear/blob/master/demo/common/file_browser.c), since the `icon_load` uses direct OpenGL calls. This is already fixed in #590.

### Notes
1.  Some work on Emscripten support has already been done by @DeXP:\
    1.1. In the old repository, [this PR](https://github.com/vurtun/nuklear/pull/534) added `sdl_opengles2` and a `web` rule to `Makefile`.
    1.2. About 10 years ago, a [live version](https://github.com/DeXP/nuklear-webdemo) was created, but it is no longer maintained.
2.  I believe, for `sdl3_renderer`, it may also make sense to add a `web` rule in the `Makefile` to improve consistency.

sleeptightAnsiC:
(I'm well aware the text above is LLM-generated/edited)

[...]

PR #894, opened 2026-02-10
demo/sdl3_renderer: rewrite Makefile in order to support Emscripten

sleeptightAnsiC:
I wanted to fix this Makefile ~~and add a CI for it~~ for quite some time, and since there is some interest in Emscripten support (https://github.com/Immediate-Mode-UI/Nuklear/issues/888), I also added a support for this use-case. Read the corresponding commit message for more details.

This PR is split into 3 commits in order to be easier to review, but feel free to Squash it.

### DIFF
--- a/demo/sdl3_renderer/Makefile
+++ b/demo/sdl3_renderer/Makefile
@@ -23,28 +23,32 @@ BINEXT := ${binext_${OSNAME}}
 TEMPDIR := ./bin
 BIN := ${TEMPDIR}/demo${BINEXT}
 
-CFLAGS += -std=c89 -Wall -Wextra -Wpedantic
-CFLAGS += -O2
-#CFLAGS += -O0 -g
-#CFLAGS += -fsanitize=address
-#CFLAGS += -fsanitize=undefined
+cflags += -std=c89 -Wall -Wextra -Wpedantic
+cflags += -O2
+#cflags += -O0 -g
+#cflags += -fsanitize=address
+#cflags += -fsanitize=undefined
+cflags += ${CFLAGS}
 
-CPPFLAGS += $(shell ${PKG_CONFIG} ${PKG_SDL3} --cflags --keep-system-cflags)
+cppflags += $(shell ${PKG_CONFIG} ${PKG_SDL3} --cflags --keep-system-cflags)
+cppflags += ${CPPFLAGS}
 
-LDFLAGS += $(shell ${PKG_CONFIG} ${PKG_SDL3} --libs-only-L --libs-only-other --keep-system-libs)
+ldflags += $(shell ${PKG_CONFIG} ${PKG_SDL3} --libs-only-L --libs-only-other --keep-system-libs)
+ldflags += ${LDFLAGS}
 
-LDLIBS += $(shell ${PKG_CONFIG} ${PKG_SDL3} --libs-only-l --keep-system-libs)
-LDLIBS += -lm
+ldlibs += $(shell ${PKG_CONFIG} ${PKG_SDL3} --libs-only-l --keep-system-libs)
+ldlibs += -lm
+ldlibs += ${LDLIBS}
 # HACK: this one is for compatibility with other demos
-LDLIBS += ${LIBS}
+ldlibs += ${LIBS}
 
-DEP := ${BIN}.d
+DEP := ${TEMPDIR}/$(notdir ${BIN}).d
 
 SRC := main.c
 
 ${BIN}:
 	mkdir -p $(dir $@)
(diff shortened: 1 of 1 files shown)


sleeptightAnsiC:
Converting this to DRAFT as I want to retest it in few other environments.

If someone is wondering how to compile this thing with Emscripten: make sure you have Emscripten installed and active (either via package manager or EMSDK) and then do:
```sh
$ git clone https://github.com/sleeptightAnsiC/Nuklear.git --branch fix__sdl3renderer_makefile
$ cd nuklear/demo/sdl3_renderer
$ embuilder build sdl3
$ emmake make
$ emrun bin/demo.html
```
^ if everything is setup correctly, this should: compile Emscripten's SDL3 port and make it globally available via spoofed pkg-config, and then it will build the demo like usual under emmake.

This works on my Linux machine.

~~I also attempted to test it on Windows under many environments (MSYS2 UCRT64, MSYS2 CLANG64, w64devkit and scoop), but no matter how much I try, I always hit some issue with Emscripten (I even had a BSOD during compilation at some point... lol). So I just nuked my whole Windows VM and started setting up a new one. Would love, if someone could help retesting this.~~ EDIT: Pavel confirmed this is working under MSYS2 in https://github.com/Immediate-Mode-UI/Nuklear/pull/894#issuecomment-3880797791

I also won't be able to fully check it on MacOS, because even when you install Emscripten via Brew, it still requires Xcode (and I don't want to waste my time on setting it up right now). EDIT: but I retested it there and non-Emscripten build works fine.

sleeptightAnsiC:
@RobLoach seems like Ubuntu 24.04.3 LTS (the one that our CI uses) does not ship with SDL3.

Any ideas how to deal with this one? Can we simply bump our CI to newer Ubuntu release, or fetch SDL3 from somewhere else? If that's not possible, I will be fine with removing the commit that touches CI, but it would be nice to test this thing.

PavelSharp:
Actually, while discussing this in Discord DMs, we ran into an issue with MSYS2 path expansion. It looked pretty tricky at first. But a few days later this PR showed up, and the path expansion problem is no longer happening.
Currently, It works great in my environment (MSYS2/CLANG64), since I was able to build both the Emscripten version(by calling `emmake make`) and the native one.

RobLoach:
Since SDL3 uses CMake, using something like this should work across platforms:
https://github.com/Ravbug/sdl3-sample

Don't need all that the template provides, but it does have the basics available for both desktop and emscripten. We could have CMake download SDL, without needing the git submodules.

sleeptightAnsiC:
> We could have CMake download SDL, without needing the git submodules.

Wouldn't that kill our CI time? (considering that it would also have to compile it)

I also don't feel very comfortable with introducing a CMake only for fetching/building one dependency.

RobLoach:
Good point. Don't want slow build on the CI.

sleeptightAnsiC:
EDIT: reported a task for it: https://github.com/Immediate-Mode-UI/Nuklear/issues/895

(unrelated to this PR)

> > > We could have CMake download SDL, without needing the git submodules.
> >
> > Wouldn't that kill our CI time? (considering that it would also have to compile it)
>
> Good point. Don't want slow build on the CI.

btw, we also have few demos that build with `-O2` enabled by default (few SDL ones), which taxes the CI time a lot. I was about to report it, and do some cleanup on the demo flags, as currently we have quite a mess...

RobLoach:
Let's change those to `-O0` in a diff PR

sleeptightAnsiC:
@RobLoach How about I just remove the CI change from this PR, and we will reconsider it later?

I thought about this problem a bit, and I think the only "_correct_" solution is to bump CI to newer Ubuntu release. Obviously, we can't do this right now as some stuff may break, and such change simply doesn't belong here, but maybe in another PR? All other solutions, I don't believe it would be worth to use a different workflow only for supplying SDL3 binaries.

What do you think?

RobLoach:
Newer Ubuntu should be fine. Could use different workflows if we need older one still.

sleeptightAnsiC:
Moved the CI thing to a different issue. We will deal with it later https://github.com/Immediate-Mode-UI/Nuklear/issues/897

@PavelSharp Now, you should be able to provide your own C/CPP/LD flags at command line correctly, for example: `emmake make CFLAGS='-Os' LDFLAGS='--shell-file path/to/custom_shell.html'` Before, it could overwrite some flags required for SDL. Note that when multiple flags are passed into compiler, the latter one will take priority (e.g. in `emcc -O2 -Os` the active one is `-Os`)

@RobLoach Remember to squash this one during merge. Those commits are only for easier review. Unless you see any value in keeping them separated, squash+merge should be fine.

Issue #700, opened 2024-09-23
NK_STRTOD has extra const qualifier that blocks using standard library


Xeverous:
Nuklear's code:

```c
NK_API float
nk_strtof(const char *str, const char **endptr)
{
    float float_value;
    double double_value;
    double_value = NK_STRTOD(str, endptr);
    float_value = (float)double_value;
    return float_value;
}
```

The problem - standard library is incompatible:

```cpp
// C89
double strtod(const char *restrict str, char **restrict str_end);
// C++ (any version)
double strtod(const char* str, char** str_end);
```

If I `#define NK_STRTOD strtod` then I'm getting the following error:

```
[...]/Nuklear/nuklear.h: In function ‘float nk_strtof(const char*, const char**)’:
[...]/Nuklear/nuklear.h:6841:35: error: invalid conversion from ‘const char**’ to ‘char**’ [-fpermissive]
 6841 |     double_value = NK_STRTOD(str, endptr);
      |                                   ^~~~~~
      |                                   |
      |                                   const char**
/usr/include/stdlib.h:118:27: note:   initializing argument 2 of ‘double strtod(const char*, char**)’
  118 |         char **__restrict __endptr)
      |         ~~~~~~~~~~~~~~~~~~^~~~~~~~
```

Can you change the implementation so that it supports the standard library (without `const`)?

PROP65:
This should have been fixed in #682, but I forgot to rebuild nuklear.h at the time.

It's been rebuilt since, so this issue should be resolved.

Xeverous:
I have seen some commits changing this back and forth so I worry that someone later might open an issue that the function has no `const` but it should.

The root cause of the problem is that the standard library function is missing `const`. So Nuklear has to decide whether to go for const-correctness or stdlib compatibility (I would choose the latter). And ideally, Nuklear should apply this logic consistently to the entire implementation.

If Nuklear does not form specific consistent approach it risks 1) inconsistency in API 2) back-and-forth issues to "fix" the code to the other way.

PROP65:
I'm not sure why `endptr` was made `const` to begin with (I checked the git blame and I belive Cong made the change in `2e4db87` after vurtun fixed it in `f675cea`), but the function needs to write to `endptr` to behave correctly. The incorrect `const` was there for at least 7 years, but I guess no one had tried utilizing that behavior.

`strtof` is from c99 but is just a float version of the c89 function `strtod`.

[...]

Issue #699, opened 2024-09-23
NK_DTOA needs more documentation


Xeverous:
The current documentation states:

> You can define this to dtoa or your own double to string conversion implementation replacement. If not defined nuklear will use its own imprecise and possibly unsafe version (does not handle nan or infinity!).

The problem is, there is no such function. `dtoa` is neither in C standard library, C++ standard library or in POSIX. I just don't know what is the "contract" of this function.

Based on Nuklear's code:

https://github.com/Immediate-Mode-UI/Nuklear/blob/ca8aaf3f65e9fc8ac5d90fefc1204585add4b89b/nuklear.h#L7116C1-L7117C27

The function takes a buffer and a value of type `double` and returns the end of the string in the buffer.

*How large the buffer should be?* I checked the implementation and found out it is `NK_MAX_NUMBER_BUFFER`. This constant is documented but it is easy to lose it - NK_DTOA does not mention it and does not even specify how the function should look like (and there is no standard function to look into).

IMO it should be documented that:
- `NK_DTOA` relies on `NK_MAX_NUMBER_BUFFER`
- what the expected function prototype is
- what the expected behavior is (e.g. returns start or end of buffer)

hanatos:
fwiw the return value seems to be unused in `nuklear.h`. since there are cases where `nk_dtoa` gets caught in an infinite loop (didn't have time to reproduce and fix, sorry), i replaced it like so:
```
#define NK_DTOA(S, D) sprintf(S, "%g", D)
#include "nuklear.h"
```

sleeptightAnsiC:
~Maintainer note: might be nice to revisit this when dealing with https://github.com/Immediate-Mode-UI/Nuklear/issues/415 (something in my guts tells me those two are related)~ EDIT: Sh1t... I think I pasted wrong link when writing this.

> [...] i replaced it like so:
> 
> ```
> #define NK_DTOA(S, D) sprintf(S, "%g", D)
> ```

nice hack btw :D

sleeptightAnsiC:
I did some research on `dtoa()` recently so dumping all my thought here:
- found lots of useful info in this benchmark: https://github.com/miloyip/dtoa-benchmark
- this commit message may give some insight: https://github.com/Immediate-Mode-UI/Nuklear/commit/f675cea3057f1bfd0cb544a8516a123fd410229e
[...]

PR #898, opened 2026-02-16
Fix previous property focus issues

rswinkle:
Rebase of #688.  Not sure what the (char*) cast is for but other than that it seems to solve both problems it set out to, Thanks @xzn!

EDIT: forgot to tag you @RobLoach

### DIFF
--- a/nuklear.h
+++ b/nuklear.h
@@ -5773,6 +5773,10 @@ struct nk_property_state {
     unsigned int seq;
     unsigned int old;
     int state;
+    int prev_state;
+    nk_hash prev_name;
+    char prev_buffer[NK_MAX_NUMBER_BUFFER];
+    int prev_length;
 };
 
 struct nk_window {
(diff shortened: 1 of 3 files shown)


src/nuklear_util.c
```diff
@@ -217,7 +217,7 @@ nk_strtod(const char *str, char **endptr)
     }
     number = value * neg;
     if (endptr)
-        *endptr = p;
+        *endptr = (char *)p;
```
sleeptightAnsiC, line 220: > Not sure what the (char*) cast is for

Seems like a leftover from when this function had wrong `const` in the arguments in the past. I remember seeing several Issues reported about it (e.g. https://github.com/Immediate-Mode-UI/Nuklear/issues/700)

Line 171 outcasts the const from `str`, and then assigns the offset from that pointer to `*endptr` here, loosing the const semantic in whole process (btw, feels like a bug, right?)

I think, this cast is redundant now. It may made sense back when https://github.com/Immediate-Mode-UI/Nuklear/pull/688 was submitted, but not anymore. Does the lack of cast give you any warning?

sleeptightAnsiC: (not related to this PR) 

> Line 171 outcasts the const from str, and then assigns the offset from that pointer to *endptr here, loosing the const semantic in whole process (btw, feels like a bug, right?)

I feel like the synopsis of nk_strtod is still wrong, or we're still doing something wrong in its implementation. nk_dtoa has similar issues (https://github.com/Immediate-Mode-UI/Nuklear/issues/699)

RobLoach: Let's handle any follow ups elsewhere. Thanks!

RobLoach:
Merging this one. Thanks for the rebase and clean up. Anything that's missing, we can handle in a follow up. The cast seems fine, feel free to send a PR if it's not needed.

Issue #899, opened 2026-02-17
add a way to keep certain widgets inside of visible area


sleeptightAnsiC:
When talking about _"screen area"_, what we really have in mind is "_the backbuffer that Nuklear is drawing to_". Not to be confused with nk_window or the physical display! (I feel like I only made it more confusing now...)

Follow up from https://github.com/Immediate-Mode-UI/Nuklear/pull/881#discussion_r2746144225

> > It also would be really nice for the tooltip to stay in the available screen space. Right now it just draws outside so you can't see its contents [screenshot below], however, I'm not sure if this problem is solvable right now (like does Nuklear keep the OS Window size somewhere?)
[...]

PR #900, opened 2026-02-17
Add tooltip positioning V2

sleeptightAnsiC:
Version 2 (or rather an alternative) of https://github.com/Immediate-Mode-UI/Nuklear/pull/881

Noticeable differences from the original:
- use bitmask flags instead of enums (https://github.com/Immediate-Mode-UI/Nuklear/pull/881#discussion_r2749267957)
- tooltip related symbols have prefix (https://github.com/Immediate-Mode-UI/Nuklear/pull/881#discussion_r2777453052)
- positioning is flipped (relative to cursor as user sees it, not to screen coordinates)
- offsets are relative to positioning (can be absolute after setting the flag, by default the offset feels more like a "_distance_" to the tooltip's corner/edge) **but maybe we should make it absolute by default??**
- added some reasonable offset to the default tooltip, so even the old code benefits from this change
- removed the big duplicated code chunk (https://github.com/Immediate-Mode-UI/Nuklear/pull/881#discussion_r2816291069)
- moved the offscreen problem to another task (https://github.com/Immediate-Mode-UI/Nuklear/issues/899)
- added tooltip editor to demo/overview.c so you can play around a bit

And probably more... I'll need to polish this a bit (code looks way too raw) so submitting as DRAFT for now.

I think I may have overcooked this one, but I really wanted to move this out of my backlog, so we can reiterate on all the ideas as fast as possible. Apologies for this thing taking so long on my side. There are many other problems related to Tooltips/Popups that I wanted to revisit with this work.

### DIFF
--- a/demo/common/overview.c
+++ b/demo/common/overview.c
@@ -839,12 +839,85 @@ overview(struct nk_context *ctx)
                 } else popup_active = nk_false;
             }
 
-            /* tooltip */
-            nk_layout_row_static(ctx, 30, 150, 1);
+            /* tooltips */
+            nk_layout_row_static(ctx, 30, 400, 1);
             bounds = nk_widget_bounds(ctx);
-            nk_label(ctx, "Hover me for tooltip", NK_TEXT_LEFT);
-            if (nk_input_is_mouse_hovering_rect(in, bounds))
-                nk_tooltip(ctx, "This is a tooltip");
+            nk_label(ctx, "Hover for default tooltip", NK_TEXT_LEFT);
+            if (nk_input_is_mouse_hovering_rect(in, bounds)) {
+                nk_tooltip(ctx, "This is very boring default tooltip...");
+            }
+            bounds = nk_widget_bounds(ctx);
+            nk_label(ctx, "Hover for Gnome-like tooltip", NK_TEXT_LEFT);
+            if (nk_input_is_mouse_hovering_rect(in, bounds)) {
+                struct nk_vec2 offset = { 0, 15 };
+                nk_tooltip_offset(ctx, "Gnome centers above cursor with greater y offset", NK_TOOLTIP_ABOVE, offset);
+            }
+            bounds = nk_widget_bounds(ctx);
+            nk_label(ctx, "Hover for above-on-left tooltip", NK_TEXT_LEFT);
+            if (nk_input_is_mouse_hovering_rect(in, bounds)) {
+                struct nk_vec2 offset = { 0, 0 };
+                nk_tooltip_offset(ctx, "above-on-left from cursor with 0:0 offset", NK_TOOLTIP_ABOVE|NK_TOOLTIP_ON_LEFT, offset);
+            }
+            bounds = nk_widget_bounds(ctx);
+            nk_label(ctx, "Hover for MAGIC!", NK_TEXT_LEFT);
+            if (nk_input_is_mouse_hovering_rect(in, bounds)) {
+                static double accum_time_seconds = 0.0;
+                const double speed = 3.0, radius = 50.0;
+                struct nk_vec2 offset;
+                offset.x = radius * cos(accum_time_seconds * speed);
+                offset.y = radius * sin(accum_time_seconds * speed);
+                nk_tooltip_offset(ctx, "WOW!", NK_TOOLTIP_ABS_OFFSET, offset);
+                accum_time_seconds += (double)(ctx->delta_time_seconds);
+            }
+
+            /* editor for custom tooltip */
+            {
+                static char text_buf[64] = {0};
+                static int text_len = 0;
+                static int text_initialized = 0;
+                static nk_flags tooltip = 0;
+                static struct nk_vec2 offset = {0};
+                if (!text_initialized) {
+                    const char text_default[] = "you can customize this!";
+                    NK_ASSERT(sizeof(text_default) < sizeof(text_buf));
+                    memcpy(text_buf, text_default, sizeof(text_default));
+                    text_len = sizeof(text_default) - 1;
+                    text_initialized = 1;
+                }
+                bounds = nk_widget_bounds(ctx);
+                nk_label(ctx, "Hover for custom tooltip (you can customize it below)", NK_TEXT_LEFT);
+                if (nk_input_is_mouse_hovering_rect(in, bounds)) {
+                    nk_tooltip_offset(ctx, text_buf, tooltip, offset);
+                }
+                nk_layout_row_dynamic(ctx, 1, 1);
+                nk_rule_horizontal(ctx, nk_white, nk_true);
+                nk_layout_row_dynamic(ctx, 30, 2);
+                nk_label(ctx, "custom tooltip text:", NK_TEXT_LEFT);
+                nk_edit_string(ctx, NK_EDIT_FIELD, text_buf, &text_len, sizeof(text_buf), nk_filter_default);
+                NK_ASSERT(text_len < (int)sizeof(text_buf));
+                text_buf[text_len] = '\0';  /* TODO: why nk_edit_string is NOT setting this on its own? */
+                nk_layout_row_dynamic(ctx, 30, 1);
+            #define TOOLTIP_FOREACH_FLAG(BODY) \
+                    BODY(ABOVE)                \
+                    BODY(BELOW)                \
+                    BODY(ON_LEFT)              \
+                    BODY(ON_RIGHT)             \
+                    BODY(ABS_OFFSET)
+            #define TOOLTIP_CHECKBOX_FLAG(FLAG)                                                                          \
+                {                                                                                                        \
+                    nk_bool checked = !!(tooltip & NK_TOOLTIP_##FLAG);                                                   \
+                    nk_checkbox_label_align(ctx, "custom tooltip flag: " #FLAG, &checked, NK_WIDGET_RIGHT, NK_TEXT_LEFT);\
+                    tooltip = checked ? (tooltip | NK_TOOLTIP_##FLAG) : (tooltip & ~NK_TOOLTIP_##FLAG);                  \
+                }
+            TOOLTIP_FOREACH_FLAG(TOOLTIP_CHECKBOX_FLAG)
+            #undef TOOLTIP_FOREACH_FLAG
+            #undef TOOLTIP_CHECKBOX_FLAG
+                nk_layout_row_dynamic(ctx, 30, 2);
+                nk_label(ctx, "custom tooltip offset", NK_TEXT_LEFT);
+                nk_property_float(ctx, "x", -100.0f, &offset.x, 100.0f, 5.0f, 0.5f);
+                nk_label(ctx, "custom tooltip offset", NK_TEXT_LEFT);
+                nk_property_float(ctx, "y", -100.0f, &offset.y, 100.0f, 5.0f, 0.5f);
+            }
 
             nk_tree_pop(ctx);
         }
--- a/src/nuklear_tooltip.c
+++ b/src/nuklear_tooltip.c
@@ -6,10 +6,40 @@
  *                              TOOLTIP
  *
  * ===============================================================*/
+NK_LIB struct nk_vec2
+nk_tooltip_get_default_offset(const struct nk_context *ctx)
+{
+    struct nk_vec2 offset = {0};
+    NK_ASSERT(ctx);
+    if (!ctx) return offset;
+    if (ctx->style.cursor_active) {
+        /* nuklear is drawing its own cursor so we can reuse its size (best case!) */
+        offset.x = ctx->style.window.padding.x + ctx->style.cursor_active->size.x;
+        offset.x = ctx->style.window.padding.y + ctx->style.cursor_active->size.y;
+    } else if (ctx->style.font) {
+        /* assume that cursor size is similar to font height (flawed but reasonable)*/
+        offset.y = ctx->style.window.padding.x + ctx->style.font->height;
+        offset.x = ctx->style.window.padding.y + ctx->style.font->height;
+    }
+    return offset;
+}
+NK_LIB nk_flags
+nk_tooltip_get_default_flags(const struct nk_context *ctx)
+{
+    NK_UNUSED(ctx);
+    return NK_TOOLTIP_BELOW|NK_TOOLTIP_ON_RIGHT;
+}
 NK_API nk_bool
 nk_tooltip_begin(struct nk_context *ctx, float width)
 {
-    int x,y,w,h;
+    return nk_tooltip_begin_offset(ctx, width,
+                                   nk_tooltip_get_default_flags(ctx),
+                                   nk_tooltip_get_default_offset(ctx));
+}
+NK_API nk_bool
+nk_tooltip_begin_offset(struct nk_context *ctx, float width, nk_flags tooltip, struct nk_vec2 offset)
+{
+    int x,y,w,h,mul_x,mul_y;
     struct nk_window *win;
     const struct nk_input *in;
     struct nk_rect bounds;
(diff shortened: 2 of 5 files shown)


sleeptightAnsiC:
EDIT: swapped video with the one from never commit

https://github.com/user-attachments/assets/ae6c548b-6b3f-4d85-a0ed-fa2511dc3d48

demo/common/overview.c
```diff
@@ -839,12 +839,85 @@ overview(struct nk_context *ctx)
                 } else popup_active = nk_false;
             }
 
-            /* tooltip */
-            nk_layout_row_static(ctx, 30, 150, 1);
+            /* tooltips */
+            nk_layout_row_static(ctx, 30, 400, 1);
             bounds = nk_widget_bounds(ctx);
-            nk_label(ctx, "Hover me for tooltip", NK_TEXT_LEFT);
-            if (nk_input_is_mouse_hovering_rect(in, bounds))
```
sleeptightAnsiC, line 846: I was wondering, how do we fix the bug, where hovering a rect always triggers the tooltip to appear, even though there is another nk_window above it? [screenshot below]

I did not attempt to fix it as I thought maybe you guys know a more reliable way to trigger these? My idea was to simply check, if this branch runs on currently active nk_window, but I'm not so sure about it.

Image

EDIT: for comparison, "chart" does not have this issue because it activates tooltip based on active chart index:
<p float="left">
image
image
</p>

src/nuklear_tooltip.c
```diff
@@ -28,14 +58,34 @@ nk_tooltip_begin(struct nk_context *ctx, float width)
...
+    /* turn relative offset into absolute one, unless it's already absolute */
+    if ((tooltip & NK_TOOLTIP_ON_LEFT) && !(tooltip & NK_TOOLTIP_ABS_OFFSET))
+        offset.x *= -1;
+    if ((tooltip & NK_TOOLTIP_ABOVE)   && !(tooltip & NK_TOOLTIP_ABS_OFFSET))
+        offset.y *= -1;
+
+    x = 0;
+    x -= !!(tooltip & NK_TOOLTIP_ON_LEFT );
+    x += !!(tooltip & NK_TOOLTIP_ON_RIGHT);
+    NK_ASSERT(x == -1 || x == 0 || x == 1);
```
sleeptightAnsiC: EDIT: fixed in https://github.com/Immediate-Mode-UI/Nuklear/compare/db6f6baf6a901611539fef3d6a9006a64bc938fe..0d62612da12fae269a9bd45796c6c6d33213cf13

probably needs a temp variable instead of reusing `x` all over the place
same for `y` on lines 77-80

src/nuklear_tooltip.c
```diff
@@ -28,14 +58,34 @@ nk_tooltip_begin(struct nk_context *ctx, float width)
...
+    NK_ASSERT(y == -1 || y == 0 || y == 1);
+    y = -h/2 + (y * h/2);
+    y += nk_ifloorf(in->mouse.pos.y + 1) - (int)win->layout->clip.y;
+    y += offset.y;
 
     bounds.x = (float)x;
     bounds.y = (float)y;
     bounds.w = (float)w;
-    bounds.h = (float)h;
+    bounds.h = (float)nk_iceilf(nk_null_rect.h);
```
sleeptightAnsiC, line 92: I have no idea why this was changed during https://github.com/Immediate-Mode-UI/Nuklear/pull/881 but I just kept it order to prevent incompatibilities between those PR.

same for `h` assignment on line 61

demo/common/overview.c
```diff
@@ -839,12 +839,85 @@ overview(struct nk_context *ctx)
...
+                    BODY(ABS_OFFSET)
+            #define TOOLTIP_CHECKBOX_FLAG(FLAG)                                                                          \
+                {                                                                                                        \
+                    nk_bool checked = !!(tooltip & NK_TOOLTIP_##FLAG);                                                   \
+                    nk_checkbox_label_align(ctx, "custom tooltip flag: " #FLAG, &checked, NK_WIDGET_RIGHT, NK_TEXT_LEFT);\
+                    tooltip = checked ? (tooltip | NK_TOOLTIP_##FLAG) : (tooltip & ~NK_TOOLTIP_##FLAG);                  \
+                }
+            TOOLTIP_FOREACH_FLAG(TOOLTIP_CHECKBOX_FLAG)
+            #undef TOOLTIP_FOREACH_FLAG
+            #undef TOOLTIP_CHECKBOX_FLAG
```
sleeptightAnsiC, line 914: this XMACRO is a bit overcomplicated. I wanted to save on space. Not sure, if it was worth it. (note: this can be changed to array and for-loop, maybe? but it would be a ton of code)

src/nuklear_tooltip.c
```diff
@@ -28,14 +58,34 @@ nk_tooltip_begin(struct nk_context *ctx, float width)
         return 0;
 
     w = nk_iceilf(width);
-    h = nk_iceilf(nk_null_rect.h);
-    x = nk_ifloorf(in->mouse.pos.x + 1) - (int)win->layout->clip.x;
-    y = nk_ifloorf(in->mouse.pos.y + 1) - (int)win->layout->clip.y;
+    h = ctx->current->layout->row.min_height;
+
+    /* turn relative offset into absolute one, unless it's already absolute */
+    if ((tooltip & NK_TOOLTIP_ON_LEFT) && !(tooltip & NK_TOOLTIP_ABS_OFFSET))
```
sleeptightAnsiC: EDIT: fixed in https://github.com/Immediate-Mode-UI/Nuklear/compare/db6f6baf6a901611539fef3d6a9006a64bc938fe..0d62612da12fae269a9bd45796c6c6d33213cf13

todo: add a condition for `&& !(tooltip & NK_TOOLTIP_ON_RIGHT)`, and a similar one on line 66. Right now, the flip triggers when both flags are set and negating each other, which is a rare corner-case, but I don't believe should happen.

demo/common/overview.c
rswinkle: This is similar to the bug where a scrollbar will activate (when you have AUTO_HIDE on) even if your mouse is in a popup on in front of the window.

rswinkle:
> * use bitmask flags instead of enums ([Add tooltip positioning #881 (comment)](https://github.com/Immediate-Mode-UI/Nuklear/pull/881#discussion_r2749267957))

Still not a fan. I think the code is more complicated/harder to read and the ability to put in enums that contradict or cancel each other out is confusing. Also the fact that getting the normal/default behavior takes 2, even if most won't be calling it directly.

> * tooltip related symbols have prefix ([Add tooltip positioning #881 (comment)](https://github.com/Immediate-Mode-UI/Nuklear/pull/881#discussion_r2777453052))

I'm fine with adding a prefix if we must but as I said I'd just rather only have a single set of alignment constants and prefix plus bitmask makes the length of arguments even worse

> * positioning is flipped (relative to cursor as user sees it, not to screen coordinates)

This is definitely confusing, especially when combined with point one above. Offsets should be the same regardless of the origin. I think sticking with screen coordinates is fine (-y offset means tooltip moves up) but whatever we choose it should be consistent across any starting position. Just as an example, ON_LEFT|ABS vs ON_RIGHT|ABS vs ON_LEFT|ON_RIGHT|ABS vs ABS vs 0. You can say, well they shouldn't use anything with ABS or combine contradictory masks, or have a non-zero offset without an origin, but the fact is they can, and the behavior is confusing. I understand what you were going for but simplicity is better.

> * offsets are relative to positioning (can be absolute after setting the flag, by default the offset feels more like a "_distance_" to the tooltip's corner/edge) **but maybe we should make it absolute by default??**

This is not a difference with my version. My offsets are relative to the origin selected as well. The difference is everything in the above point and you've picked a "default" origin of the center for your "absolute" offsets; again you've made it confusing by making your offset meaning change based on your origin as well.

> * added some reasonable offset to the default tooltip, so even the old code benefits from this change
> * removed the big duplicated code chunk ([Add tooltip positioning #881 (comment)]

I'm fine with these two, I'm happy too make the same changes to mine as long as we agree on a default position/offset

> * moved the offscreen problem to another task ([add a way to keep certain widgets inside of visible area #899](https://github.com/Immediate-Mode-UI/Nuklear/issues/899))

Agreed. Sadly, how we solve that impacts how we implement this a bit so we should probably come to some decision there before merging either of these.

> * added tooltip editor to demo/overview.c so you can play around a bit

I definitely like this for showing what can be done and how it works though the macro system is a bit much as you said. I could definitely add the same to mine so we'd have an easier time comparing behavior.

> And probably more... I'll need to polish this a bit (code looks way too raw) so submitting as DRAFT for now.
> I think I may have overcooked this one, but I really wanted to move this out of my backlog, so we can reiterate on all the ideas as fast as possible. Apologies for this thing taking so long on my side. There are many other problems related to Tooltips/Popups that I wanted to revisit with this work.

No worries, we've probably all got other projects and jobs/life that take precedence. I'll update mine as soon as I can and think about the offscreen issue as well. I'll add a CENTER option

PS, I did run into some weird behavior. When I clicked on a property, I got focus and a cursor on your edit string as well and have to click the property again to actually have it accept input.  Very strange.  
Screenshot from 2026-02-18 17-49-33

sleeptightAnsiC:
**(I'm intentionally NOT responding to everything. I will address all those points after thinking a bit)**

I believe the current "_smart relative offset by default_" gives a horrible first impression for V2. I should definitely make it optional and maybe even redesign this part. I will most likely ask you for retest after that.

My understanding was that people would usually keep the offset the same, and only flip the positioning based on some pre condition (e.g. if your static widget is in the top-left corner of the screen all the time, you probably want your tooltip to appear on the bottom-right all the time, and so on...)

> > moved the offscreen problem to another task (https://github.com/Immediate-Mode-UI/Nuklear/issues/899)
>
> Agreed. Sadly, how we solve that impacts how we implement this a bit so we should probably come to some decision there before merging either of these.

^ THIS! feels like the most important thing we need to discuss. The whole reason why my V2 bitmask exists and why I don't agree with enum approach is because I don't really see a way to extend it. When we find the solution for detecting offscreen, I can just add another bitmask flag, but with enum I don't really see a different solution than introducing new function permutation or causing a breaking-change (which makes no sense given the fact that we're about to solve this problem soon). 

Another thing that we seriously need to discuss is: _How the tooltip should behave in the offscreen scenario?_ I can only imagine two behaviors, either (a) tooltip flips to the entirely new positioning on the other side of cursor, or (b) the position is clamped to the screen dimensions and doesn't care about cursor at all.

(a) seems hard to implement right, but I think it is possible. Basically: if the first pass detects that x/y/w/h are outside of the screen, we run the same code again, but with flipped flags and negated offset. I don't want the tooltip to abnormally flicker or do something strange, but I'm not entirely sure how to avoid it.

(b) seems easy but it will cause the situations where cursor appears over the tooltip. Dead simple to implement: if x/y/w/h is outside the screen, clamp it inside, and remember to apply the same diff to corresponding x/y/w/h (example: if x is -10, move x to 0 and move w to w+10)

I guess with the bitmask we can implement both and even let it coexist. With both enabled it could clamp until it runs out of available offset, and only then it would trigger the flip. (makes sense? I'll need to experiment with this)

@rswinkle btw, would you mind turning V1 (https://github.com/Immediate-Mode-UI/Nuklear/pull/881) into a DRAFT? I feel like Rob just wants to get it in and ignores most of our conversation. Given that we're dealing with two versions right now may confuse him even further.

Also, feel free to scrap as much code from here as you want. I would be happy if you could mark me as coauthor (but I don't mind if you won't)

demo/common/overview.c
```diff
@@ -839,12 +839,85 @@ overview(struct nk_context *ctx)
...
+                if (nk_input_is_mouse_hovering_rect(in, bounds)) {
+                    nk_tooltip_offset(ctx, text_buf, tooltip, offset);
+                }
+                nk_layout_row_dynamic(ctx, 1, 1);
+                nk_rule_horizontal(ctx, nk_white, nk_true);
+                nk_layout_row_dynamic(ctx, 30, 2);
+                nk_label(ctx, "custom tooltip text:", NK_TEXT_LEFT);
+                nk_edit_string(ctx, NK_EDIT_FIELD, text_buf, &text_len, sizeof(text_buf), nk_filter_default);
+                NK_ASSERT(text_len < (int)sizeof(text_buf));
+                text_buf[text_len] = '\0';  /* TODO: why nk_edit_string is NOT setting this on its own? */
```
sleeptightAnsiC, line 898: (from bottom of https://github.com/Immediate-Mode-UI/Nuklear/pull/900#issuecomment-3924069488)
> PS, I did run into some weird behavior. When I clicked on a property, I got focus and a cursor on your edit string as well and have to click the property again to actually have it accept input. Very strange.

I think I've also seen this. I'm not entirely sure where this issue comes from (maybe two widgets get the same hash?) but given the fact that I was already confused by nk_edit_string behavior (the todo note), maybe I did something wrong here.

demo/common/overview.c
sleeptightAnsiC: > a scrollbar will activate (when you have AUTO_HIDE on) even if your mouse is in a popup on in front of the window

Hmn... Tried to reproduce this with nk_window that comes with demo/overview but I couldn't. Though the scrollbar does not seem to be hiding as it should, so maybe I'm doing something wrong.

However, this does feel like the same bug. I've checked the code and it seems to be doing a similar thing: https://github.com/Immediate-Mode-UI/Nuklear/blob/3bcdec202920d5f3e8251c14af20a233a1590e2a/src/nuklear_scrollbar.c#L29-L30

If you could provide me with a repro I would just move it into a different task. This really feels like a bigger problem.

sleeptightAnsiC:
@rswinkle I just got an idea

> > > moved the offscreen problem to another task (https://github.com/Immediate-Mode-UI/Nuklear/issues/899)
> >
> > Agreed. Sadly, how we solve that impacts how we implement this a bit so we should probably come to some decision there before merging either of these.
>
> ^ THIS! [...] The whole reason why my V2 bitmask exists and why I don't agree with enum approach is because I don't really see a way to extend it. When we find the solution for detecting offscreen, I can just add another bitmask flag [...]

How about we abandon the approach of creating new `nk_tooltip_*` altogether?
**Instead, how about we introduce a new styling struct specific to tooltip?**
I feel like having a function permutation will never be enough,
and I'll be fine with having enum inside of styling struct.

It would help when changing the default tooltip globally, which isn't possible right now.
We could extend the struct in the future, so fixing off-screen-behavior won't be a problem.
The offset will be simply a member of the struct. Same with other options.
We will reuse the widget alignment flags for positioning, just like you wanted.
Instead of having tooltip editor in demo/overview.c, we will have one in demo/style_configurator.c
Tooltip right now inherits most of its look from nk_window, which doesn't seem correct, so having dedicated tooltip styling makes sense, and someone will probably want one in the future (right?)
Bunch of other widgets already use this approach.

Right now I only see two problems with this idea: When creating tooltips with custom behavior, you need to nk_style_push_* (and pop) but whether this interface is good or bad is a matter of preference. Another problem is that `nk_style_window` already has 3 members that seem to be related to tooltip styling (`tooltip_border_color`, `tooltip_border`, `tooltip_padding`) however, I couldn't find any code that would read from those members, I grepped a bunch and this seems like dead code ([proof here](https://gist.github.com/sleeptightAnsiC/7c7dbc0453dbdb3a8a59be779598792e)) so we can just ignore those.

That was a very spontaneous idea of mine, but unless I missed something important that would deem it impossible to implement, I really think this might be the way. If you find this interesting, I can close V2 and start over with V3.

rswinkle:
> @rswinkle I just got an idea
> How about we abandon the approach of creating new `nk_tooltip_*` altogether? **Instead, how about we introduce a new styling struct specific to tooltip?** I feel like having a function permutation will never be enough, and I'll be fine with having enum inside of styling struct.
> 
> It would help when changing the default tooltip globally, which isn't possible right now. We could extend the struct in the future, so fixing off-screen-behavior won't be a problem
> That was a very spontaneous idea of mine, but unless I missed something important that would deem it impossible to implement, I really think this might be the way. If you find this interesting, I can close V2 and start over with V3.

I think that sounds like a good idea. I can't think of very many cases where developers wouldn't want consistent tooltip appearance/behavior across their app so the ability to set it globally once on startup is appealing and we can always have functions to set the positioning directly for an individual tooltip if needed, even if they're just wrappers around push/pops.

I'll still update mine but other implementation details, this does sound a better idea in the long run, and as you found there's some dead code in there that should be cleaned out/updated anyway.

demo/common/overview.c
rswinkle: I just tested and I get the bug at least with the sdl2 and sd3_renderer backends with the standard demo. Here's a video of the sdl2 renderer, sorry it doesn't record my cursor but I swear I keep it in the menu popup and the second window for the test.

rswinkle: https://github.com/user-attachments/assets/f49cbe06-a6a6-4443-a594-61ad1214af48

Forgot the video, oops.

sleeptightAnsiC: Few notes:

There is `nk_widget_is_hovered` but from what I see it needs active layout (?) and grepping for this shows no use inside of the repo. Strange...

https://github.com/Immediate-Mode-UI/Nuklear/blob/729516425ae272809736c15eadd240a904f98f7e/src/nuklear_widget.c#L68-L90

On the other hand, the chart is storing its own hover state based on `NK_WINDOW_ROM` (which few other widgets also seem to be doing) and given how this flag was documented, this seems to be a hack, like it expects that active windows will always have this flag NOT set (which is correct but hacky) 
https://github.com/Immediate-Mode-UI/Nuklear/blob/729516425ae272809736c15eadd240a904f98f7e/src/nuklear.h#L5509
https://github.com/Immediate-Mode-UI/Nuklear/blob/729516425ae272809736c15eadd240a904f98f7e/src/nuklear_chart.c#L173-L181

I think, this is a refactor angle... Right now, every widget is trying to reimplement the wheel...

RobLoach:
Is there anything missing to get this going?

sleeptightAnsiC:
> Is there anything missing to get this going?

@RobLoach Only my free time :)

I had few more severe issues noted that I wanted to take care of before revisiting this.

sleeptightAnsiC:
https://github.com/Immediate-Mode-UI/Nuklear/pull/911 https://github.com/Immediate-Mode-UI/Nuklear/issues/974 https://github.com/Immediate-Mode-UI/Nuklear/pull/975

Issue #897, opened 2026-02-12
Ubuntu release used by our CI is too old for fetching SDL3


sleeptightAnsiC:
Follow up from https://github.com/Immediate-Mode-UI/Nuklear/pull/894:

> > > > > > seems like Ubuntu 24.04.3 LTS (the one that our CI uses) does not ship with SDL3.  Any ideas how to deal with this one? Can we simply bump our CI to newer Ubuntu release, or fetch SDL3 from somewhere else?
> > > > > 
> > > > > Since SDL3 uses CMake, using something like this should work across platforms: https://github.com/Ravbug/sdl3-sample [...] We could have CMake download SDL, without needing the git submodules.
> > > > 
> > > > Wouldn't that kill our CI time? (considering that it would also have to compile it) I also don't feel very comfortable with introducing a CMake only for fetching/building one dependency.
> > > 
> > > Good point. Don't want slow build on the CI.
> > 
> > I thought about this problem a bit, and I think the only "correct" solution is to bump CI to newer Ubuntu release. Obviously, we can't do this right now as some stuff may break [...] but maybe in another PR? All other solutions, I don't believe it would be worth to use a different workflow only for supplying SDL3 binaries. What do you think?
>
> Newer Ubuntu should be fine. Could use different workflows if we need older one still.

I'll probably do exactly what Rob suggested, unless our demos are fine with newer GCC and all other packages in latest Ubuntu.

Another solution would be to "_cache_" all SDL3 artifacts. Not sure how Github runners do this but I have a feeling this should be possible. Though, I would prefer to use the very same workflow for all dependencies.

sleeptightAnsiC:
caching seems possible, and this Action is officially supported - https://github.com/libsdl-org/setup-sdl - but I have few notes:
- it should be pinned to the oldest stable SDL3 release (3.2)
- we don't need SDL_ttf/SDL_image/etc, we only need core library

still, updating Ubuntu release might be better in long run, because someone will have to do this eventually

PR #902, opened 2026-02-18
Webversion introduction and integration

PavelSharp:
Hi everyone! After a couple of weeks of work, I'm happy to propose a webversion of our Nuklear library. The process wasn't completely smooth, I had to fight through bugs and tried to be as thorough as possible. Ironically, this led to quite a few changes in the repository, so I decided to open this PR offering a **basic web integration**. Any additional ideas or improvements can be added later.

Try it with [showcase shell](https://pavelsharp.github.io/Nuklear/webver/). Try it with [minimal shell](https://pavelsharp.github.io/Nuklear/webver/minshell.html) And yes, it's also [in the docs](https://pavelsharp.github.io/Nuklear/) now. 

Key features
------------
1.  Uses SDL3 and Emscripten.
1.  **Works on mobile devices**, correctly handling scale and DPI (exceptions noted below).
1.  Comes with two shells: `index.html` and `minshell.html`.
1.  Added a `webver` target in the Makefile to build the webversion locally.
1.  **Automatic webversion updates** on push (Github Actions).
1.  **No changes to core**, no `#ifdef __EMSCRIPTEN__` hacks added.

Known issues
------------
1.  **Safari on Apple devices** doesn't work fully. On Mac, the `minshell` renders but then freezes and stops responding to events. We probably have a chance to fix it for desktop.
1.  **Text input on browsers by OSK** is currently unsupported in [SDL3 7348](https://github.com/libsdl-org/SDL/issues/7348).
1.  **Browser hotkeys** don't work (including `CTRL+R`), but you can focus the address bar as a workaround.
1.  `"Unhandled scale event! Nuklear may appear blurry"` appears on page load. This hasn't been fully investigated yet. Oddly, it doesn't happen on a local server.
1.  ~~**Orientation changes in fullscreen**~~.  Already fixed in [SDL3 15024](https://github.com/libsdl-org/SDL/issues/15024).

Mini FAQ
------------

Q: **Why use the SDL3 backend here?**
<details>
<summary>Answer</summary>

At first glance, this might seem not very minimalistic and a bit against the Nuklear spirit. However, we didn’t have a well-working backend for Emscripten yet, testing reports are available in #888. By luck, our SDL3 backend uses the high-level `SDL3_Renderer` instead of a lower-level GPU API like OpenGL or Vulkan. Early on, this was already very attractive and saved us from many compatibility issues.

As I worked through this, it became clear that SDL3 was the only right choice, because it’s the only backend that can request the virtual text keyboard. Ironically, this feature is still not implemented under Emscripten, but I really like the potential here. It shows how broadly cross-platform Nuklear can be, from mobile devices all the way to desktop.
</details>

Q: **Why are there two shells?**
<details>
<summary>Answer</summary>

The web integration was also discussed on [our Discord channel](https://github.com/Immediate-Mode-UI/Nuklear/wiki#community), where community members suggested keeping **two pages**:
1.  A **standalone page**, fully independent and attractive for newcomers.
1.  A page **embedded in the Doxygen docs** via an `iframe`.

I made some progress with embedding in Doxygen, but in the end I decided **not to include it in this PR**, to keep it small. Also, the Doxygen template has a large header and doesn't allow an `iframe` to take 100% height. A fixed height would lead to scrolling and a worse UX.

Still, `minshell.html` was kept because it provides **maximum compatibility and easy testing**.
</details>

Q: **Why is the webversion generated in a subfolder of the docs directory?**
<details>
<summary>Answer</summary>

This is actually pretty simple. Our Doxygen docs are available at `immediate-mode-ui.github.io/Nuklear`, and it's very important to keep them up to date (i.e. keep the structure). The only practical option is to add the webversion as a **subfolder**, which is why it lives at `doc/html/webver`.
</details>

Q: **Why does** `docs.yml` **build the webversion on pull requests instead of only on push?**
<details>
<summary>Answer</summary>

In short, the main reason is that our CI/CD could use a big refactor. I mostly followed a **zero surprise policy**, so that after a custom PR is merged, the webversion and docs builds are guaranteed to succeed. Sure, one might argue that Emscripten builds should be tested in `ccpp.yml`, but that would either duplicate work or create issues with sequential workflow execution.
</details>

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -15,7 +15,7 @@ jobs:
         sudo apt-get install -y doxygen
     - name: build doc
       run: make docs
-      #Web-version
+      #Webversion
     - name: Install packages for webver
       run: |
         sudo apt-get update -qq
@@ -26,7 +26,7 @@ jobs:
       run: emcc -v
     - name: Build SDL3 port for Emscripten
       run: embuilder build sdl3
-    - name: Build web-version
+    - name: Build webversion
       run: make webver
       #Deploy to GitHub Pages
     - name: Deploy to GitHub Pages
--- a/webver/Readme.md
+++ b/webver/Readme.md
@@ -1,11 +1,22 @@
 Webversion notes
 ===
 
-1. **Folder Structure**
+## Terminology
 
-   We use the `site` subfolder to store all public webversion files. 
-   Current folder contains **internal** files for webversion support that do not need to be published.
+**Webversion** (or **webver** for short) is a demonstration of Nuklear being run in the browser via Emscripten.
+Currently it supports only the SDL3 backend, but long-term, additional backends may be added.
 
-2. **Update Policy**
+We use the consistent term **webversion** because it is already referenced across other files and may appear in user-created issues and pull requests.
 
-   The webversion is auto-updated through CI/CD workflow scripts. Any push triggers an automated build and deployment process, so no manual steps are required.
+## Folder Structure
+
+We use the `site` subfolder to store all public webversion files.
+Current folder contains **internal** files for webversion support that do not need to be published.
+
+## Update Policy
+
+The webversion is auto-updated through CI/CD workflow scripts. Any push triggers an automated build and deployment process, so no manual steps are required.
+
+## About Logo
+
+The Nuklear logo was suggested by Rafał Jopek and published in [issue #401](https://github.com/Immediate-Mode-UI/Nuklear/issues/401#issuecomment-2066737874)
--- a/webver/site/nuklear_logo_readme.txt
+++ b/webver/site/nuklear_logo_readme.txt
@@ -1,1 +1,0 @@
-Logo was suggested by Rafał Jopek and was publeshed here: https://github.com/Immediate-Mode-UI/Nuklear/issues/401#issuecomment-2066737874


.github/workflows/ccpp.yml
```diff
@@ -1,4 +1,4 @@
-name: C/C++ CI
+name: Checkers
```
RobLoach, line 1: Unrelated to this change?

webver/site/emscripten_logo.png
RobLoach, line 1: There isn't a real reason to provide the Emscripten logo. Have the minshell be the default one?

webver/site/nuklear_logo_readme.txt
```diff
@@ -0,0 +1 @@
+Logo was suggested by Rafał Jopek and was publeshed here: https://github.com/Immediate-Mode-UI/Nuklear/issues/401#issuecomment-2066737874
```
RobLoach: This documentation can live in `webver/Readme.md`

webver/Readme.md
```diff
@@ -0,0 +1,11 @@
+Webversion notes
+===
```
RobLoach, line 2:
``` suggestion
# Nuklear Web Demonstration
```

webver/Readme.md
```diff
@@ -0,0 +1,11 @@
...
+
+1. **Folder Structure**
+
+   We use the `site` subfolder to store all public webversion files. 
+   Current folder contains **internal** files for webversion support that do not need to be published.
+
+2. **Update Policy**
+
+   The webversion is auto-updated through CI/CD workflow scripts. Any push triggers an automated build and deployment process, so no manual steps are required.
```
RobLoach: This doesn't really need an ordered list. Best to just have straight text with headings.
``` suggestion
This is a web demonstration of Nuklear, built with Emscripten.

## Folder Structure

We use the `site` subfolder to store all public webversion files. Current folder contains **internal** files for webversion support that do not need to be published.

## Update Policy

The webversion is auto-updated through CI/CD workflow scripts. Any push triggers an automated build and deployment process, so no manual steps are required.
```

webver/site/index.html
```diff
@@ -0,0 +1,295 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+    <meta charset="UTF-8">
+    <meta name="viewport" content="width=device-width, initial-scale=1.0">
+    <title>Nuklear Emscripten Demo</title>
```
RobLoach, line 6: Since `webver` is essentially a demonstration built with Emscripten, we may be able to keep the root driectory clean by having the `webver` contents for this live in `demo/emscripten`.

Makefile
```diff
@@ -75,6 +83,16 @@ $(DOCS_PATH)/doxygen-awesome-css/doxygen-awesome.css:
 
 
 
+########################################################################################
+##   webver
```
RobLoach, line 87:
``` suggestion
##   Emscripten
```

Makefile
```diff
@@ -75,6 +83,16 @@ $(DOCS_PATH)/doxygen-awesome-css/doxygen-awesome.css:
...
+
+webver: 
+	mkdir -p $(WEBVER_OUT)
+	emmake make -C $(WEBVER_BACKEND) CFLAGS="$(WEBVER_CFLAGS)" BIN=$(abspath $(WEBVER_OUT)/demo_sdl3_renderer.js) TEMPDIR=$(abspath $(WEBVER_OUT))
+	cp -r $(WEBVER_SITE)/* $(WEBVER_OUT)/
+#[NOTE]We pass TEMPDIR directly for hide a potential issue
```
RobLoach, line 93: What potential issue are we hiding? It's probably best to not hide it. Building out to the temp directory instead of the destination directoy seems hacky.

Makefile
```diff
@@ -75,6 +83,16 @@ $(DOCS_PATH)/doxygen-awesome-css/doxygen-awesome.css:
 
 
 
+########################################################################################
+##   webver
+
+webver: 
+	mkdir -p $(WEBVER_OUT)
+	emmake make -C $(WEBVER_BACKEND) CFLAGS="$(WEBVER_CFLAGS)" BIN=$(abspath $(WEBVER_OUT)/demo_sdl3_renderer.js) TEMPDIR=$(abspath $(WEBVER_OUT))
```
RobLoach, line 91: Since this it building out assets, we could take advantage of Makefile's command caching...

``` suggestion
webver:
	mkdir -p $(WEBVER_OUT)

$(WEBVER_OUT)/demo_sdl3_renderer.js: webver
	emmake make -C $(WEBVER_BACKEND) CFLAGS="$(WEBVER_CFLAGS)" BIN=$(abspath $(WEBVER_OUT)/demo_sdl3_renderer.js) TEMPDIR=$(abspath $(WEBVER_OUT))
```

.github/workflows/docs.yml
```diff
@@ -1,22 +1,36 @@
-name: Documentation on github.io
+name: Docs, Webverion and Deploy
 
-on:
-  push:
-    branches: [ master ]
+on: [push, pull_request]
```
RobLoach, line 3: Docs should only be built on pushes to `master`. Plan on fixing this up once it's merged?

.github/workflows/docs.yml
```diff
@@ -1,22 +1,36 @@
...
+        sudo apt-get update -qq
+        sudo apt-get install -y make
+    - name: Setup Emscripten SDK
+      uses: mymindstorm/setup-emsdk@v14
+    - name: Verify Emscripten
+      run: emcc -v
+    - name: Build SDL3 port for Emscripten
+      run: embuilder build sdl3
+    - name: Build web-version
+      run: make webver
```
RobLoach: Since this is using Emscripten, we should likely just reference it as such. There could be non-emscripten builds that are web-enabled.

webver/site/nuklear_logo_readme.txt
PavelSharp: Thanks for the suggestion, fixed.

webver/Readme.md
PavelSharp: Agreed. Headings work better here.

.github/workflows/docs.yml
PavelSharp: I added some explanation in `webver/Readme.md` and replaced web-version with webversion. Webversion (or webver) means a specific term for the Nuklear+SDL3 build with Emscripten, which gets deployed to GitHub Pages.  It's something like an internal target for Nuklear support.
In my opinion, keeping one term across the project makes sense. But what do you think?

.github/workflows/docs.yml
PavelSharp: Just to clarify, this is already covered in the Mini FAQ. So my main question is **should we build the webversion on pull requests**?

Now, this PR prefers a "zero surprise" policy, which means we want to catch issues early before merging. This workflow also runs in parallel with `ccpp.yml`.
If that doesn't sound convincing, we could consider one of four options:
- On pull_request: only build webver
- On pull_request: only build docs [strange, but possible]
- On pull_request: build both webver + docs [current behavior]
- On pull_request: don't run docs.yml at all

Which would be better?

Makefile
PavelSharp: Would you like to read https://github.com/Immediate-Mode-UI/Nuklear/pull/902#discussion_r2838099837?
I believe consistency of terminology is important, but I'm open to your thoughts.

.github/workflows/docs.yml
sleeptightAnsiC: > Which would be better?

just like it was was before, only when pushing to master

Getting webver rebuild on each PR is costly and verbose. If it ever breaks, it will most likely be an issue with demo/sdl3_renderer (or with Emscripten, or with SDL itself) so full rebuild is unnecessary.

Instead, how about we use something similar to [.github/ci_compile_sources.sh](https://github.com/Immediate-Mode-UI/Nuklear/blob/master/.github/ci_compile_sources.sh) that will attempt to compile the demo with `emcc` without linking? (we will need SDL headers for this, so we have to deal with https://github.com/Immediate-Mode-UI/Nuklear/issues/897 first)

Makefile
sleeptightAnsiC: > we could take advantage of Makefile's command caching...

@RobLoach  This demo produces more files than just `$(WEBVER_OUT)/demo_sdl3_renderer.js`, so only catching this one here wouldn't be correct. Besides that, the `demo/sdl3_renderer/Makefile` is already caching under the hood (aka incremental build).

Also, the change that you suggested would not work, because this...

```mk
webver:
	mkdir -p $(WEBVER_OUT)
```
...would only end up calling `mkdir` and nothing else. What you probably had in mind was...

```mk
webver: $(WEBVER_OUT)/demo_sdl3_renderer.js

$(WEBVER_OUT)/demo_sdl3_renderer.js:
	mkdir -p $(WEBVER_OUT)
	emmake ...
```

...but as I said, this is unnecessary. This code is already okey as it is.

.github/workflows/docs.yml
sleeptightAnsiC: > > Since this is using Emscripten, we should likely just reference it as such.
>
> Webversion (or webver) means a specific term for the Nuklear+SDL3 build with Emscripten

Pavel is partially right here. Calling this thing "Emscripten" feels too ambiguous IMO. You can already build other demos with Emscripten, so calling it as such could be confusing.

However, Rob brings a good point, the name "_Webversion_" itself is quite confusing. Maybe we should consider a different name? like "_Webdemo_"? Still a bit ambiguous... Any better ideas?

> There could be non-emscripten builds that are web-enabled.

Could you clarify, please?

Makefile
sleeptightAnsiC: It needs tempdir because `demo/sdl3_renderer/Makefile` produces depfiles for incremental building, however, this whole `TEMPDIR=$(abspath $(WEBVER_OUT))` hack dumps all those depfiles into `WEBVER_OUT`, so this whole thing seems wrong.

@PavelSharp What you did with `BIN=$(abspath $(WEBVER_OUT)/demo_sdl3_renderer.js)` should be already enough, it should already work with default tempdir and produce binaries under WEBVER_OUT as expected. If you're getting any "_potential issue_", just tell us what it is exactly.

webver/site/emscripten_logo.png
sleeptightAnsiC: These logos are used as hyperlinks, so it would be nice to keep this functionality in one way or another. Also, the images just look nice overall, and make some sense in [showcase shell](https://pavelsharp.github.io/Nuklear/webver/).

The only problem I have here is .PNG format. Git doesn't play well with binary files, and small .PNG can't scale with higher resolution. Vendoring this inside of git repo feels wrong.

Can't we use .SVG format instead? Aren't web browsers supporting this by default?
We already have .SVG version of nuklear logo, here: https://github.com/Immediate-Mode-UI/Nuklear/files/15042034/nuklear.zip
For the Emscripten logo we can embed it by referencing this link: https://raw.githubusercontent.com/sthagen/emscripten-core-emscripten/refs/heads/main/media/powered_by_logo.svg

webver/site/index.html
sleeptightAnsiC: > we may be able to keep the root driectory clean by having the webver contents for this live in demo/emscripten.

Having this under `./demo/*` could be confusing, because this isn't really a demo, and because it already reuses most of the code from `./demo/sdl3_renderer`. Also, we may implement native/pure Emscripten backend in the future, so reserving this directory is problematic.

However, I do agree this thing living under root directory might be wrong. Not only this, it also uses two sub directories instead of one [screenshot below]. What is the purpose of `*/site` ?
image

@RobLoach How about `./example/webver` or `./doc/webver` ?

Makefile
```diff
@@ -5,9 +5,18 @@
...
+# Convenience shortcut to docs output path (as defined in Doxyfile)
+DOCS_OUT:=$(DOCS_PATH)/html
+
 DEMO_PATH=demo
 SRC_PATH=src
 
+WEBVER_BACKEND:=./demo/sdl3_renderer
+WEBVER_SITE:=./webver/site
+WEBVER_OUT:=$(DOCS_OUT)/webver
+WEBVER_CFLAGS:=-O2 -DINCLUDE_ALL -sEXPORTED_RUNTIME_METHODS=requestFullscreen
```
sleeptightAnsiC, line 17: `-O2` is fine, but you may consider `-Os`
People usually use `-0s` for release builds in Emscripten.
https://emscripten.org/docs/tools_reference/emcc.html#emcc-os

Makefile
```diff
@@ -5,9 +5,18 @@
...
+#Possible choice: ccall,cwrap,requestFullscreen,FS
+WEBVER_CFLAGS+=${CFLAGS}
```
sleeptightAnsiC, line 19: I feel like this section doesn't belong here.
Top level `./Makefile` is meant for launching other makefiles/programs.
Can't we have configuration in something like `vebver/Makefile` instead?

Makefile
```diff
@@ -75,6 +83,16 @@ $(DOCS_PATH)/doxygen-awesome-css/doxygen-awesome.css:
 
 
 
+########################################################################################
+##   webver
+
+webver: 
+	mkdir -p $(WEBVER_OUT)
```
sleeptightAnsiC, line 90: unless you have a very good reason to keep this line, you can probably remove it, and patch this instead
```diff
diff --git a/demo/sdl3_renderer/Makefile b/demo/sdl3_renderer/Makefile
index 48ee50e..22b83cd 100644
--- a/demo/sdl3_renderer/Makefile
+++ b/demo/sdl3_renderer/Makefile
@@ -47,7 +47,7 @@ DEP := ${TEMPDIR}/$(notdir ${BIN}).d
 SRC := main.c
 
 ${BIN}:
-    mkdir -p $(dir $@)
+    mkdir -p $(dir $@) ${TEMPDIR}
     ${CC} ${SRC} -o $@ -MD -MF ${DEP} ${cppflags} ${ldflags} ${ldlibs} ${cflags}
 
 ${BIN}: ${SRC}
```

webver/site/nuklear_logo.png
sleeptightAnsiC, line 1: https://github.com/Immediate-Mode-UI/Nuklear/pull/902#discussion_r2838744409

webver/site/index.html
RobLoach: `doc/webver` makes sense to me, since the goal is docs.

webver/Readme.md
```diff
@@ -0,0 +1,22 @@
...
+We use the `site` subfolder to store all public webversion files.
+Current folder contains **internal** files for webversion support that do not need to be published.
+
+## Update Policy
+
+The webversion is auto-updated through CI/CD workflow scripts. Any push triggers an automated build and deployment process, so no manual steps are required.
+
+## About Logo
+
+The Nuklear logo was suggested by Rafał Jopek and published in [issue #401](https://github.com/Immediate-Mode-UI/Nuklear/issues/401#issuecomment-2066737874)
```
RobLoach, line 22: The logo is likely more broadly scoped to the Docs. If we are uploading it, may be good to have it displayed in the main docs, and credited there.

Makefile
```diff
@@ -37,15 +46,16 @@ DEMO_LIST = $(shell find $(DEMO_PATH) -type f -name Makefile -printf "%h ")
...
 usage:
-	echo "make docs		to create documentation"
-	echo "make nuke		to rebuild the single header nuklear.h from source"
-	echo "make demos	to build all of the demos"
-	echo "make all 		to re-pack the header and create documentation"
+	@echo "make docs		to create documentation"
+	@echo "make webver		to build webversion on SDL3 + Emscripten (try to run 'embuilder build sdl3' if first time)"
+	@echo "make nuke		to rebuild the single header nuklear.h from source"
+	@echo "make demos		to build all of the demos"
+	@echo "make all 		to re-pack the header and create documentation"
```
RobLoach, line 56: `@echo` unrelated change.

Makefile
```diff
@@ -37,15 +46,16 @@ DEMO_LIST = $(shell find $(DEMO_PATH) -type f -name Makefile -printf "%h ")
 ######################################################################################
 
 
-.PHONY: usage all demos $(DEMO_LIST)
+.PHONY: usage all demos webver $(DEMO_LIST)
```
RobLoach, line 49: Don't see a need to build the `webver` on every `make` run.

webver/Readme.md
RobLoach, line 1: Rename to README.md

.github/workflows/docs.yml
```diff
@@ -1,22 +1,36 @@
...
       uses: actions/checkout@v4
-    - name: apt-update
-      run: sudo apt-get update -qq
-    - name: apt-get doxygen
-      run: sudo apt-get install -y doxygen
+      #Documentation
+    - name: Install packages for docs
+      run: |
+        sudo apt-get update -qq
+        sudo apt-get install -y doxygen
```
RobLoach, line 15: Unrelated change

.github/workflows/docs.yml
```diff
@@ -1,22 +1,36 @@
...
     - name: build doc
       run: make docs
-    - name: deploy
+      #Webversion
+    - name: Install packages for webver
+      run: |
+        sudo apt-get update -qq
```
RobLoach, line 21: We already do an update prior to this. Don't need to do it again.

.github/workflows/docs.yml
```diff
@@ -1,22 +1,36 @@
...
+        sudo apt-get install -y make
```
RobLoach, line 22: Could bundle the `make` install with doxygen above

.github/workflows/docs.yml
RobLoach:
``` suggestion
on:
  push:
    branches: [ master ]
```

Makefile
RobLoach:
```suggestion
.PHONY: usage all demos $(DEMO_LIST)
```

PR #903, opened 2026-02-22
docs: Move doc/Doxyfile to .Doxyfile

### LINKED ISSUE #901: [REFACTORING] .gitignore ignores `doc/Doxyfile` and needs cleanup
### Critical bug

The current [.gitignore](https://github.com/Immediate-Mode-UI/Nuklear/blob/master/.gitignore) rules cause `doc/Doxyfile` to be ignored and not tracked by Git. Of course, a quick fix  `!doc/Doxyfile` is available, but I would also like to talk about **refactoring**.

### Documentation layout problems

At the moment, the `docs/` directories are ignored, and the project does not actively use them. It would be better to choose a single name for the documentation directory (`doc/` vs `docs/`) and use it consistently. Since our Makefile has a [docs target](https://github.com/Immediate-Mode-UI/Nuklear/blob/3bcdec202920d5f3e8251c14af20a233a1590e2a/Makefile#L65), I would prefer this name for the output directory too.

Currently, all documentation is [generated into](https://github.com/Immediate-Mode-UI/Nuklear/blob/3bcdec202920d5f3e8251c14af20a233a1590e2a/doc/Doxyfile#L1274) `doc/html`, while files like `doxygen-awesome.css` are [downloaded](https://github.com/Immediate-Mode-UI/Nuklear/blob/3bcdec202920d5f3e8251c14af20a233a1590e2a/Makefile#L74) into the `doc/`. So we cannot ignore only `doc/html`.

### IDE-related ignores

The current `.gitignore` does not handle directories created by common IDEs, which makes development less convenient. For example, Visual Studio prefers a `.vs` directory, while Qt Creator prefers a `.qtcreator` directory. I'm sure these directories should be ignored by default. We can consider using well-known `.gitignore` templates from the [official GitHub repository](https://github.com/github/gitignore/tree/main).

RobLoach:
This moves the Doxyfile to a hidden file in the root. Allows a cleaner file structer, and avoids some of the problems brought up in #901.

Fixes #901.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -23,7 +23,7 @@ OUTPUT = nuklear.h
 
 PRIV1 = nuklear_internal.h nuklear_math.c nuklear_util.c nuklear_color.c nuklear_utf8.c nuklear_buffer.c nuklear_string.c nuklear_draw.c nuklear_vertex.c 
 
-EXTERN =  stb_rect_pack.h stb_truetype.h
+EXTERN =  stb_rect_pack.h stb_truetype.h 
 
 PRIV2 = nuklear_font.c nuklear_input.c nuklear_style.c nuklear_context.c nuklear_pool.c nuklear_page_element.c nuklear_table.c nuklear_panel.c nuklear_window.c nuklear_popup.c nuklear_contextual.c nuklear_menu.c nuklear_layout.c nuklear_tree.c nuklear_group.c nuklear_list_view.c nuklear_widget.c nuklear_text.c nuklear_image.c nuklear_9slice.c nuklear_button.c nuklear_toggle.c nuklear_selectable.c nuklear_slider.c nuklear_knob.c nuklear_progress.c nuklear_scrollbar.c nuklear_text_editor.c nuklear_edit.c nuklear_property.c nuklear_chart.c nuklear_color_picker.c nuklear_combo.c nuklear_tooltip.c
 


Makefile
```diff
@@ -62,7 +62,7 @@ nuke: $(addprefix $(SRC_PATH)/, $(SRC))
 ########################################################################################
 ##   Docs
 
-docs: $(DOCS_PATH)/html/index.html 
+docs: $(DOCS_PATH)/html/index.html
```
RobLoach:
``` suggestion
docs: $(DOCS_PATH)/html/index.html 
```

Makefile
```diff
@@ -45,7 +45,7 @@ usage:
 	echo "make demos	to build all of the demos"
 	echo "make all 		to re-pack the header and create documentation"
 
-all: docs nuke demos 
+all: docs nuke demos
```
RobLoach:
``` suggestion
all: docs nuke demos 
```

Makefile
```diff
@@ -21,9 +21,9 @@ INTRO =  HEADER.md
 PUB = nuklear.h
 OUTPUT = nuklear.h
 
-PRIV1 = nuklear_internal.h nuklear_math.c nuklear_util.c nuklear_color.c nuklear_utf8.c nuklear_buffer.c nuklear_string.c nuklear_draw.c nuklear_vertex.c 
+PRIV1 = nuklear_internal.h nuklear_math.c nuklear_util.c nuklear_color.c nuklear_utf8.c nuklear_buffer.c nuklear_string.c nuklear_draw.c nuklear_vertex.c
```
RobLoach:
``` suggestion
PRIV1 = nuklear_internal.h nuklear_math.c nuklear_util.c nuklear_color.c nuklear_utf8.c nuklear_buffer.c nuklear_string.c nuklear_draw.c nuklear_vertex.c 
```

Makefile
```diff
@@ -21,9 +21,9 @@ INTRO =  HEADER.md
...
 
-EXTERN =  stb_rect_pack.h stb_truetype.h 
+EXTERN =  stb_rect_pack.h stb_truetype.h
```
RobLoach:
``` suggestion
EXTERN =  stb_rect_pack.h stb_truetype.h 
```

PR #905, opened 2026-02-28
Add prefix to problematic stb_truetype equal() func

rswinkle:
I don't know how this function has lasted so long in such a widely used library but we are on the latest version of stb_truetype and it's [still there](https://github.com/nothings/stb/blob/master/stb_truetype.h#L4462).

It's such a small function and only used twice so I'm also fine with just eliminating it entirely but I figure just changing the name keeps it closer to upstream.

### DIFF
--- a/nuklear.h
+++ b/nuklear.h
@@ -16225,7 +16225,7 @@ static int stbtt__ray_intersect_bezier(float orig[2], float ray[2], float q0[2],
    }
 }
 
-static int equal(float *a, float *b)
+static int stbtt__equal(float *a, float *b)
 {
    return (a[0] == b[0] && a[1] == b[1]);
 }
@@ -16273,7 +16273,7 @@ static int stbtt__compute_crossings_x(float x, float y, int nverts, stbtt_vertex
             q1[1] = (float)y1;
             q2[0] = (float)x2;
             q2[1] = (float)y2;
-            if (equal(q0,q1) || equal(q1,q2)) {
+            if (stbtt__equal(q0,q1) || stbtt__equal(q1,q2)) {
                x0 = (int)verts[i-1].x;
                y0 = (int)verts[i-1].y;
                x1 = (int)verts[i  ].x;
--- a/src/stb_truetype.h
+++ b/src/stb_truetype.h
@@ -4459,7 +4459,7 @@ static int stbtt__ray_intersect_bezier(float orig[2], float ray[2], float q0[2],
    }
 }
 
-static int equal(float *a, float *b)
+static int stbtt__equal(float *a, float *b)
 {
    return (a[0] == b[0] && a[1] == b[1]);
 }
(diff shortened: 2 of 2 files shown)


sleeptightAnsiC:
Is the lack of prefix causing any symbol conflicts in your code? I didn't found anything in Nuklear where this would be the case.

Shouldn't this be reported to STB upstream first? I feel a bit uneasy that you're trying to patch it here. Someone may overwrite your changes when bumping this file.

sleeptightAnsiC:
If you take a look at git history of that file (`git log -p -- src/stb_truetype.h`) you'll notice this already has happened:
- f27ac6ef02c8aecd8a78a8af0620dad3d2ca3530 reintroduced all typos that 297e463d378427fbd989731ee30f1d4c06bffc01 fixed
- https://github.com/Immediate-Mode-UI/Nuklear/commit/cd7c7c555ce20be3fda3014eb9419a8a523c73fa reverted to `//` comments that 1f48b5c860c92a199c3375cfee4b2d920034aab2 changed
- probably more I didn't notice...

I'm afraid your patch will meet the same fate ;/

rswinkle:
> Is the lack of prefix causing any symbol conflicts in your code? I didn't found anything in Nuklear where this would be the case.

It's conflicting with my code not Nuklear or it would have been discovered much immediately. However, whether as stb_truetype alone, or as part of Nuklear, having a function named "equal" in a C library, especially a single header library, is problematic. Sure it's declared static and I could put it in a separate TU, but I tend to prefer unity builds, (or at least unity-ish builds), so that doesn't help. And honestly, I'm surprised  someone even made a function so small that's used literally twice.

> Shouldn't this be reported to STB upstream first? I feel a bit uneasy that you're trying to patch it here. Someone may overwrite your changes when bumping this file.
>I'm afraid your patch will meet the same fate ;/

Theoretically, but stb_truetype hasn't been updated in almost 5 years so I wouldn't hold my breath for a new release.

Even if there were an update eventually, unless it's a significant fix/improvement there's no real need to update our local copy, and if it is worth it to update, and the people involved in updating forgot to keep the change, no big deal.
It can always be re-added if someone wants it, and having it in for all the time between now and then is still more than worth  it. 

Remember, no one forces developers to update Nuklear (and most developers using it seriously make local changes to libraries like Nuklear anyway), meaning if/when stb_truetype get's updated and this eventually possibly gets overwritten, they won't even notice even if it would affect them because they're not grabbing the latest version immediately just because it exists because updating means they have to re-apply all their changes again and handle any conflicts.

Anyway, long story short, I think it's worth changing, whether prefixing or eliminating it entirely.

sleeptightAnsiC (approved):
I just entered an upstream issue for this: https://github.com/nothings/stb/issues/1898

I personally don't mind this being patched here. It's not a big deal.

nothings:
Yes, this is absolutely a bug that we will fix upstream as well.

RobLoach (approved):
```c
equals(AWESOME);
// => true
```

Issue #848, opened 2025-11-02
Add example usage of NK_BUTTON_DOUBLE


sleeptightAnsiC:
Although every demo implements some kind of `NK_BUTTON_DOUBLE` support, none of the examples shows it's usage, neither does the library use it anywhere internally. [Searching for it literally shows nothing.](https://github.com/search?q=repo%3AImmediate-Mode-UI%2FNuklear+NK_BUTTON_DOUBLE&type=code&p=1) There should be at least one small example, so we can properly test the double-click behavior.

(found this when working on https://github.com/Immediate-Mode-UI/Nuklear/pull/779#discussion_r2322188493 )

sleeptightAnsiC:
similar problem with: `NK_BUTTON_MIDDLE`

RobLoach:
Not really an example, but a good way to test it perhaps?
<img width="432" height="504" alt="Image" src="https://github.com/user-attachments/assets/0a399fa2-e609-4776-afc4-be952f78b76e" />

sleeptightAnsiC:
Yea, that is fine.

Literally anything, just to get it tested. No worries that it looks raw.

sleeptightAnsiC:
Fixed as part of https://github.com/Immediate-Mode-UI/Nuklear/pull/906

PR #906, opened 2026-03-06
mouse: Add the Back/Forward Mouse Buttons

RobLoach:
Found I wanted to allow use of the back/forward 4/5 butons on the mouse.

### DIFF
--- a/src/nuklear.h
+++ b/src/nuklear.h
@@ -599,9 +599,9 @@ enum nk_buttons {
     NK_BUTTON_LEFT,
     NK_BUTTON_MIDDLE,
     NK_BUTTON_RIGHT,
-    NK_BUTTON_DOUBLE,
-    NK_BUTTON_X1,
-    NK_BUTTON_X2,
+    NK_BUTTON_DOUBLE, /* Double click of the Left mouse button. */
+    NK_BUTTON_X1, /* Commonly used for "Back" in UI navigation. Mouse Button 4. */
+    NK_BUTTON_X2, /* Commonly used for "Forward" in UI navigation. Mouse Button 5. */
     NK_BUTTON_MAX
 };
 
--- a/nuklear.h
+++ b/nuklear.h
@@ -822,9 +822,9 @@ enum nk_buttons {
     NK_BUTTON_LEFT,
     NK_BUTTON_MIDDLE,
     NK_BUTTON_RIGHT,
-    NK_BUTTON_DOUBLE,
-    NK_BUTTON_X1,
-    NK_BUTTON_X2,
+    NK_BUTTON_DOUBLE, /* Double click of the Left mouse button. */
+    NK_BUTTON_X1, /* Commonly used for "Back" in UI navigation. Mouse Button 4. */
+    NK_BUTTON_X2, /* Commonly used for "Forward" in UI navigation. Mouse Button 5. */
     NK_BUTTON_MAX
 };
 


sleeptightAnsiC:
What these will be used for? Asking out of curiosity.

You may want to add an example usage somewhere to `demo/common/overview.c`, just to have some place to test these. You could also add one for `NK_BUTTON_MIDDLE` and `NK_BUTTON_DOUBLE` in order to address https://github.com/Immediate-Mode-UI/Nuklear/issues/848 at the same time. I could quickly write this and send you a diff, if you're interested.

EDIT: Also, how about other backends? Pretty sure most of them support mouse 4/5

RobLoach:
Mostly for nuklear_console to allow going back in the menu. Do have to figure out a history in it though. https://github.com/RobLoach/nuklear_console/

demo/common/overview.c
```diff
@@ -637,6 +637,8 @@ overview(struct nk_context *ctx)
                 nk_layout_row_static(ctx, 180, 278, 1);
                 nk_edit_string(ctx, NK_EDIT_BOX, box_buffer, &box_len, 512, nk_filter_default);
 
+
+
```
RobLoach:
``` suggestion
```

demo/d3d11/nuklear_d3d11.h
```diff
@@ -404,6 +404,30 @@ nk_d3d11_handle_event(HWND wnd, UINT msg, WPARAM wparam, LPARAM lparam)
...
+            nk_input_button(&d3d9.ctx, NK_BUTTON_X2, (short)LOWORD(lparam), (short)HIWORD(lparam), 1);
+            break;
+        }
+        SetCapture(wnd);
+        return 1;
+
+    case WM_XBUTTONUP:
+        switch (GET_XBUTTON_WPARAM(wparam)) {
+        case XBUTTON1:
+            nk_input_button(&d3d9.ctx, NK_BUTTON_X1, (short)LOWORD(lparam), (short)HIWORD(lparam), 0);
```
RobLoach: I have no way to test D3D11, but I've done my best to get it to what it should be, according to the docs I'm reading.

demo/x11/nuklear_xlib.h
rswinkle, line 1: Did X11 just never add macros for these buttons? That seems like a glaring lack since they've been around forever at this point. Then again it looks like Wayland is using magic numbers for everything... shrug

demo/common/overview.c
```diff
@@ -1354,6 +1354,56 @@ overview(struct nk_context *ctx)
...
+            /* Mouse Wheel */
+            nk_layout_row_dynamic(ctx, 30, 1);
+            nk_label(ctx, "Mouse Wheel", NK_TEXT_LEFT);
+            nk_layout_row_dynamic(ctx, 20, 2);
+            nk_labelf(ctx, NK_TEXT_LEFT, "X: %.2f", in->mouse.scroll_delta.x);
+            nk_labelf(ctx, NK_TEXT_LEFT, "Y: %.2f", in->mouse.scroll_delta.y);
+            nk_tree_pop(ctx);
+        }
+
```
rswinkle, line 1391: This is nice, and long overdue imo. Did we want a way to show a double click in there? I think I remember that being mentioned. It can always be added later.

rswinkle:
This looks good. I just tested a couple backends but it seems fine. Interesting X1 is the back button and X2 is the forward button (both in physical positioning and what they trigger). I wonder how they decided that. Maybe just that back is the more common action, and on mine a least it's the larger button.

demo/x11/nuklear_xlib.h
RobLoach: Yeah I could not find a define anywhere for these. I also found those values not from their official docs, but some random forum post online 🤣

They seem to work though, which is good. The only ones I didn't test were the D3D ones 🤷

src/nuklear.h
```diff
@@ -600,6 +600,8 @@ enum nk_buttons {
     NK_BUTTON_MIDDLE,
     NK_BUTTON_RIGHT,
     NK_BUTTON_DOUBLE,
+    NK_BUTTON_X1,
+    NK_BUTTON_X2,
```
sleeptightAnsiC: @RobLoach maybe this should be reverted back to:
```
NK_BUTTON_BACK
NK_BUTTON_FORWARD
```
I feel like X1/X2 is a bit too ambiguous for a name.
Most people understand what Left/Right/Middle/Double click means and does,
but mouse4/5/X1/X2 doesn't really suggest how this could be used by the library.

demo/common/overview.c
RobLoach: Double click is in there. Right above the mouse wheel one 👍

src/nuklear.h
RobLoach: I switched to it X1 and X2 because back/forewards that may not be their intended function. None of the renderers also refer to them as Back/Forwards. Majority call it X1/X2, which is likely safer to use.

demo/x11/nuklear_xlib.h
sleeptightAnsiC: > The only ones I didn't test were the D3D ones

Booting up my VM :)

demo/common/overview.c
RobLoach: Cleaned up the code to include double click within the loop.

demo/d3d11/nuklear_d3d11.h
```diff
@@ -404,6 +404,30 @@ nk_d3d11_handle_event(HWND wnd, UINT msg, WPARAM wparam, LPARAM lparam)
...
+            break;
+        case XBUTTON2:
+            nk_input_button(&d3d9.ctx, NK_BUTTON_X2, (short)LOWORD(lparam), (short)HIWORD(lparam), 0);
+            break;
+        }
```
sleeptightAnsiC: This doesn't compile. I had to change all `d3d9` to `d3d11` like so:

```suggestion
            nk_input_button(&d3d11.ctx, NK_BUTTON_X1, (short)LOWORD(lparam), (short)HIWORD(lparam), 1);
            break;
        case XBUTTON2:
            nk_input_button(&d3d11.ctx, NK_BUTTON_X2, (short)LOWORD(lparam), (short)HIWORD(lparam), 1);
            break;
        }
        SetCapture(wnd);
        return 1;

    case WM_XBUTTONUP:
        switch (GET_XBUTTON_WPARAM(wparam)) {
        case XBUTTON1:
            nk_input_button(&d3d11.ctx, NK_BUTTON_X1, (short)LOWORD(lparam), (short)HIWORD(lparam), 0);
            break;
        case XBUTTON2:
            nk_input_button(&d3d11.ctx, NK_BUTTON_X2, (short)LOWORD(lparam), (short)HIWORD(lparam), 0);
            break;
        }
```

demo/d3d12/nuklear_d3d12.h
```diff
@@ -528,6 +528,30 @@ nk_d3d12_handle_event(HWND wnd, UINT msg, WPARAM wparam, LPARAM lparam)
...
+        SetCapture(wnd);
+        return 1;
+
+    case WM_XBUTTONUP:
+        switch (GET_XBUTTON_WPARAM(wparam)) {
+        case XBUTTON1:
+            nk_input_button(&d3d9.ctx, NK_BUTTON_X1, (short)LOWORD(lparam), (short)HIWORD(lparam), 0);
+            break;
+        case XBUTTON2:
+            nk_input_button(&d3d9.ctx, NK_BUTTON_X2, (short)LOWORD(lparam), (short)HIWORD(lparam), 0);
```
sleeptightAnsiC: Same problem here. Should be `d3d12` instead of `d3d9`

```suggestion
            nk_input_button(&d3d12.ctx, NK_BUTTON_X1, (short)LOWORD(lparam), (short)HIWORD(lparam), 1);
            break;
        case XBUTTON2:
            nk_input_button(&d3d12.ctx, NK_BUTTON_X2, (short)LOWORD(lparam), (short)HIWORD(lparam), 1);
            break;
        }
        SetCapture(wnd);
        return 1;

    case WM_XBUTTONUP:
        switch (GET_XBUTTON_WPARAM(wparam)) {
        case XBUTTON1:
            nk_input_button(&d3d12.ctx, NK_BUTTON_X1, (short)LOWORD(lparam), (short)HIWORD(lparam), 0);
            break;
        case XBUTTON2:
            nk_input_button(&d3d12.ctx, NK_BUTTON_X2, (short)LOWORD(lparam), (short)HIWORD(lparam), 0);
```

src/nuklear.h
RobLoach: In review, only raylib refers to them as Back/Forwards

demo/d3d11/nuklear_d3d11.h
sleeptightAnsiC: Just retested, and besides of https://github.com/Immediate-Mode-UI/Nuklear/pull/906#discussion_r2902616356 and https://github.com/Immediate-Mode-UI/Nuklear/pull/906#discussion_r2902626128 , it does work.

image

demo/x11/nuklear_xlib.h
RobLoach: Oh, and GDI, because who uses GDI these days 💀

src/nuklear.h
sleeptightAnsiC: > because back/forewards that may not be their intended function

What was your use case in nuklear_console?

It feels odd that core library would ever use these for something else than navigation. Many desktop applications reuse those buttons for going back and forward (aka reverse-back) to some spots in the UI. In web browser it causes the tab to go back/forward in the page history.

demo/x11/nuklear_xlib.h
sleeptightAnsiC: > Oh, and GDI, because who uses GDI these days 💀

@RobLoach  Anything else? I probably have VM for everything (haha)

src/nuklear.h
RobLoach: In games, I use it for the Melee button, and Specials... In nuklear_console, I'll be using it for...

1. Navigation within the UI
2. Allowing the user to use it as an input button:

Screenshot from 2026-03-08 20-21-16

demo/common/overview.c
rswinkle: I am soo blind *facepalm*. I think I was also unconsciously thinking/remembering that DOUBLE_CLICK is not inherently tied to the left button in Nuklear and was thinking it was a separate state for all buttons rather than one of Nuklear's pseudo buttons.  But we tie it to the left button in the backends as the sane default.

Anyway, it makes sense to put it in the loop as it is a "button" of sorts.

demo/x11/nuklear_xlib.h
RobLoach: I think that's all. I tested allegro, glfw, rawfbs, sdl, sdl3, sfml, x11, xcb :heavy_check_mark:

src/nuklear.h
sleeptightAnsiC: > I'll be using it for... 1. Navigation within the UI

Not sure if I understand right. This will be a different kind of navigation to what most apps do? genuinely asking what's the plan here

I found those keys quite strange. It's hard to mentally imagine which one exactly is it based on raw 4/5/X1/X2. The buttons appear on actual physical mice as _closer to you_ (Back/X2) and _further from you_ (Forward/X1) so I thought the previous naming makes some sense.

I'm just afraid we won't understand what's going on when seeing something like `mouse.buttons[NK_BUTTON_X2].down` in the code. The backend library (like SDL/GLFW) doesn't care about actual name, it only serves the input and expects the app/user to treat it in any possible way, but Nuklear associates _"common sense behavior"_ with each key/button

@rswinkle Any thoughts? You mentioned in https://github.com/Immediate-Mode-UI/Nuklear/pull/906#issuecomment-4020168807 that X1/X2 are "reversed" to what you initially expected.  [EDIT: or at least this was my impression when I read your comment.] Would BUTTON_BACK/FORWARD (or any other naming really) make a difference?

I don't want to pressure on changing anything. Just want to ensure the decision about those names is fully aware.

demo/x11/nuklear_xlib.h
sleeptightAnsiC: > GDI

Doesn't seem to compile, gimmy sec

sleeptightAnsiC: > > GDI
> 
> Doesn't seem to compile, gimmy sec

Ah, nvm. You already fixed it in https://github.com/Immediate-Mode-UI/Nuklear/pull/906/changes/95619a71bd262b525061e42e848b1329d3b9762b. I just had to pull.

src/nuklear.h
RobLoach: I'm using SDL3 as a modal for all of them here...
- SDL_BUTTON_X1: NK_BUTTON_X1
- SDL_BUTTON_X2: NK_BUTTON_X2

The button that hits the same button as SDL3's X1 is hitting the same NK_BUTTON_X1 across all other renderers I've tested.

demo/common/overview.c
```diff
@@ -1354,6 +1354,41 @@ overview(struct nk_context *ctx)
...
+            for (i = 0; i < NK_BUTTON_MAX; i++) {
+                nk_label(ctx, button_names[i], NK_TEXT_LEFT);
+                if (nk_input_is_mouse_pressed(in, i))
+                    nk_label(ctx, "Pressed", NK_TEXT_LEFT);
+                else if (nk_input_is_mouse_down(in, i))
+                    nk_label(ctx, "Down", NK_TEXT_LEFT);
+                else if (nk_input_is_mouse_released(in, i))
+                    nk_label(ctx, "Released", NK_TEXT_LEFT);
+                else
+                    nk_label(ctx, "Up", NK_TEXT_LEFT);
```
sleeptightAnsiC, line 1380: Because of how Nuklear serves those events, it's almost impossible to see Pressed/Released appearing. I even had hard time to capture recording of this (because my display is 75Hz) and I haven't seen it ever appearing in some demos (not sure if that was a bug in specific demo or not)

Any ideas on improving this? My first one was to split Down/Up and Pressed/Released events into two columns, but I don't think this would help much anyway.

This can be fixed as follow up PR but maybe you see any quick solution now?

src/nuklear.h
sleeptightAnsiC: Alright. We've talked about it, and figured out this will be resolved with doc comment: https://github.com/Immediate-Mode-UI/Nuklear/pull/906/changes/929db27caa5311105eba60892c876ce42ff4a860 / https://discord.com/channels/922234432071548968/934963188011773952/1480391263613223043


@RobLoach Feel free to resolve this thread, unless you wish to wait for rswinkle's opinion

demo/common/overview.c
RobLoach: That's true. Pressed is only displayed for 1 frame. Same with Released. Unsure of a good way to ensure that's clear without some timers and gross code. Maybe a follow up.

demo/x11/nuklear_xlib.h
sleeptightAnsiC: > Did X11 just never add macros for these buttons?

Tried everything, and couldn't find any,
not even for XCB (which is way newer than Xlib)
I guess this thing is really 40 years old, huh...

@RobLoach this thread should be fine to resolve

demo/common/overview.c
sleeptightAnsiC: Probably just split it and then make the text slowly fade away. This file already uses delta time in few places, but yea, it might become ugly.

I'll enter a task for it, right after this PR gets merged.

sleeptightAnsiC (approved):
@RobLoach 
Alright, everything that was possible to test, was tested.
I would suggest to wait a day or two, and then merge.
A lot of changes happened today, maybe we missed something in a rush.

PR #907, opened 2026-03-07
tests: Verify `nuklear.h` is Rebuilt

### LINKED ISSUE #890: add CI for detecting a mismatch between ./src/* and ./nuklear.h

RobLoach:
Fixes #890


.github/workflows/ccpp.yml
```diff
@@ -13,6 +13,14 @@ jobs:
       run: sudo apt-get update -qq
     - name: apt get demo-libs
       run: sudo apt-get install -y --no-install-recommends glslc liballegro5-dev liballegro-image5-dev liballegro-ttf5-dev libcairo2-dev libglfw3 libglfw3-dev libglew-dev libsdl2-dev libvulkan-dev libwayland-dev libx11-dev libxcb1-dev libxcb-util0-dev libxcb-keysyms1-dev libxft-dev libxkbcommon-x11-dev wayland-protocols
+
+    # Verify Nuklear.h is clean
+    - name: Rebuild Nuklear.h
+      run: make nuke
+    - name: Verify `build.py` was run
+      run: git diff --exit-code nuklear.h
```
sleeptightAnsiC: This approach "_works_" but I was thinking about something more "_user friendly_", so people won't need our involvement in order to understand what's going on. Currently, as I see it, there are bunch of downsides to this:
- It only prints a diff and exists (it would be nice to have some message)
- You can't run it without CI (I wanted to have a rule for this in ./Makefile)
- It ~~deletes~~ **overwrites** the amalgamation, so it could delete someone's progress (not a problem here, but will be in case we want to reuse this outside of CI, when testing before committing)
- ~~Does it actually work when someone only applies changes to amalgamation??~~

sleeptightAnsiC (changes requested):
I would suggest to just give me some time. I already had clear vision on how to implement this under https://github.com/Immediate-Mode-UI/Nuklear/issues/890 and haven't really expected anyone to overtake it.

The trick with `make nuke && git diff ...` may work, and we can probably accept it now and only improve it later, but if people are already confused with the workflow, I'm not sure they will understand this check.

Makes sense?

.github/workflows/ccpp.yml
sleeptightAnsiC: > Does it actually work when someone only applies changes to amalgamation??

NVM... [it does](https://github.com/Immediate-Mode-UI/Nuklear/actions/runs/22803824601/job/66149606878)

sleeptightAnsiC: > It only prints a diff and exists (it would be nice to have some message)

How about this ?

```suggestion
    # Verify ./nuklear.h mirrors ./src/*
    - name: Verify ./nuklear.h
      run: make nuke && \
           if ! git diff --exit-code nuklear.h ; then \
               echo "Did you forgot to rebuild ./nuklear.h ?" >/dev/stderr ; \
               echo "Make sure to apply your changes to ./src/* first, and only then rebuild the amalgamation with ./src/paq* !"  >/dev/stderr ; \
               exit 1 ; \
           fi ;
```

^ I just wrote this by hand. Not sure if Github's .yml even supports multiline expressions. I just wanted to show what I meant.

RobLoach: Certainly more complicated, but sure 😜

RobLoach:
@sleeptightAnsiC line 1: syntax error near unexpected token `then'

Looks pretty gross lol. Likely a cleaner way to do that.

sleeptightAnsiC:
gimme sec... It probably tries to collapse it into one line and fails miserably, haha

.github/workflows/ccpp.yml
```diff
@@ -13,6 +13,17 @@ jobs:
...
       run: sudo apt-get install -y --no-install-recommends glslc liballegro5-dev liballegro-image5-dev liballegro-ttf5-dev libcairo2-dev libglfw3 libglfw3-dev libglew-dev libsdl2-dev libvulkan-dev libwayland-dev libx11-dev libxcb1-dev libxcb-util0-dev libxcb-keysyms1-dev libxft-dev libxkbcommon-x11-dev wayland-protocols
+
+    # Verify ./nuklear.h mirrors ./src/*
+    - name: Verify ./nuklear.h
+      run: make nuke && \
+           if ! git diff --exit-code nuklear.h ; then \
+               echo "Did you forgot to rebuild ./nuklear.h ?" >/dev/stderr ; \
+               echo "Make sure to apply your changes to ./src/* first, and only then rebuild the amalgamation with ./src/paq* !"  >/dev/stderr ; \
+               exit 1 ; \
+           fi ;
```
sleeptightAnsiC: Sorry, I have no idea how to reliably test this on my local machine, and just doing it blindly. According to [this SO thread](https://stackoverflow.com/a/70319796/19562526), the `|` should enable multilines, so it should be fine now (I hope!)

```suggestion
      run: |
        make nuke
        if ! git diff --exit-code nuklear.h ; then
          echo "Did you forgot to rebuild ./nuklear.h ?" >/dev/stderr
          echo "Make sure to apply your changes to ./src/* first, and only then rebuild the amalgamation with ./src/paq* !"  >/dev/stderr
          exit 1
        fi
```

sleeptightAnsiC: > I have no idea how to reliably test this on my local machine

Found this one: https://github.com/nektos/act
I'm installing it right now.

RobLoach: Moved it to a Makefile command, and avoided the nesting.

sleeptightAnsiC: It does work!
image

sleeptightAnsiC: @RobLoach ah sh1t, I was 5sec too late. Mind reverting your last commit (https://github.com/Immediate-Mode-UI/Nuklear/pull/907/changes/c2a909bdd53d3bbd2735c9533386b1a70c9a8724) and accepting this one (https://github.com/Immediate-Mode-UI/Nuklear/pull/907#discussion_r2900175440) instead?

No need to have a partial step in the Makefile. It wouldn't really work without CI and could confuse us in the future. Better keep everything in the workflow file alone.

.github/workflows/ccpp.yml
```diff
@@ -9,10 +9,18 @@ jobs:
...
+      run: make nuke
+    - name: Verify ./nuklear.h
+      run: make nuke-test
+
     - name: apt-update
       run: sudo apt-get update -qq
     - name: apt get demo-libs
       run: sudo apt-get install -y --no-install-recommends glslc liballegro5-dev liballegro-image5-dev liballegro-ttf5-dev libcairo2-dev libglfw3 libglfw3-dev libglew-dev libsdl2-dev libvulkan-dev libwayland-dev libx11-dev libxcb1-dev libxcb-util0-dev libxcb-keysyms1-dev libxft-dev libxkbcommon-x11-dev wayland-protocols
+
+    # Build the Demos
```
RobLoach:
``` suggestion
```

nuklear.h
```diff
@@ -17792,11 +17792,12 @@ nk_font_atlas_add(struct nk_font_atlas *atlas, const struct nk_font_config *conf
     NK_ASSERT(config->ttf_size);
     NK_ASSERT(config->size > 0.0f);
 
+    // Adding stuff to Amagamation.
     if (!atlas || !config || !config->ttf_blob || !config->ttf_size || config->size <= 0.0f||
         !atlas->permanent.alloc || !atlas->permanent.free ||
         !atlas->temporary.alloc || !atlas->temporary.free)
         return 0;
-
+d
```
``` suggestion
```

.github/workflows/ccpp.yml
sleeptightAnsiC: this can be resolved

sleeptightAnsiC:
Just for the record, I had a very different idea on how to implement this. Wanted something that will be possible to run outside of CI. Got some prototype at: https://gist.github.com/sleeptightAnsiC/900062bb7bc0c47b5395e496b452500d

Rob decided this would be too complex. In case only CI is involved, I tend to agree.

nuklear.h
```diff
@@ -17792,6 +17792,7 @@ nk_font_atlas_add(struct nk_font_atlas *atlas, const struct nk_font_config *conf
     NK_ASSERT(config->ttf_size);
     NK_ASSERT(config->size > 0.0f);
 
+    // Adding stuff to Amagamation.
```
``` suggestion
```

PR #912, opened 2026-03-08
Return if value changed for property_*() and combobox*() functions

### LINKED ISSUE #910: Change nk_property_[type]()  and similar functions like nk_combobox() to return true if the value changed
I am constantly having to do something like this:

```
	tmp = nk_propertyi(ctx, "Degrees:", -180, img->rotdegs, 180, 1, 0.5);
	if (tmp != img->rotdegs) {
		img->edited = TO_ROTATE;
		img->rotdegs = tmp;
	}
```
or similar with the nk_property_int style just setting tmp beforehand.

Also this:
```
	int tmp = ts->filter;
	ts->filter = nk_combo(ctx, tex_filter_strs, NK_LEN(tex_filter_strs), ts->filter, g->font_size+20, nk_vec2(bounds.w, 400));
	if (tmp != ts->filter) {
		// TODO
		//glTextureParameteri(texi, GL_TEXTURE_MIN_FILTER, tex_filters[ch0_filter]);
		glTextureParameteri(texi, GL_TEXTURE_MAG_FILTER, tex_filters[ts->filter]);
	}
```
I can't be the only person who needs to know *if* something changed because I need to take an action either immediately or at a specific later time. I suppose in some cases you can just do unnecessary work every frame just in case but often that isn't practical because the work is expensive or there are threading issues involved or similar.

We have other functions in nuklear that return whether something changed, so why are we not using the return values of nk_property_[type]() and nk_combobox()?

I am definitely going to make these changes locally but I'd be happy to make pull request today if we can agree. And let me know if there are other similar functions wasting their return value like this.

rswinkle:
This fixes #910 exactly as discussed.

Just changes the return types of nk_property_int() and nk_property_float() as well as the 4 nk_combobox*() functions to nk_bool to indicate whether the value changed.

@RobLoach @sleeptightAnsiC

### DIFF
--- a/nuklear.h
+++ b/nuklear.h
@@ -3580,7 +3580,7 @@ NK_API nk_bool nk_color_pick(struct nk_context*, struct nk_colorf*, enum nk_colo
  *     a `#` at the beginning. It will not be shown but guarantees correct behavior.
  *
  * ```c
- * void nk_property_int(struct nk_context *ctx, const char *name, int min, int *val, int max, int step, float inc_per_pixel);
+ * nk_bool nk_property_int(struct nk_context *ctx, const char *name, int min, int *val, int max, int step, float inc_per_pixel);
  * ```
  *
  * Parameter           | Description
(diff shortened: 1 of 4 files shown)


RobLoach (approved):
Yup

sleeptightAnsiC (approved):
Yea, this is fine.

Unless we don't want to discuss https://github.com/Immediate-Mode-UI/Nuklear/issues/910#issuecomment-4019236341, this should be good as it is.

I've tested it locally under `demo/sdl3_renderer` with `-DINCLUDE_ALL` and didn't find any issues. Also, looked through other `return void` functions and didn't find anything else worth changing. Seems like you found everything already.

PR #913, opened 2026-03-08
Fix `NK_MEMCPY()` size types to `nk_size`

RobLoach:
There were a few calls to `NK_MEMCPY` that were passing the incorrect data type, resulting in warnings. This fixes it.

These warnings can pop up with...
```
-Wall -Wextra -Wconversion -Wsign-conversion
```


RobLoach (commented):
This fixes some warnings when `nk_size` isn't a standard type. I'd like to get this in. Thanks!

rswinkle:
I don't see any issues with this but could you tell me what situation caused the warning exactly? I'm probably being stupid but I just spent half an hour trying to get a warning to trigger with this snippet of code

```
	char buf1[20] = { 1, 2, 3 };
	char buf2[20] = { 10, 9, 8 };
	nk_size tmp = 10;
	unsigned int tmp2 = 5;
	int tmp3 = 15;
	NK_MEMCPY(buf1, buf2, tmp);
	NK_MEMCPY(buf1, buf2, tmp2);
	NK_MEMCPY(buf1, buf2, tmp3);
```

with gcc/g++, clang/clang++ and different flags and compiler options. Clearly I"m missing something.

RobLoach:
Did another compilation from the latest with raylib-nuklear with the warnings enabled. I guess it's when what's passed in doesn't match what `nk_size` is....


```
nuklear.h:29176:85: warning: conversion to ‘nk_size’ {aka ‘long unsigned int’} from ‘int’ may change the sign of the result [-Wsign-conversion]
29176 |             NK_MEMCPY(win->property.prev_buffer, win->property.buffer, win->property.length);
      |                                                                        ~~~~~~~~~~~~~^~~~~~~
nuklear.h: In function ‘nk_tooltip_begin_offset’:
nuklear.h:314:21: warning: conversion from ‘float’ to ‘int’ may change value [-Wfloat-conversion]
  314 | #define NK_MAX(a,b) ((a) < (b) ? (b) : (a))
      |                     ^
nuklear.h:30732:9: note: in expansion of macro ‘NK_MAX’
30732 |     h = NK_MAX(win->layout->row.min_height, ctx->style.font->height+2*ctx->style.window.padding.y);
      |         ^~~~~~
nuklear.h:30735:68: warning: conversion from ‘int’ to ‘float’ may change value [-Wconversion]
30735 |     x = nk_ifloorf(in->mouse.pos.x + 1) - (int)win->layout->clip.x + offset.x;
      |                                                                    ^
nuklear.h:30735:9: warning: conversion from ‘float’ to ‘int’ may change value [-Wfloat-conversion]
30735 |     x = nk_ifloorf(in->mouse.pos.x + 1) - (int)win->layout->clip.x + offset.x;
      |         ^~~~~~~~~~
/home/rob/Documents/raylib-nuklear/include/nuklear.h:30736:68: warning: conversion from ‘int’ to ‘float’ may change value [-Wconversion]
30736 |     y = nk_ifloorf(in->mouse.pos.y + 1) - (int)win->layout->clip.y + offset.y;
      |                                                                    ^
nuklear.h:30736:9: warning: conversion from ‘float’ to ‘int’ may change value [-Wfloat-conversion]
30736 |     y = nk_ifloorf(in->mouse.pos.y + 1) - (int)win->layout->clip.y + offset.y;
      |         ^~~~~~~~~~
```

rswinkle:
That's really weird... I should have gotten that warning with my snippet with -Wall -Wextra but I never saw it. I'll try again later, but like I said, there's nothing wrong with this commit and I'm about to submit a couple tiny PR's myself.

sleeptightAnsiC:
> > I just spent half an hour trying to get a warning to trigger with this snippet of code [...] Clearly I"m missing something.
>
> I should have gotten that warning with my snippet with -Wall -Wextra but I never saw it.

Your snippet does trigger it, just not with the default warnings. See godbolt link below. It comes from `-Wconversion` enabled. Maybe it worked differently in older GCC/Clang, I'm not entirely sure.

I was able to repro the results from https://github.com/Immediate-Mode-UI/Nuklear/pull/913#issuecomment-4107373670 with https://godbolt.org/z/5KeaPsEer , but I never got the warning in `nk__lit` (although it makes sense it could appear there)

src/nuklear_font.c
```diff
@@ -735,7 +735,7 @@ nk__lit(unsigned char *data, unsigned int length)
     NK_ASSERT (nk__dout + length <= nk__barrier);
     if (nk__dout + length > nk__barrier) { nk__dout += length; return; }
     if (data < nk__barrier2) { nk__dout = nk__barrier+1; return; }
-    NK_MEMCPY(nk__dout, data, length);
+    NK_MEMCPY(nk__dout, data, (nk_size)length);
```
sleeptightAnsiC, line 738: If you take a look at what `nk_decompress_token` does, you'll see `unsigned int` casts everywhere, and it's the only place that calls `nk__lit`/`nk__match`, so the typing in all those functions doesn't really make much sense.

I don't mind suppressing a warning here, but beware this probably not very correct.

RobLoach: That makes sense. These warnings pop up with `-Wall -Wextra -Wconversion -Wsign-conversion`, and `NK_MEMCPY` can be [`memcpy`](https://en.cppreference.com/w/cpp/string/byte/memcpy.html), which could take a `size_t`, which can be larger than an `unsigned int`.

The casting shouldn't break anything, given your NK_SIZE is the same size or greater than an unsigned int.

sleeptightAnsiC: Yea, it won't break anything.

Just saying this place is a mess overall, no matter if it emits a warning or not.

Feel free to resolve this thread.

PR #916, opened 2026-03-10
docs(layouting): Fix code blocks formatting

### LINKED ISSUE #915: Code blocks broken in Layouting docs
While browsing the documentation I noticed that the code blocks are currently broken on the [Layouting page](https://immediate-mode-ui.github.io/Nuklear/Layouting.html):

Image

TomJGooding:
While browsing the documentation I noticed that the code blocks are currently broken on the Layouting page.

I confess I'm not familiar with Doxygen, but simply removing a space from the indentation seems to fix the issue!

Fixes https://github.com/Immediate-Mode-UI/Nuklear/issues/915

### DIFF
--- a/nuklear.h
+++ b/nuklear.h
@@ -2202,8 +2202,7 @@ NK_API void nk_rule_horizontal(struct nk_context *ctx, struct nk_color color, nk
  * To actually define a layout you just call the appropriate layouting function
  * and each subsequent widget call will place the widget as specified. Important
  * here is that if you define more widgets then columns defined inside the layout
- * functions it will allocate the next row without you having to make another layouting <br /><br />
- * call.
+ * functions it will allocate the next row without you having to make another layouting call. <br /><br />
  *
  * Biggest limitation with using all these APIs outside the `nk_layout_space_xxx` API
  * is that you have to define the row height for each. However the row height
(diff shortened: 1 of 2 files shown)


TomJGooding:
Screenshot for comparison:

image

sleeptightAnsiC (approved):
Thanks! Tested locally and it does work.

I've taken a quick look into all code blocks and found one that lacks proper syntax highlight:

https://github.com/Immediate-Mode-UI/Nuklear/blob/987c8b89eb016c3611c014790735bdc3430adcc2/src/nuklear.h#L5623-L5637

^ no need to do anything, but if you're already interested in contributing to docs, you can probably submit another PR for this one too (or even fix it here, it wouldn't be a big deal).

EDIT: tried to fix it myself, but coloring does not seem to change no matter what I do. I'm not super familiar with doxygen neither.

TomJGooding:
Thanks for the review. I think the problem in the Stack docs should really be a separate PR - this fixes a borked code block rather than just missing syntax highlighting.

RobLoach (approved):
Thanks. Would like to touch up all the doxygen docs for sure. Make them consistent, and maybe even tweak the config itself.

Issue #896, opened 2026-02-12
Hidden/closed window is to be ignored due to nk_clear


QuanMz:
There's a issue when you set a window hidden (either manually set its `NK_WINDOW_HIDDEN` flag or do `nk_window_show()` ), or you close the window (window with `NK_WINDOW_CLOSED` flag), they will only take effect for a period, only then to be reinitialized, which defeats the purpose of hiding,closing and them being flags in the first place.

So this line of code in `nk_clear` function (defined in nuklear_context.c, is called everytime you render a frame)
https://github.com/Immediate-Mode-UI/Nuklear/blob/a329721db93eb336f8eaae20f9c822730dd2ed6c/src/nuklear_context.c#L137-L147
(do cleanups for 'active' window that isn't really active)
specifically the line `ctx->begin = 0;` (reset the start pointer of a linked list of windows)
the problem here is it assumes that the hidden window isn't at the 'ctx->begin' (where the 'win->prev' doesnt exist), which isn't always the case, but it set to that anyway without checking

then it choose the window before the hidden one, which may not exist, so it trigger the code and make Nuklear completely forget the window list, 
and what worse is that the forgotten window (and the ones after it) won't be freed from memory by the later section of the code since its `win->seq` (I guess this is like its heartbeat) is valid with the `ctx`'s
https://github.com/Immediate-Mode-UI/Nuklear/blob/a329721db93eb336f8eaae20f9c822730dd2ed6c/src/nuklear_context.c#L165-L171
since the reference is gone, the windows will live in the memory with the program, causing memory leak (thanks you @sleeptightAnsiC for letting me know)

and in the next frame, `nk_begin` (how you start to draw things) can't find the existing window you're talkiing about (since the list reference is gone) so it will create new one with your spec, that means your hidden window is now reseted and live somewhere else in the memory, and when it's hidden and its 'win->prev' gets invalid (due to some internal reorder for Z indexing or something), that cleanup code (for invalid active window) will be triggered, now you're stuck in a loop where you hide window just to see it popup again.
<sub>this issue doesn't happen if window is both HIDDEN and CLOSED</sub>

<details>
<summary> About its history/changelog </summary>
(hello)

---
As @sleeptightAnsiC said, it really is regression!
for example, the [web demo](https://dexp.github.io/nuklear-webdemo) of Nuklear made by DeXP ~9 years ago, the closing window feature works perfectly fine
but today it breaks?
so I went to digging, bc im very bored

[...]

PR #919, opened 2026-03-14
Check y for NULL too, remove duplicate zero'ing

rswinkle:
Ran into this bug today. I don't have a simple standalone program to reproduce it, but my current project was triggering it consistently.

What happens is that the x and y offsets are split across two different tables and then the table with the y offset gets removed in nk_clear() because table->seq != ctx->seq at some point. Then the next frame, x is found so it assumes y exists and viola, you have a segfault.

Clearly the underlying logic of how it uses seq to determine what is safe to free is flawed but I don't know what the best way to fix it.

I think this is a very rare case that is happening because I have a large number of groups (a single table holds 69 items or 34.5 x/y pairs) in a "tabbed" window so what happens as I click through the tabs more groups get added to the table and eventually we get the x/y table split. Then when you switch to another tab you "touch" at least some of the items in the first table thus updating its seq member to the current ctx->seq, but not the newest table with the y member of the no-longer visible group, so that table gets tossed. Finally, switching back to the that tab it still finds the x offset in the first table and assumes the y must also exist even though it doesn't.

One thing I didn't look deeply into but I should is if we can avoid this code entirely when NK_WINDOW_NO_SCROLLBAR is set. It's just a waste of time/space and most groups are used for simple layouting purposes and so will not need a scrollbar. Complex arrangements require lots of groups, thus making running into this issue more likely. Just a thought.

In the process I noticed a redundant nk_zero() call, so I removed it.

@RobLoach @sleeptightAnsiC

### DIFF
--- a/src/nuklear_table.c
+++ b/src/nuklear_table.c
@@ -12,7 +12,6 @@ nk_create_table(struct nk_context *ctx)
     struct nk_page_element *elem;
     elem = nk_create_page_element(ctx);
     if (!elem) return 0;
-    nk_zero_struct(*elem);
     return &elem->data.tbl;
 }
 NK_LIB void
--- a/nuklear.h
+++ b/nuklear.h
@@ -23340,7 +23340,13 @@ nk_group_begin_titled(struct nk_context *ctx, const char *id,
         NK_ASSERT(y_offset);
         if (!x_offset || !y_offset) return 0;
         *x_offset = *y_offset = 0;
-    } else y_offset = nk_find_value(win, id_hash+1);
+    } else if (!(y_offset = nk_find_value(win, id_hash+1))) {
+        y_offset = nk_add_value(ctx, win, id_hash+1, 0);
+        NK_ASSERT(y_offset);
+        if (!y_offset) return 0;
+        *x_offset = *y_offset = 0; /* I think this covers the degenerate case */
+    }
+
     return nk_group_scrolled_offset_begin(ctx, x_offset, y_offset, title, flags);
 }
 NK_API nk_bool
@@ -23382,7 +23388,12 @@ nk_group_get_scroll(struct nk_context *ctx, const char *id, nk_uint *x_offset, n
         NK_ASSERT(y_offset_ptr);
         if (!x_offset_ptr || !y_offset_ptr) return;
         *x_offset_ptr = *y_offset_ptr = 0;
-    } else y_offset_ptr = nk_find_value(win, id_hash+1);
+    } else if (!(y_offset_ptr = nk_find_value(win, id_hash+1))) {
+        y_offset_ptr = nk_add_value(ctx, win, id_hash+1, 0);
+        NK_ASSERT(y_offset_ptr);
+        if (!y_offset_ptr) return;
+        *x_offset_ptr = *y_offset_ptr = 0;
+    }
     if (x_offset)
       *x_offset = *x_offset_ptr;
     if (y_offset)
@@ -23417,7 +23428,11 @@ nk_group_set_scroll(struct nk_context *ctx, const char *id, nk_uint x_offset, nk
         NK_ASSERT(y_offset_ptr);
         if (!x_offset_ptr || !y_offset_ptr) return;
         *x_offset_ptr = *y_offset_ptr = 0;
-    } else y_offset_ptr = nk_find_value(win, id_hash+1);
+    } else if (!(y_offset_ptr = nk_find_value(win, id_hash+1))) {
+        NK_ASSERT(y_offset_ptr);
+        if (!y_offset_ptr) return;
+        *x_offset_ptr = *y_offset_ptr = 0;
+    }
     *x_offset_ptr = x_offset;
     *y_offset_ptr = y_offset;
 }
@@ -23467,7 +23482,11 @@ nk_list_view_begin(struct nk_context *ctx, struct nk_list_view *view,
         NK_ASSERT(y_offset);
         if (!x_offset || !y_offset) return 0;
         *x_offset = *y_offset = 0;
-    } else y_offset = nk_find_value(win, title_hash+1);
+    } else if (!(y_offset = nk_find_value(win, title_hash+1))) {
+        NK_ASSERT(y_offset);
+        if (!y_offset) return 0;
+        *x_offset = *y_offset = 0;
+    }
     view->scroll_value = *y_offset;
     view->scroll_pointer = y_offset;
 
--- a/nuklear.h
+++ b/nuklear.h
@@ -19896,7 +19896,6 @@ nk_create_table(struct nk_context *ctx)
     struct nk_page_element *elem;
     elem = nk_create_page_element(ctx);
     if (!elem) return 0;
-    nk_zero_struct(*elem);
     return &elem->data.tbl;
 }
 NK_LIB void
@@ -19986,7 +19985,6 @@ nk_create_panel(struct nk_context *ctx)
     struct nk_page_element *elem;
     elem = nk_create_page_element(ctx);
     if (!elem) return 0;
-    nk_zero_struct(*elem);
     return &elem->data.pan;
 }
 NK_LIB void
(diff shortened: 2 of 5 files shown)


src/nuklear_table.c
```diff
@@ -12,7 +12,6 @@ nk_create_table(struct nk_context *ctx)
     struct nk_page_element *elem;
     elem = nk_create_page_element(ctx);
     if (!elem) return 0;
-    nk_zero_struct(*elem);
```
sleeptightAnsiC, line 15: (haven't read nor tested anything else yet)

> I noticed a redundant nk_zero() call, so I removed it.

There is one other function that does the very same thing - [nk_create_panel](https://github.com/Immediate-Mode-UI/Nuklear/blob/987c8b89eb016c3611c014790735bdc3430adcc2/src/nuklear_panel.c#L15) - both of these don't do anything meaningful other than zero'ing memory, so it's a bit suspicious to remove the only meaningful line from it, however, you're right [nk_create_page_element](https://github.com/Immediate-Mode-UI/Nuklear/blob/987c8b89eb016c3611c014790735bdc3430adcc2/src/nuklear_page_element.c#L30-L32) already zero'es the values.

I would suggest to remove the same line from [nk_create_panel](https://github.com/Immediate-Mode-UI/Nuklear/blob/987c8b89eb016c3611c014790735bdc3430adcc2/src/nuklear_panel.c#L15) in order to keep it consistent, but honesty I feel uneasy any of this is being done here, since it isn't really affecting your fix in any way.

rswinkle: > (haven't read nor tested anything else yet)
> 
> > I noticed a redundant nk_zero() call, so I removed it.
> 
> There is one other function that does the very same thing - [nk_create_panel](https://github.com/Immediate-Mode-UI/Nuklear/blob/987c8b89eb016c3611c014790735bdc3430adcc2/src/nuklear_panel.c#L15) - both of these don't do anything meaningful other than zero'ing memory, so it's a bit suspicious to remove the only meaningful line from it, however, you're right [nk_create_page_element](https://github.com/Immediate-Mode-UI/Nuklear/blob/987c8b89eb016c3611c014790735bdc3430adcc2/src/nuklear_page_element.c#L30-L32) already zero'es the values.
> 
> I would suggest to remove the same line from [nk_create_panel](https://github.com/Immediate-Mode-UI/Nuklear/blob/987c8b89eb016c3611c014790735bdc3430adcc2/src/nuklear_panel.c#L15) in order to keep it consistent, but honesty I feel uneasy any of this is being done here, since it isn't really affecting your fix in any way.

Both nk_create_table() and nk_craete_panel() do something besides the nk_zero() it's just very minimal. They return the pointer of the correct type of the union so the caller doesn't have to.

But I agree, I should remove it from both, and yeah it should probably be its own commit, it just bothers me. There's so many little things (even typos) that I want to just slip in sometimes because it doesn't seem worth it to create a whole other branch + pull request to wait on.

RobLoach (approved):
Seems fine. I agree that other suggestion is a good one, but may live in a different PR better.

rswinkle:
> Seems fine. I agree that other suggestion is a good one, but may live in a different PR better.

K, I'll look into avoiding unnecessary table use when I get the chance, and create another PR for removing the redundant nk_zero calls (maybe I will throw some other similar minor fixes into that one though).

src/nuklear_table.c
```diff
@@ -12,6 +12,7 @@ nk_create_table(struct nk_context *ctx)
     struct nk_page_element *elem;
     elem = nk_create_page_element(ctx);
     if (!elem) return 0;
+    nk_zero_struct(*elem);
```
RobLoach: Oh, sorry. I meant the other panel stuff sleep suggested 🙃 
``` suggestion
```

sleeptightAnsiC:
Splitting to another commit should be fine. Just make sure to rework commit history during rebase, so there will be only two when this gets merged (take a look at [kimgr/git-rewrite-guide](https://github.com/kimgr/git-rewrite-guide) in case of troubles)

With the changes done to both functions, this should be okey. I don't mind it being in the same PR.

rswinkle:
> Splitting to another commit should be fine. Just make sure to rework commit history during rebase, so there will be only two when this gets merged (take a look at [kimgr/git-rewrite-guide](https://github.com/kimgr/git-rewrite-guide) in case of troubles)
> 
> With the changes done to both functions, this should be okey. I don't mind it being in the same PR.

Are you talking to me or Rob here? I can try rewriting the history into 2 commits and doing a force push...

RobLoach:
Can Squash when bringing in, but there's only two commits. While resolving git conflicts is annoying, this only has ~50 line changes so it should be easy if there are big problems.

rswinkle:
> Can Squash when bringing in, but there's only two commits. While resolving git conflicts is annoying, this only has ~50 line changes so it should be easy if there are big problems.

I think I might try fixing it myself anyway. I don't think I've ever done an interactive rebase before and something with 2 commits and such simple changes seems like a good time to practice.  Just having to pull out one line change from the first commit and add it and another to the second.... what could go wrong? ha

sleeptightAnsiC:
> Are you talking to me or Rob here?

To you :)

Seems like Rob didn't read our conversation fully. I was also confused about his request.

> I don't think I've ever done an interactive rebase before and something with 2 commits and such simple changes seems like a good time to practice. [...] what could go wrong? ha

Don't worry. Even if you force-push something by mistake, Github tracks all those changes. Nothing ever gets lost in PR branches.

rswinkle:
Well I rebased and logically split the commits. As far as I know there is no way to avoid a force push when you do a rebase since the tags change even if nothing else.

EDIT: English, do I speak it??

sleeptightAnsiC (approved):
Thanks.

I've looked into the diffs, and this feels more like a workaround rather than a fix, since y_offset disappearing should (probably) never happen in the first place. Did you notice the window to rapidly "_flicker_" when any of those new branches are being hit? I imagine this would be a symptom.

Anyway, that code was already doing many questionable things, and I don't have any "_real_" proof against this patch, so I'm going to trust you and Rob's judgement.

rswinkle:
> Thanks.
> 
> I've looked into the diffs, and this feels more like a workaround rather than a fix, since y_offset disappearing should (probably) never happen in the first place. Did you notice the window to rapidly "_flicker_" when any of those new branches are being hit? I imagine this would be a symptom.

No flicker. Like I said there is a flaw in the way the "garbage collection" determines and I don't think there is a fix without larger more serious changes which probably aren't worth it, at least not right this second as they would require quite a bit of thought (and likely debate). For instance, the fact that iirc, both groups and windows are treated the same for the purposes of seq and subsequent garbage collection when they shouldn't be, and how even aside from that, a single integer seq is probably insufficient information to determine if a table can be freed.

We could just remove garbage collection for tables.

Or we could have a "touched" flag for each table that is reset each frame which makes determining whether it's garbage straightforward, but I would need to look into whether table.seq is used for anything else and can be removed and whether there are any other places where it is misused, then if those misuses require different fixes or not, because if there's flawed logic here, it's entirely likely there is similar elsewhere.

> Anyway, that code was already doing many questionable things, and I don't have any "_real_" proof against this patch, so I'm going to trust you and Rob's judgement.

Up to you guys if you want me to try for a deeper fix now or just get this bug fix in and look into that later. I would probably try go for option B above.

sleeptightAnsiC:
Let's merge it and hope for the best, at least for now.

Even if it's wrong, I would only imagine this thing resetting the scrollbar/content position in some very rare cases. Probably way better than getting a random crash. The garbage collection logic is already giving me nightmares (https://github.com/Immediate-Mode-UI/Nuklear/issues/896#issuecomment-4055910256) and I would sleep more peacefully knowing no one is trying to actively change it.

@RobLoach remember to NOT squash it :)

rswinkle:
> Even if it's wrong, I would only imagine this thing resetting the scrollbar/content position in some very rare cases. 

Oh I have no worries that it's wrong. Logically it can't be. The new code only applies in the edge case that causes a crash when they both would have been set to 0 on creation anyway if they were in the same garbage collected table.

My only concern was as you said it is a fix of an underlying problem and adding 1 additional branch over resolving the real issue.

PR #922, opened 2026-03-19
Created demo for sdl3 and vulkan

Lemento:
It's just a slightly modified version of… the base sdl_vulkan demo

### DIFF
--- a/demo/sdl3_vulkan/Makefile
+++ b/demo/sdl3_vulkan/Makefile
@@ -1,0 +1,30 @@
+# Install
+BIN = demo
+
+# Flags
+CFLAGS += -std=c89 -Wall -Wextra -pedantic -fsanitize=address -O2
+CFLAGS += -DSDL_DISABLE_IMMINTRIN_H
+
+SRC = main.c
+OBJ = $(SRC:.c=.o)
+
+ifeq ($(OS),Windows_NT)
+BIN := $(BIN).exe
+LIBS = -lsdl2 -lvulkan -lm
+else
+	UNAME_S := $(shell uname -s)
+	SDL2 := $(shell pkg-config --libs sdl2)
+	LIBS = $(SDL2) -lvulkan -lm
+endif
+
+
+$(BIN): shaders/demo.vert.spv shaders/demo.frag.spv
+	@mkdir -p bin
+	rm -f bin/$(BIN) $(OBJS)
+	$(CC) $(SRC) $(CFLAGS) -o bin/$(BIN) $(LIBS)
+
+shaders/demo.vert.spv: shaders/demo.vert
+	glslc --target-env=vulkan shaders/demo.vert -o shaders/demo.vert.spv
+
+shaders/demo.frag.spv: shaders/demo.frag
+	glslc --target-env=vulkan shaders/demo.frag -o shaders/demo.frag.spv
--- a/demo/sdl3_vulkan/main.c
+++ b/demo/sdl3_vulkan/main.c
@@ -1,0 +1,2295 @@
+/* nuklear - 1.32.0 - public domain */
+#include <limits.h>
+#include <stdbool.h>
+#include <stddef.h>
+#include <stdint.h>
+#include <stdio.h>
+#include <time.h>
+
+#include <SDL3/SDL.h>
+#include <SDL3/SDL_vulkan.h>
+#include <vulkan/vulkan.h>
+
+#define NK_INCLUDE_FIXED_TYPES
+#define NK_INCLUDE_STANDARD_IO
+#define NK_INCLUDE_STANDARD_VARARGS
+#define NK_INCLUDE_DEFAULT_ALLOCATOR
+#define NK_INCLUDE_VERTEX_BUFFER_OUTPUT
+#define NK_INCLUDE_FONT_BAKING
+#define NK_INCLUDE_DEFAULT_FONT
+#define NK_IMPLEMENTATION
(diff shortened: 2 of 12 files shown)


Lemento:
Found that reading from binary shader ended early because read mode wasn't set to "rb" binary.
Still doesn't fix visual bugs with the node_editor demo, everything else appears to be fine.

demo/sdl3_vulkan/main.c
```diff
@@ -0,0 +1,2294 @@
...
+    }
+
+    printf("available instance extensions:\n");
+    for (i = 0; i < available_instance_extension_count; i++) {
+        printf("  %s\n", available_instance_extensions[i].extensionName);
+    }
+
+    const char*const* found_extensions =
+        SDL_Vulkan_GetInstanceExtensions(&sdl_extension_count);
+    if (found_extensions == NULL)
```
sleeptightAnsiC, line 285: syntax error here ?

```suggestion
    if (found_extensions == NULL) {
```

demo/sdl3_vulkan/Makefile
```diff
@@ -0,0 +1,30 @@
...
+OBJ = $(SRC:.c=.o)
+
+ifeq ($(OS),Windows_NT)
+BIN := $(BIN).exe
+LIBS = -lsdl2 -lvulkan -lm
+else
+	UNAME_S := $(shell uname -s)
+	SDL2 := $(shell pkg-config --libs sdl2)
+	LIBS = $(SDL2) -lvulkan -lm
+endif
```
sleeptightAnsiC, line 18: this is still trying to link with old sdl2

demo/sdl3_vulkan/main.c
```diff
@@ -0,0 +1,2294 @@
...
+    uint32_t image_index;
+    VkResult result;
+    VkSemaphore nk_semaphore;
+
+    #ifdef INCLUDE_CONFIGURATOR
+    static struct nk_color color_table[NK_COLOR_COUNT];
+    memcpy(color_table, nk_default_color_style, sizeof(color_table));
+    #endif
+
+    if (SDL_Init(SDL_INIT_VIDEO) < 0) {
```
sleeptightAnsiC, line 2127: sdl3 doesn't use `< 0` idiom any more. This would work back in sdl2, but misbehaves under sdl3

sleeptightAnsiC (changes requested):
I have no idea how you were able to test it. Not only this doesn't compile at all on my machine, there are clearly leftovers from sdl2 that would make it impossible to work on sdl3.

demo/sdl3_vulkan/src/nuklear_sdl_vulkan.in.h
```diff
@@ -0,0 +1,1426 @@
...
+unsigned char nuklearshaders_nuklear_vert_spv[] = {};
+unsigned int nuklearshaders_nuklear_vert_spv_len = 0;
+unsigned char nuklearshaders_nuklear_frag_spv[] = {};
+unsigned int nuklearshaders_nuklear_frag_spv_len = 0;
+// NUKLEAR_SHADERS_END
+
+#include <assert.h>
+#include <stddef.h>
+#include <string.h>
+#include <SDL2/SDL.h>
```
RobLoach, line 30: Here too.

demo/sdl3_vulkan/nuklear_sdl_vulkan.h
RobLoach, line 1: Should probably name the file nuklear_sdl3_vulkan.h

demo/sdl3_vulkan/nuklear_sdl_vulkan.h
```diff
@@ -0,0 +1,1668 @@
...
+ *
+ *                              API
+ *
+ * ===============================================================
+ */
+#ifndef NK_SDL_VULKAN_H_
+#define NK_SDL_VULKAN_H_
+
+unsigned char nuklearshaders_nuklear_vert_spv[] = {
+  0x03, 0x02, 0x23, 0x07, 0x00, 0x00, 0x01, 0x00, 0x0b, 0x00, 0x0d, 0x00,
```
RobLoach, line 17: To do things like this in the past, I've used CMake embed-binaries:
https://github.com/andoalon/embed-binaries

sleeptightAnsiC:
@Lemento Did you submit this PR by mistake or something?

The old code from `demo/sdl_vulkan` would never work with SDL3 out of the box. It would require way more than being "_slightly modified_", to even get it compiling and booting in the first place.

Lemento:
> @Lemento Did you submit this PR by mistake or something?
> 
> 
> 
> The old code from `demo/sdl_vulkan` would never work with SDL3 out of the box. It would require way more than being "_slightly modified_", to even get it compiling and booting in the first place.

No I just have a bad habit of linking everything myself and not using the build system.
I've built this lib tons of times before so I forgot to even look at the Makefile, Ill fix this.

sleeptightAnsiC:
@Lemento

You may have better luck taking SDL3 code from [demo/sdl3_renderer](https://github.com/Immediate-Mode-UI/Nuklear/tree/master/demo/sdl3_renderer) and applying Vulkan rendering to it.

If you aren't comfortable with how demo/sdl_vulkan is being build, I would suggest to just abandon the idea of porting it. This thing is huge and overcomplicated, and getting it wrong will cost us a lot of time.

RobLoach (changes requested):
Definitely building off of the other SDL3 demo would be best. Good luck!

PR #924, opened 2026-03-23
Increase property button size so you can actually click them

rswinkle:
I've had this change locally in [sdl_img](https://github.com/rswinkle/sdl_img) for years.

Before:
Screenshot from 2026-03-22 20-38-47

After:
Screenshot from 2026-03-22 20-39-01

### DIFF
--- a/nuklear.h
+++ b/nuklear.h
@@ -28919,7 +28919,7 @@ nk_do_property(nk_flags *ws,
     struct nk_rect empty;
 
     /* left decrement button */
-    left.h = font->height/2;
+    left.h = font->height;
     left.w = left.h;
     left.x = property.x + style->border + style->padding.x;
     left.y = property.y + style->border + property.h/2.0f - left.h/2;
--- a/src/nuklear_property.c
+++ b/src/nuklear_property.c
@@ -162,7 +162,7 @@ nk_do_property(nk_flags *ws,
     struct nk_rect empty;
 
     /* left decrement button */
-    left.h = font->height/2;
+    left.h = font->height;
     left.w = left.h;
     left.x = property.x + style->border + style->padding.x;
     left.y = property.y + style->border + property.h/2.0f - left.h/2;


RobLoach:
Big left and righties. I like it.

RobLoach (approved):
I'd be good with this. Makes it easier to see & use.

sleeptightAnsiC (changes requested):
This was already possible without patching the library, just by changing some padding/border style values to the negative ones, example:

<details><summary>screenshot</summary>
<p>

image

</p>
</details>

I have code that depends on this, here:
https://github.com/Immediate-Mode-UI/Nuklear/blob/9c24f9c2b9434df3073d7bf66ccdb8043526adb3/demo/sdl3_renderer/main.c#L251-L252
and I'm pretty sure someone did a similar thing elsewhere. So the change, as it is right now, could mess with custom styling.

I agree for making property buttons bigger by default, just not in the way you did it here. Could you simply change default style values, so everyone's custom styling would remain the same?

rswinkle:
> This was already possible without patching the library, just by changing some padding/border style values to the negative ones, example:

Many things are possible without patching the library but we patch the library anyway to make it more convenient and better by default.

> I have code that depends on this, here:
> 
> https://github.com/Immediate-Mode-UI/Nuklear/blob/9c24f9c2b9434df3073d7bf66ccdb8043526adb3/demo/sdl3_renderer/main.c#L251-L252
> 
> 
> and I'm pretty sure someone did a similar thing elsewhere. So the change, as it is right now, could mess with custom styling.
> I agree for making property buttons bigger by default, just not in the way you did it here. Could you simply change default style values, so everyone's custom styling would remain the same?

You realize you're being inconsistent and hypocritical here right? Not only is having a *default* padding value be negative ridiculous imo (without checking, I'm sure we don't have that anywhere else), but by changing the default we are *also* *"forcing"* other people to change their code to keep the same behavior. The fact that in this case you wouldn't have to because you already made the same styling change doesn't change that it is, in fact, a "breaking" change, even though according to the definition you used in the other PR and elsewhere it isn't breaking at all.

It's clear that your position is always "don't make me do anything to my own code if I update the library" regardless of if it's compile time breaking (aka changing prototypes other than adding a return type, changing existing enum or struct member names, etc.), or appearance/behavior "breaking" in this case. Like I said, it seems almost pathological.

Between our two positions, I would say it makes much more sense for you to fork Nuklear and stop holding back the progress for everyone else rather than have someone like me fork Nuklear and try to pull a neovim (which lets face it, forced Vim to change drastically for the first time and quickly add the major new features neovim had in years to avoid becoming irrelevant). It makes far more sense for the fork that wants to progress to have the larger community and contributors to share the load and come up with ideas. On the other hand, a fork focused on never changing will obviously require less work to maintain and manage and as a sole maintainer/authority, you have final authority to reject any PR's you do get (though most would naturally go to the more dynamic project over time).

For reference, here's a bit of the history:

Neovim was forked from Vim in February 2014 by Brazilian developer Thiago de Arruda (tarruda).  The project was officially announced on the vim_use mailing list with the primary goal of aggressively refactoring the codebase to improve extensibility, usability, and modern features like asynchronous job control, which were not being accepted into mainline Vim at the time

Vim versions:
2001–2013: Major functional expansions occurred with Version 6.0 (folding, plugins, 2001), Version 7.0 (spell checking, tabs, 2006), and Version 7.4 (faster regex engine, 2013). 
2016–2022: The editor evolved with Version 8.0 (asynchronous I/O, 2016), Version 8.2 (popup windows, 2019), and Version 9.0 (Vim9 script, 2022). 

If Bram had accepted Thiago's changes to begin with neovim probably would never have been started. Too be fair it probably was a good thing in the long run, as we got a more active/vibrant/advancing project with more contributors and long term goals, compared to Vim, especially after Bram's death:

Vim Current Status: While the development continues with active contributions, Brabandt has described the project as being in "maintenance mode" for the foreseeable future, noting that the team lacks the resources for full-time work or major feature overhauls like those seen in Neovim.

EDIT: clarification on compile time breaking.

RobLoach:
No need to get personal. While you're right that padding -2 would be an anti-pattern, it does demonstrate that it's possible to change styling without patching.

I think having sane defaults should be a standard here. Having the property arrows be the same height as the rest of the text seems like a sane default. Thanks!

rswinkle:
> No need to get personal. While you're right that padding -2 would be an anti-pattern, it does demonstrate that it's possible to change styling without patching.

I might have gone a bit far with the "pathological" comments and could maybe have found a better way to express the same idea, but I definitely wanted to show the inconsistency.

Too your second point, again, anything is possible but we shouldn't force users to jump through hoops to accomplish common use cases/desires. A good example of this that I want to get in sometime is what he have [here](https://github.com/Immediate-Mode-UI/Nuklear/blob/master/demo/common/overview.c#L1092) in overview where we use pushes/pops and other manual styling to get buttons to behave like selectables to get a tab system that looks nicer than pure selectables. We should definitely be able to get that behavior without all that work.

> I think having sane defaults should be a standard here. Having the property arrows be the same height as the rest of the text seems like a sane default. Thanks!

Agreed. I'm not saying there are no other ways to do it but this seems the simplest change to make.

sleeptightAnsiC:
> https://github.com/Immediate-Mode-UI/Nuklear/pull/924#issuecomment-4129138343

Jesus Christ... What is this? Why would someone bring Vim/Neovim history into this thread? Are you copy pasting the text from LLM or something? Calling me inconsistent/hypocritical/pathological for using negative values that style_configurator itself allows me to do? What...?

You are repeatedly bringing low effort changes that nobody else asked for, and each time I ask you to just keep some old stuff working, you burst into tears and post a wall of text about it.

Alright, I've seen enough. I won't be replying to you any more. Do whatever.

rswinkle:
> > [#924 (comment)](https://github.com/Immediate-Mode-UI/Nuklear/pull/924#issuecomment-4129138343)
> 
> Jesus Christ... What is this? Why would someone bring Vim/Neovim history into this thread? Are you copy pasting the text from LLM or something?

I brought them up to compare the two philosophies and how things turned out. Competition willing to make breaking changes forced Bram to man up and start breaking things in his big "catch up" from 7.4 to 8.0, and to a lesser extent years later with 8.0 to 9.0 for his faster vim9 script to better compete with neovim's luajit. While how much varies according to the domain, in general people, especially technical people, value clear and obvious improvement/progress over never breaking backward compatibility or never dropping support for ancient/rare platforms, etc. You pay more for a better product, you wait longer in line for better seat or better amusement park ride, people are generally willing to do a little extra work to get a better result. In this case that would be addressing any breakage when updating a library across major versions. And if they're not or they have no need to, they can continue to use the old version with no issues.

All search engines produce LLM content as the top search result these days, so yes some of that was directly copied from those results. I'm not sure how that's relevant. Whether I copied from a website or LLM output it's the same info.

>Calling me inconsistent/hypocritical/pathological for using negative values that style_configurator itself allows me to do? What...?

I wasn't calling you inconsistent for using negative values for styling. I was calling you inconsistent for two reasons which I clearly stated:

1. For changing definitions of what you consider breaking
2. For advocating for a breaking change while calling it a non-breaking change (Changing the default styling for property buttons would "break" things exactly as much as my change here setting button height to text height, arguably more so, regardless of if the new styling default is negative or not).

As an aside, why would I limit the Configurator to positive values? It was (and still is primarily, regardless of my current DRAFT) a method to understand how the styling works not a school marm enforcing (unwritten) rules about about styling. And even if I think we should never have a negative padding as a default, I'm a big believer in letting people do what they want with their code. It's why I don't really see the point of private members. If someone wants to use negative padding, have at it.

> You are repeatedly bringing low effort changes that nobody else asked for, and each time I ask you to just keep some old stuff working, you burst into tears and post a wall of text about it.

The difference between professional/production quality products that people actually use/love and amateur/hobby projects is not the big things, it's 100s of small things, small frustrations, quirks, minor bugs, slow-downs, etc. The difference is polish and refinement. Nuklear is severely lacking in that department. Most of the big changes Nuklear needs are ... breaking changes and would be much harder to get in than many smaller changes. You can't correct the Tower of Pisa by continuing to build up, you have to be willing to just do things and move forward. The only exceptions are complete self-contained new widgets like the knob introduced awhile back but we don't need much of that. We need refinement of what we do have to work better more than we need new stuff. 

PR's like #466 that stall out are a perfect example of this. It is a massive improvement in working with images. It works for some backends but not all of them; it stalls because we don't want to break anything, not even to the level of "this feature is only supported for vertex buffer based backends". Someone else does a bit of work on it a year later, making minor fixes; stalls. Original developer comes back, rebases, new demo. Then it get's picked up again by someone else (me), try to start working around some backend limitaitons, get frustrated with all the issues in the code that would be "breaking" to rework/rename etc. and it stalls again because who likes doing a ton of work just to have it all be for nothing when it's rejected.

There's a reason the best OSS projects have always had strong BDFL's (or at most 2-3 core team members) driving them forward with a specific vision/philosophy, especially in the beginning before it becomes more stable. It's the only way to get the best of both sides of that saying "If you want to go fast, go alone. If you want to go far, go together" or the similar Rudyard Kipling poem opening, "Down to Gehenna or up to the Throne, he travels fasted who travels alone."

> Alright, I've seen enough. I won't be replying to you any more. Do whatever.

That is of course your prerogative. Of course as long as I contribute to Nuklear I will continue to advocate for improvements regardless of whether they are breaking or not. I will likely get more forceful over time until things change or I leave. That's just human nature.

RobLoach:
466! YES :heart: ... As for breaking changes, we should not keep ourselves from making smart changes to help push the library forwards. Holding ourselves back will only stagnate it, as you've mentioned. Movement is essential.

Appreciate both of your contributions. Let's keep focus on pushing things forwards.

RobLoach:
Going to merge this in. If this design change causes issues, we can tweak it.

Issue #421, opened 2022-03-13
Correct way of ensuring focus of widgets?


tree786isback:
Hello

I am very new to nuklear, and attempting at integrating it into an OpenGL 3.3 c++ game framework which I have been using. I've pretty much finished the integration, mostly seems to be working fine. However I am experiencing strange issues when it comes to widgets taking focus, and not allowing my main game loop execute custom key callbacks. To better clarify the issue, when I try and select an empty area in my game window, I always seem to get a return value of 1 from `nk_window_focus`:
![Peek 2022-03-13 14-37](https://user-images.githubusercontent.com/97201835/158064644-34f4701d-686b-4a13-88a5-d3e00ab0a053.gif)
I.e. in the GIF, `nk_window_focus` still returns 1, even when I don't select the floating nuklear window. Here is my code which is running on every update (snippets taken from nuklear demos):
```
    nk_glfw3_new_frame(&glfw);
    /* GUI */
    if (nk_begin(ctx, "Demo", nk_rect(50, 50, 230, 250),            NK_WINDOW_BORDER|NK_WINDOW_MOVABLE|NK_WINDOW_SCALABLE|            NK_WINDOW_MINIMIZABLE|NK_WINDOW_TITLE)){
        enum {EASY, HARD};
        static int op = EASY;

        static int property = 20;
        nk_layout_row_static(ctx, 30, 80, 1);
        if (nk_button_label(ctx, "button")){
          std::cout << "pressed!" << '\n';
        }

        nk_layout_row_dynamic(ctx, 30, 2);
        if (nk_option_label(ctx, "easy", op == EASY)) op = EASY;
        if (nk_option_label(ctx, "hard", op == HARD)) op = HARD;

        nk_layout_row_dynamic(ctx, 25, 1);
        nk_property_int(ctx, "Compression:", 0, &property, 100, 10, 1);

        nk_layout_row_dynamic(ctx, 20, 1);
        nk_label(ctx, "background:", NK_TEXT_LEFT);
        nk_layout_row_dynamic(ctx, 25, 1);
        if (nk_combo_begin_color(ctx, nk_rgb_cf(bg), nk_vec2(nk_widget_width(ctx),400))) {
            nk_layout_row_dynamic(ctx, 120, 1);
            bg = nk_color_picker(ctx, bg, NK_RGBA);
            nk_layout_row_dynamic(ctx, 25, 1);
[...]

PR #926, opened 2026-03-24
Typos and minor issues

rswinkle:
I broke this up into 3 separate commits

vertexes -> vertices

Remove comment about non-existent function nk_window_focus(). I can't find any evidence it ever existed, even in the original repo. All I could find was [this](https://github.com/Immediate-Mode-UI/Nuklear/issues/421) which doesn't really make sense since even searching the original repo all I could find is [the comment](https://github.com/search?q=repo%3Avurtun%2Fnuklear+nk_window_focus&type=code). Very strange. Anyone know what's going on here? Is there a programmatic window focus function by another name?

NK_EDIT_COMMITED -> NK_EDIT_COMMITTED
Every time I haven't used an edit field in a while I forget that it's spelled wrong. I'm fine to appease @sleeptightAnsiC and just do this for now instead:

```
enum nk_edit_events {
    NK_EDIT_ACTIVE      = NK_FLAG(0), /**!< edit widget is currently being modified */
    NK_EDIT_INACTIVE    = NK_FLAG(1), /**!< edit widget is not active and is not being modified */
    NK_EDIT_ACTIVATED   = NK_FLAG(2), /**!< edit widget went from state inactive to state active */
    NK_EDIT_DEACTIVATED = NK_FLAG(3), /**!< edit widget went from state active to state inactive */
    NK_EDIT_COMMITED   = NK_FLAG(4)  /**!< old deprecated name for NK_EDIT_COMMITTED */
    NK_EDIT_COMMITTED   = NK_FLAG(4)  /**!< edit widget has received an enter and lost focus */
};
```
but honestly... everyone just needs to accept that breaking changes and major version bumps are both inevitable and even a good thing, not something to be avoided at all costs. Otherwise stuff like this just spreads over time.

@RobLoach, sorry I meant to get both this and my other mini one in over the weekend, and then yesterday but my internet was acting up (or maybe just Github? hard to tell).  No hurry.

### DIFF
--- a/demo/common/overview.c
+++ b/demo/common/overview.c
@@ -640,7 +640,7 @@ overview(struct nk_context *ctx)
                 nk_layout_row(ctx, NK_STATIC, 25, 2, ratio);
                 active = nk_edit_string(ctx, NK_EDIT_FIELD|NK_EDIT_SIG_ENTER, text[7], &text_len[7], 64,  nk_filter_ascii);
                 if (nk_button_label(ctx, "Submit") ||
-                    (active & NK_EDIT_COMMITED))
+                    (active & NK_EDIT_COMMITTED))
                 {
                     text[7][text_len[7]] = '\n';
                     text_len[7]++;
--- a/src/nuklear.h
+++ b/src/nuklear.h
@@ -928,7 +928,7 @@ NK_API void nk_input_end(struct nk_context*);
  * hardware directly. Therefore it is possible to just define
  * `NK_INCLUDE_VERTEX_BUFFER_OUTPUT` which includes optional vertex output.
  * To access the vertex output you first have to convert all draw commands into
- * vertexes by calling `nk_convert` which takes in your preferred vertex format.
+ * vertices by calling `nk_convert` which takes in your preferred vertex format.
  * After successfully converting all draw commands just iterate over and execute all
  * vertex draw commands:
  *
@@ -1059,7 +1059,7 @@ NK_API const struct nk_command* nk__next(struct nk_context*, const struct nk_com
 
 /**
  * \brief Converts all internal draw commands into vertex draw commands and fills
- * three buffers with vertexes, vertex draw commands and vertex indices.
+ * three buffers with vertices, vertex draw commands and vertex indices.
  *
  * \details
  * The vertex format as well as some other configuration values have to be
@@ -4770,7 +4770,7 @@ NK_API nk_bool nk_input_is_key_down(const struct nk_input*, enum nk_keys);
  * \page "Draw List"
  * The optional vertex buffer draw list provides a 2D drawing context
  * with antialiasing functionality which takes basic filled or outlined shapes
- * or a path and outputs vertexes, elements and draw commands.
+ * or a path and outputs vertices, elements and draw commands.
  * The actual draw list API is not required to be used directly while using this
  * library since converting the default library draw command output is done by
  * just calling `nk_convert` but I decided to still make this library accessible
--- a/src/nuklear.h
+++ b/src/nuklear.h
@@ -3539,8 +3539,11 @@ enum nk_edit_events {
     NK_EDIT_INACTIVE    = NK_FLAG(1), /**!< edit widget is not active and is not being modified */
     NK_EDIT_ACTIVATED   = NK_FLAG(2), /**!< edit widget went from state inactive to state active */
     NK_EDIT_DEACTIVATED = NK_FLAG(3), /**!< edit widget went from state active to state inactive */
-    NK_EDIT_COMMITED    = NK_FLAG(4)  /**!< edit widget has received an enter and lost focus */
+    NK_EDIT_COMMITTED   = NK_FLAG(4)  /**!< edit widget has received an enter and lost focus */
 };
+
+#define NK_EDIT_COMMITED NK_EDIT_COMMITTED
+
 NK_API nk_flags nk_edit_string(struct nk_context*, nk_flags, char *buffer, int *len, int max, nk_plugin_filter);
 NK_API nk_flags nk_edit_string_zero_terminated(struct nk_context*, nk_flags, char *buffer, int max, nk_plugin_filter);
 NK_API nk_flags nk_edit_buffer(struct nk_context*, nk_flags, struct nk_text_edit*, nk_plugin_filter);
--- a/src/nuklear.h
+++ b/src/nuklear.h
@@ -1178,7 +1178,7 @@ NK_API const struct nk_draw_command* nk__draw_next(const struct nk_draw_command*
  * order. The topmost window thereby is the currently active window.<br /><br />
  *
  * To change window position inside the stack occurs either automatically by
- * user input by being clicked on or programmatically by calling `nk_window_focus`.
+ * user input by being clicked on or programmatically by calling `nk_window_set_focus`.
  * Windows by default are visible unless explicitly being defined with flag
  * `NK_WINDOW_HIDDEN`, the user clicked the close button on windows with flag
  * `NK_WINDOW_CLOSABLE` or if a window was explicitly hidden by calling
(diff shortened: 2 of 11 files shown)


demo/common/overview.c
```diff
@@ -640,7 +640,7 @@ overview(struct nk_context *ctx)
                 nk_layout_row(ctx, NK_STATIC, 25, 2, ratio);
                 active = nk_edit_string(ctx, NK_EDIT_FIELD|NK_EDIT_SIG_ENTER, text[7], &text_len[7], 64,  nk_filter_ascii);
                 if (nk_button_label(ctx, "Submit") ||
-                    (active & NK_EDIT_COMMITED))
+                    (active & NK_EDIT_COMMITTED))
```
RobLoach, line 643: It's listed as `NK_EDIT_COMMITED` in `nk_edit_events`. May need to change it there?

Along with a backwards compatibility...
```
#define NK_EDIT_COMMITED NK_EDIT_COMMITTED
```

RobLoach (changes requested):

sleeptightAnsiC:
> Remove comment about non-existent function nk_window_focus(). I can't find any evidence it ever existed, even in the original repo.

This function does exist. The name was simply misspelled in the doc. It should be: [nk_window_set_focus](https://github.com/Immediate-Mode-UI/Nuklear/blob/9c24f9c2b9434df3073d7bf66ccdb8043526adb3/src/nuklear.h#L1835-L1848)

> I'm fine to appease sleeptightAnsiC and just do this for now instead [...]

Just add something like

```c
/* for backwards compatibility */
#define NK_EDIT_COMMITED NK_EDIT_COMMITTED
```
right below nk_edit_events definitions, and no one will have any problems with this.

EDIT: Rob posted his comment like 5sec before mine with the same suggestion.

> [...] but honestly... everyone just needs to accept that breaking changes and major version bumps are both inevitable and even a good thing, not something to be avoided at all costs.

Trying to snuck breakage under "_fixed typo_" is pointless. This is easily avoidable.

src/nuklear.h
```diff
@@ -1177,8 +1177,8 @@ NK_API const struct nk_draw_command* nk__draw_next(const struct nk_draw_command*
  * is linked into a window stack list which determines the drawing and overlapping
  * order. The topmost window thereby is the currently active window.<br /><br />
  *
- * To change window position inside the stack occurs either automatically by
- * user input by being clicked on or programmatically by calling `nk_window_focus`.
+ * Changing the window position inside the stack occurs automatically via
+ * user input, ie by being clicked.
```
sleeptightAnsiC: per https://github.com/Immediate-Mode-UI/Nuklear/pull/926#issuecomment-4121038417, could you please revert this, and simply change:

`nk_window_focus` ->  `nk_window_set_focus`

The description fits the logic of that function perfectly.

rswinkle:
> > Remove comment about non-existent function nk_window_focus(). I can't find any evidence it ever existed, even in the original repo.
> 
> This function does exist. The name was simply misspelled in the doc. It should be: [nk_window_set_focus](https://github.com/Immediate-Mode-UI/Nuklear/blob/9c24f9c2b9434df3073d7bf66ccdb8043526adb3/src/nuklear.h#L1835-L1848)

I swear I search for and didn't find that function years ago when I first noticed the comment. This is what I get for making notes about things with the intention to do something about them eventually, and then when I finally do something, I don't do as much due diligence as I should sometimes because surely I did it earlier. Mea culpa.

> > I'm fine to appease sleeptightAnsiC and just do this for now instead [...]
> 
> Just add something like
> 
> ```c
> /* for backwards compatibility */
> #define NK_EDIT_COMMITED NK_EDIT_COMMITTED
> ```
> 
> right below nk_edit_events definitions, and no one will have any problems with this.
> 
> EDIT: Rob posted his comment like 5sec before mine with the same suggestion.
> 
> > [...] but honestly... everyone just needs to accept that breaking changes and major version bumps are both inevitable and even a good thing, not something to be avoided at all costs.
> 
> Trying to snuck breakage under "_fixed typo_" is pointless. This is easily avoidable.

I wasn't trying to sneak it in otherwise I wouldn't have explicitly called your attention to it. And my comment was not about this issue alone as I said, and it wouldn't matter what my commit was about, your opinion would be the same regarding any breaking change. Am I wrong?

It's a chronic problem that adds up. Sure an extra define/enum here is harmless, especially relative to most of these "do we break things or not" debate. However, we're just making our lives harder because when we finally do start making breaking changes we'll never remember all the things we *wanted* to break but were just saving to bundle together, and we'll forget any hacks/workarounds we make and any sacrifices in functionality we made on the altar of, "well we can't do that, that would break things."

demo/common/overview.c
rswinkle: I did change it there?  I'm adding the define but I'm not sure what your first sentence is referring to. There were only 3 places I could find it and I changed all 3 (overview, nk_edit_events, and nk_do_edit()).

EDIT: i can spell good

rswinkle:
On the plus side, at least I'm getting more practice with interactive rebasing and editing commits.

RobLoach:
Don't worry about rebasing things if you're concerned about conflicts or anything. Github has a squash and merge button if you're worried about the git commits being hairy.
Screenshot from 2026-03-24 19-59-05

sleeptightAnsiC:
@RobLoach No need to squash it here. The separation in this PR is quite good.

You usually want to keep unrelated stuff in different commits when patching something important. Otherwise, doing things like `git bisect` or `git blame` becomes problematic.

demo/common/overview.c
RobLoach: excellent, thanks!

RobLoach (approved):

sleeptightAnsiC:
> > > remove comment about non-existent function 
> >
> > This function does exist. The name was simply misspelled in the doc.
>
> I swear I search for and didn't find that function [...] Mea culpa.

No worries.

> it wouldn't matter what my commit [comment?] was about, your opinion would be the same regarding any breaking change. Am I wrong?

I agreed for a change here, and under https://github.com/Immediate-Mode-UI/Nuklear/pull/912, and those are still technically "_breaking_". I only care about old C code to remain compilable. Don't force people to patch their code and I won't complain.

I'm not entirely sure, why are you trying to start discussion about "_breaking changes_" here, under PR about typos. This is probably the worst place to discuss such things. This time, even Rob was faster than me with requesting the old symbol to remain. Fixing the typo is the weakest argument for breaking anything.

Keeping the old symbol by introducing a macro is a good enough compromise. I get my fake compatibility and you get your pedantic fix. If it was up to me, I wouldn't allow even that, but I'm not in charge here. If someone gets affected by this change anyway, well, it's not in my interest to waste my free time arguing about it.

sleeptightAnsiC (approved):
Thanks.

rswinkle:
> I agreed for a change here, and under #912, and those are still technically "_breaking_". I only care about old C code to remain compilable. Don't force people to patch their code and I won't complain.

You agreed to a non-breaking change, same as #912 where we explicitly discussed how it was non-breaking according to the definition you give right here: the developer can update the library and not be forced to change anything. And for nth time, updating the library does not force people to patch their code. The library does not magically update itself in there repos. This isn't libc. Heck, even with dynamic libraries, between flatpaks, appimages, snaps, and plain old-fashioned tarball packages, no one is *forced* to update anything under almost any circumstances these days.

> I'm not entirely sure, why are you trying to start discussion about "_breaking changes_" here, under PR about typos. This is probably the worst place to discuss such things. This time, even Rob was faster than me with requesting the old symbol to remain. Fixing the typo is the weakest argument for breaking anything.

I'm trying to wear you down. If I can't convince with logic and common sense it's my only option. Looks like I have another opportunity in another PR today. I'm only half joking.  And actually I was faster than Rob, considering I listed an equivalent change to keep it backward compatible in my very first message. The typo is not the argument. The argument is the same regardless of the message carrier. Major updates are normal and semver exists for a reason. Your personal preference to both always use the latest version of Nuklear *and* never have to make any changes to do so doesn't change that.

> If it was up to me, I wouldn't allow even that, but I'm not in charge here.

Exactly. Not only would you never agree to *any* breaking change, but if you were in charge you would resist even non-breaking changes that *could've* been breaking without a concession. That points to something almost pathological imo.

I am posting this at the same time I comment on the other PR. They sort of go together.

PR #928, opened 2026-03-25
demo: Fix sdl_vulkan for Windows, from @gravnem

RobLoach:
Fixes #923

### DIFF
--- a/demo/sdl_vulkan/main.c
+++ b/demo/sdl_vulkan/main.c
@@ -811,7 +811,7 @@ bool create_swap_chain(struct vulkan_demo *demo) {
     }
 
     create_info.preTransform = swap_chain_support.capabilities.currentTransform;
-        
+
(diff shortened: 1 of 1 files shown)


sleeptightAnsiC:
apparently, SDL requires specific VK_KHR extension which I don't have under my Win11 VM. I may try with real hardware later, but tbh this patch is fine even without retesting.

@RobLoach In the mean time, would you mind adding gravnem as co-author?

```
Co-authored-by: gvanem <gvanem@yahoo.no>
```

He pretty much provided us with diffs ready to be applied so would be nice to properly credit him. I hope it's not a problem. If in trouble, see [Creating a commit with multiple authors](https://docs.github.com/en/pull-requests/committing-changes-to-your-project/creating-and-editing-commits/creating-a-commit-with-multiple-authors)

PR #929, opened 2026-03-25
cmake: Add basic CMake setup

### LINKED ISSUE #731: pls add cmake support
subj

RobLoach:
While there is a large CMake set up over at https://github.com/Immediate-Mode-UI/Nuklear/pull/462 , this takes on a simple approach of having a basic file so that it's easy to install/use from CMake.

Fixes #731


SamuelMarks:
What about all the `option`s? - I did make that other PR, if it's too big tell me what size you want for Nuklear.

PR #930, opened 2026-03-26
docs: Clean up the packer

RobLoach:
This makes some changes to the way the packer works so that we don't duplicate as much documentation.

- Clean up the header docs
- Make it work cleaner with Doxygen
- Use the README.md in the source so that we don't have to duplicate it in the header


src/README.md
```diff
@@ -0,0 +1,8 @@
+# File Packer
+
+This will build out [nuklear.h](../nuklear.h).
+
+## Usage
+
+- On Linux/Mac just run `./paq.sh > ../nuklear.h`
```
rswinkle: This is a good opportunity to fix this inaccuracy.  It should just say:
```
- On Linux/Mac just run `./paq.sh`
```
paq.sh has no output and hard codes > ../nuklear.h and has for as long as I can remember but we never updated the readme, probably because it worked either way.

rswinkle (changes requested):
Looks good, other than the minor paq.sh issue. I even ran Doxygen. The new site will look nice once we're ready to update it.

Image

src/README.md
RobLoach:
``` suggestion
- On Linux/Mac, run `./paq.sh`
```

PR #931, opened 2026-03-26
glfw: Fix scaling in Vulkan with Retina

RobLoach:
Fix #918


rswinkle (approved):
Again, no HiDPI/Retina display but everything looks and runs fine here. Maybe I should've splurged when I got my latest monitors just for testing purposes.

PR #932, opened 2026-03-26
glfw: Use framebuffer size

### LINKED ISSUE #325: issue regarding example
I see in examples Window size is being passed to glViewport, also just found this warning
image

RobLoach:
Fixes #325


rswinkle (approved):
I don't have a HiDPI monitor to really verify the change but it looks fine and they compile and run the same on my normal one.

PR #933, opened 2026-03-26
button: Add chevron and hamburger symbol

RobLoach:
This adds chevron and hamburger symbols that can be used.

Screenshot from 2026-03-26 12-58-12


rswinkle (approved):
I only ran compiled and ran it in a few backends but I don't think we really need more. The code looks pretty simple. Nice with the Hamburger, now we can have a menu button like that without having to resort to a separate image or Awesome Font.

rswinkle (approved):
Take two, actually checking the "Viewed" check boxes this time.

RobLoach:
Four questions....

1. Should we switch the X from using the font to draw an actual X?
2. Sleep also brought up a good point that the line thickness is hard-coded. Think we should change that?
3. Any other symbols you could think of?
4. Any places you think these should be used?

rswinkle:
> Four questions....
> 
> 1. Should we switch the X from using the font to draw an actual X?

Hmm, either that or at least change to using a capital 'X'?
> 2. Sleep also brought up a good point that the line thickness is hard-coded. Think we should change that?

If he's commented in this PR I can't see it.

But if you're talking about something like this:
``` 
nk_draw_symbol(out, type, *content, bg, sym, 1, font);
to
nk_draw_symbol(out, type, *content, bg, sym, style->border, font);
or
nk_draw_symbol(out, type, *content, bg, sym, (style->border > 0.0f) ? style->border : 1.0f, font);

Or maybe just putting that ternary statement at the top of draw_symbol then we can decide about the individual calls to draw_symbol whenever (there's only about 5 of them).
```
I think that's fine.
> 3. Any other symbols you could think of?

Not anything as useful and easy to make as the hamburger. The only maybe I see [here](https://fontawesome.com/search?c=coding&ic=free-collection) is the staggered hamburger. Probably not worth it.
> 4. Any places you think these should be used?

Besides what you added to Overview? I mean maybe some people would like them better for tree/tabs (why are they called trees but the style struct in nk_style_tab?) instead of our current filled triangles?

RobLoach:
Moved the X over to vector drawing, which does look better than when you're using a funky font. Going to keep the expander as the triangle. While it does work, I think it's too big of a UX change, at least as part of this.

Screenshot from 2026-05-26 11-59-21

PR #934, opened 2026-03-28
Update configurator and more

rswinkle:
As always I can't start anything without finding other things, so once again I've broken it up into separate commits and still left work for a more serious PR later

1. Update configurator, it was missing a decent amount including knob entirely and new symbols.
2. Fix bugs in knob
3. Fix some issues in slider

The more serious PR for later is to fix nk_convert() to correctly draw our rect based "circles" using both width and height. And maybe some renaming and new functions/enums should be in order in several places as a "circle" implies a uniform radius but other than this bug in nk_convert, we treat it as an axis aligned ellipse placed in a designated rect.

@RobLoach I know you've been super active with a bunch of PRs in the last few days. I'll try to review and test out a few more this weekend.

### DIFF
--- a/demo/common/style_configurator.c
+++ b/demo/common/style_configurator.c
@@ -20,7 +20,11 @@ static const char* symbols[NK_SYMBOL_MAX] =
     "TRIANGLE_LEFT",
     "TRIANGLE_RIGHT",
     "PLUS",
-    "MINUS"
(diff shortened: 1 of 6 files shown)


RobLoach (approved):
Was able to run the configurator just fine. Thanks.

Issue #939, opened 2026-04-11
Various SDL3 Renderer backend issues


rswinkle:
It does not compile as C++ due to missing casts.

It should use key codes not scancodes (half our backends use key codes and the ones that don't like glfw are because it's difficult/painful to do so; we have to use the key callback for proper key repeat behavior already).

It does not need the timing method for detecting a double click, that's what evt->button.clicks is for.

I think that's it.

PR #940, opened 2026-04-11
Fix minor sdl3_renderer issues

rswinkle:
Fixes issue #939 

@RobLoach @sleeptightAnsiC

### DIFF
--- a/demo/sdl3_renderer/nuklear_sdl3_renderer.h
+++ b/demo/sdl3_renderer/nuklear_sdl3_renderer.h
@@ -54,13 +54,6 @@ NK_API struct nk_allocator  nk_sdl_allocator(void);
 #ifndef NK_SDL3_RENDERER_IMPLEMENTATION_ONCE
 #define NK_SDL3_RENDERER_IMPLEMENTATION_ONCE
 
-#ifndef NK_SDL_DOUBLE_CLICK_LO
-#define NK_SDL_DOUBLE_CLICK_LO 0.02
-#endif
-#ifndef NK_SDL_DOUBLE_CLICK_HI
-#define NK_SDL_DOUBLE_CLICK_HI 0.2
-#endif
-
 struct nk_sdl_device {
     struct nk_buffer cmds;
     struct nk_draw_null_texture tex_null;
@@ -84,7 +77,6 @@ struct nk_sdl {
 #endif
     struct nk_allocator allocator;
     nk_handle userdata;
-    Uint64 last_left_click;
     Uint64 last_render;
     bool insert_toggle;
     bool edit_was_active;
@@ -362,7 +354,7 @@ nk_sdl_clipboard_copy(nk_handle usr, const char *text, int len)
     buflen = (size_t)(ptext - text) + 1;
 #endif
 
-    str = sdl->allocator.alloc(sdl->allocator.userdata, 0, buflen);
+    str = (char*)sdl->allocator.alloc(sdl->allocator.userdata, 0, buflen);
     if (!str) return;
     SDL_strlcpy(str, text, buflen);
     SDL_SetClipboardText(str);
@@ -377,7 +369,7 @@ nk_sdl_init(SDL_Window *win, SDL_Renderer *renderer, struct nk_allocator allocat
     NK_ASSERT(renderer);
     NK_ASSERT(allocator.alloc);
     NK_ASSERT(allocator.free);
-    sdl = allocator.alloc(allocator.userdata, 0, sizeof(*sdl));
+    sdl = (struct nk_sdl*)allocator.alloc(allocator.userdata, 0, sizeof(*sdl));
     NK_ASSERT(sdl);
     SDL_zerop(sdl);
     sdl->allocator.userdata = allocator.userdata;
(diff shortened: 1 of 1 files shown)


demo/sdl3_renderer/nuklear_sdl3_renderer.h
```diff
@@ -512,14 +503,12 @@ nk_sdl_handle_event(struct nk_context* ctx, SDL_Event *evt)
...
-                const double dt = (double)(evt->button.timestamp - sdl->last_left_click) / 1000000000.0;
                 switch(evt->button.button)
                 {
                     case SDL_BUTTON_LEFT:
+                        if (evt->button.clicks > 1)
+                            nk_input_button(ctx, NK_BUTTON_DOUBLE, x, y, down);
                         nk_input_button(ctx, NK_BUTTON_LEFT, x, y, down);
-                        nk_input_button(ctx, NK_BUTTON_DOUBLE, x, y,
-                                down && dt > NK_SDL_DOUBLE_CLICK_LO && dt < NK_SDL_DOUBLE_CLICK_HI);
-                        sdl->last_left_click = evt->button.timestamp;
```
sleeptightAnsiC, line 522: (from https://github.com/Immediate-Mode-UI/Nuklear/issues/939)
> It does not need the timing method for detecting a double click, that's what evt->button.clicks is for.

I was specifically asked to add this in https://github.com/Immediate-Mode-UI/Nuklear/pull/779#discussion_r2328703697

When using `evt->button.clicks > 1`, all consecutive clicks trigger double-click. That's why I went with the timing approach, even though I was skeptic about it at first.

Don't have any hard feelings about this change. Both approaches aren't perfect and the library doesn't even use double-click anywhere. Just saying.

demo/sdl3_renderer/nuklear_sdl3_renderer.h
```diff
@@ -452,50 +443,50 @@ nk_sdl_handle_event(struct nk_context* ctx, SDL_Event *evt)
...
                 int down = evt->type == SDL_EVENT_KEY_DOWN;
-                int ctrl_down = evt->key.mod & (SDL_KMOD_LCTRL | SDL_KMOD_RCTRL);
+                int ctrl_down = evt->key.mod & SDL_KMOD_CTRL;
 
-                /* In 99% of the time, you want to use scancodes, not real key codes,
-                 * see: https://wiki.libsdl.org/SDL3/BestKeyboardPractices */
-                switch(evt->key.scancode)
+                /* Scancodes usually make more sense for games, but key codes are better for
+                 * general application */
+                switch(evt->key.key)
```
sleeptightAnsiC, line 456: (from https://github.com/Immediate-Mode-UI/Nuklear/issues/939)

> It should use key codes not scancodes (half our backends use key codes and the ones that don't like glfw are because it's difficult/painful to do so; we have to use the key callback for proper key repeat behavior already).

I don't remember why I decided on scancodes back when I wrote it.

I think I had an impression that common shortcuts are tied to scancodes but apparently this is not so simple according to: https://stackoverflow.com/questions/22842638/dvorak-typers-how-do-you-handle-shortcuts-like-ctrl-c

Don't have any opinion about this change, but if it's here to stay, could you also remove the comment from lines 448-449? It doesn't really have a purpose any more.

demo/sdl3_renderer/nuklear_sdl3_renderer.h
```diff
@@ -362,7 +354,7 @@ nk_sdl_clipboard_copy(nk_handle usr, const char *text, int len)
     buflen = (size_t)(ptext - text) + 1;
 #endif
 
-    str = sdl->allocator.alloc(sdl->allocator.userdata, 0, buflen);
+    str = (char*)sdl->allocator.alloc(sdl->allocator.userdata, 0, buflen);
```
sleeptightAnsiC, line 365: (from https://github.com/Immediate-Mode-UI/Nuklear/issues/939)

> It does not compile as C++ due to missing casts.

Could you also wrap the whole header with:

```c
#ifdef __cplusplus
extern "C" {
#endif

/* ... */

#ifdef __cplusplus
}
#endif
```

I remember someone asked for this in https://github.com/Immediate-Mode-UI/Nuklear/issues/760#issue-2768929733 but it seems like I forgot about it back then.

demo/sdl3_renderer/nuklear_sdl3_renderer.h
rswinkle: Thanks to Rob's new Input section in Overview, it's pretty easy to test and verify that it works fine. As for GLFW, like with being forced to use scancodes due to its design, the need for the manual timer method for double clicks is because GLFW has no built in way to distinguish a single click from a double click.

demo/sdl3_renderer/nuklear_sdl3_renderer.h
rswinkle: > I think I had an impression that common shortcuts are tied to scancodes but apparently this is not so simple according to: https://stackoverflow.com/questions/22842638/dvorak-typers-how-do-you-handle-shortcuts-like-ctrl-c

Yeah, I am a dvorak typer myself and you learn pretty quick that you should just get used the new position for shortcuts with the new layout, otherwise you'd have to change the shortcuts everywhere, system shortcuts, application shortcuts etc. Fortunately things like vim movements and  CTRL+C/V map out decently in dvorak.

> Don't have any opinion about this change, but if it's here to stay, could you also remove the comment from lines 448-449? It doesn't really have a purpose any more.

Sure.

demo/sdl3_renderer/nuklear_sdl3_renderer.h
rswinkle: Sure.

You don't really need to wrap the whole header, just the prototypes in the header section. The only other backend that actually has extern "C" (gdi_native_nuklear) does this for example. Might be a good idea to add that to all the rest at some point.

RobLoach (approved):
While I haven't gone through an extensive review, I think this should be appropriate.

sleeptightAnsiC:
@RobLoach mind merging? Retested under Linux and MSYS2. Everything is fine.

PR #942, opened 2026-04-20
demo: Use Nuklear math functions for examples

RobLoach:
This flips the fabs(), cos() and sin() functions to the Nuklear counterpart so that you don't get a warning when you don't compile with `m`.


rswinkle (approved):
Looks good.

PR #944, opened 2026-04-21
fix: nk_option_text* unticking itself when not visible

### LINKED ISSUE #853: nk_option_text return value inconsistant
The nk_option_text return value is inconsistant, based on whether the widget is on- or off-screen.

nk_option_text will return 0 (state), if the widget is not on-screen. 
It will return is_active (the widget select state, after processing) when the widget is on-screen.

Instead of returning state when off-screen, it should return is_active.  There's no way to know from the call that the reason it is returning 0 is because the widget was off-screen.

The code, I think, should read:
    state = nk_widget(&bounds, ctx);
    if (!state) return (int)is_active; // this is currently: if (!state) return (int)state;

This will return a "proper" state, whether drawn or not.  

Currently, the following code will not work because my_bool will become 0 if the widget is off-screen.
my_bool = nk_option_text(ctx, "bool" my_bool);

With the suggested change, my_bool is only altered if the users alters the state of the widget.

The same holds true for nk_option_text_align

sleeptightAnsiC:
Every branch in those functions returns `is_active` but not those two,
which was clearly a bug that could manifest in rare occasions.
See linked issue for more information.

Fixes: https://github.com/Immediate-Mode-UI/Nuklear/issues/853

### DIFF
--- a/nuklear.h
+++ b/nuklear.h
@@ -25466,7 +25466,7 @@ nk_option_text(struct nk_context *ctx, const char *text, int len, nk_bool is_act
     layout = win->layout;
 
     state = nk_widget(&bounds, ctx);
-    if (!state) return (int)state;
+    if (!state) return is_active;
     in = (state == NK_WIDGET_ROM || state == NK_WIDGET_DISABLED || layout->flags & NK_WINDOW_ROM) ? 0 : &ctx->input;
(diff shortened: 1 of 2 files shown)


RobLoach (approved):

sleeptightAnsiC:
no changes in latest force push, just added StewBC as commit co-author :)

Issue #895, opened 2026-02-10
todo: use same CFLAGS for all demos (general pass for each ./demo/*/Makefile)


sleeptightAnsiC:
We have some differences in how each demo is being build. This creates few problems mentioned below. I want to make a pass on all of them and make all compilation flags consistent. Here are the changes that I'm considering:

1. Ensure optimizations are disabled with `-O0`. Currently, few demos use `-O2` and waste our CI time. (reference: https://github.com/Immediate-Mode-UI/Nuklear/pull/894#issuecomment-3880979941 )
1. Ensure that debug symbols are enabled with `-g`. This will allow for better debugging without affecting compilation times as much. (I'm still not entirely sure, if it's worth it, but I think it is, GCC enables this by default anyway)
1. Ensure the same C standard with `-std=c99` (?) I think some people may disagree so I will quickly explain: Nuklear itself is C89 but the demos are not. Most of the demos already need C99 in some form or another, and even if you aren't getting any pedantic warnings with your setup, you will get them after using `#define INCLUDE_ALL`. We can keep the backend headers C89 compliant, but we should compile the demos with C99. This was allowed many years ago with https://github.com/Immediate-Mode-UI/Nuklear/commit/177f391347bf1c13dd3b3297e39ea16fa0041b18 and if someone disagree, then we should consider rewriting the demos with C89 in mind and reverting said rule, instead of lying about the support with `-std=c89`. Right now, people are wasting their time for submitting "_warning fixes_" that someone is about to reintroduce anyway. **(EDIT: after grepping the demos, it seems like most are using `-std=c89` so maybe my take wasn't so accurate after all, I still think we should talk about it though)**
1. Ensure we have the same warnings enabled with `-Wall -Wextra -pedantic`. Any ideas for other flags? We haven't tested with anything else, but maybe we should...
1. Add missing CIs as few demos do not have them (and this is must have when changing so much stuff)
1. I could also fix this one at the same time: https://github.com/Immediate-Mode-UI/Nuklear/issues/813
1. More... ?

Any thoughts?

cc @RobLoach @rswinkle

sleeptightAnsiC:
Given that this would be a pretty big change, I will probably separate it into multiple PRs, starting with points 1 and 2 as first PR (because these are the most obvious ones).

grepping shows me there are only ~~three~~ **five** demos with optimizations enabled, so it shouldn't be such a big deal...
```
[...]

PR #946, opened 2026-04-21
disable optimizations and other costly CFLAGS from demos

sleeptightAnsiC:
Demos must compile as fast as possible, in order to be quick to reiterate and retest. Given that each needs to link and recompile amalgamation every time, the difference between -O2 and -O0 is noticeable. This also slightly affects CI times. This topic was discussed under several occasions, for example: https://github.com/Immediate-Mode-UI/Nuklear/pull/894#issuecomment-3880985649 or https://github.com/Immediate-Mode-UI/Nuklear/issues/895#issuecomment-3884228468

### DIFF
--- a/demo/d3d11/build.bat
+++ b/demo/d3d11/build.bat
@@ -3,7 +3,7 @@
 rem This will use VS2015 for compiler
 call "C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\vcvarsall.bat" x86
 
-fxc.exe /nologo /T vs_4_0_level_9_0 /E vs /O3 /Zpc /Ges /Fh nuklear_d3d11_vertex_shader.h /Vn nk_d3d11_vertex_shader /Qstrip_reflect /Qstrip_debug /Qstrip_priv nuklear_d3d11.hlsl
-fxc.exe /nologo /T ps_4_0_level_9_0 /E ps /O3 /Zpc /Ges /Fh nuklear_d3d11_pixel_shader.h /Vn nk_d3d11_pixel_shader /Qstrip_reflect /Qstrip_debug /Qstrip_priv nuklear_d3d11.hlsl
+fxc.exe /nologo /T vs_4_0_level_9_0 /E vs /Zpc /Ges /Fh nuklear_d3d11_vertex_shader.h /Vn nk_d3d11_vertex_shader /Qstrip_reflect /Qstrip_debug /Qstrip_priv nuklear_d3d11.hlsl
+fxc.exe /nologo /T ps_4_0_level_9_0 /E ps /Zpc /Ges /Fh nuklear_d3d11_pixel_shader.h /Vn nk_d3d11_pixel_shader /Qstrip_reflect /Qstrip_debug /Qstrip_priv nuklear_d3d11.hlsl
 
-cl /D_CRT_SECURE_NO_DEPRECATE /nologo /W3 /O2 /fp:fast /Gm- /Fedemo.exe main.c user32.lib dxguid.lib d3d11.lib /link /incremental:no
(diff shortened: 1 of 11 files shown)


demo/sdl3_renderer/Makefile
```diff
@@ -23,9 +23,9 @@ BINEXT := ${binext_${OSNAME}}
 TEMPDIR := ./bin
 BIN := ${TEMPDIR}/demo${BINEXT}
 
-cflags += -std=c89 -Wall -Wextra -Wpedantic
-cflags += -O2
-#cflags += -O0 -g
+cflags += -std=c89 -Wall -Wextra -pedantic
```
sleeptightAnsiC, line 26: changed `-Wpedantic` to `-pedantic` (without `W`) in order to keep it consistent with other demos since I was already touching these lines

demo/sdl_opengles2/Makefile
```diff
@@ -18,7 +19,7 @@ $(BIN): prepare
 	$(CC) $(SRC) $(CFLAGS) -o bin/$(BIN) $(LIBS)
 
 web: prepare
-	emcc $(SRC) -Os -s USE_SDL=2 -o bin/index.html
+	emcc $(SRC) $(CFLAGS) -s USE_SDL=2 -o bin/index.html
```
sleeptightAnsiC, line 22: Danger zone here:

this target was NOT reusing CFLAGS before, but since I don't see any reason against it, and since I was already touching it, I just added it here to keep it consistent with other targets.

I gave it a try and it compiles just fine. It yields a lot of warnings but nothing fatal.

sleeptightAnsiC:
> This also slightly affects CI times.

Actually, after taking a look at CI times of my last PRs, it's not just "slightly". From around [110](https://github.com/Immediate-Mode-UI/Nuklear/actions/runs/24733994082/job/72355815195?pr=945#:~:text=1m%2050s)-[117](https://github.com/Immediate-Mode-UI/Nuklear/actions/runs/24734319302/job/72356971719?pr=944#:~:text=1m%2057s)sec to [86](https://github.com/Immediate-Mode-UI/Nuklear/actions/runs/24742095641/job/72384173636?pr=946#:~:text=1m%2026s)sec, and that whole difference was taxed by just 3 out of 17 demos. Noooice.

rswinkle (approved):
Long overdue.

sleeptightAnsiC:
Converting to draft. I missed the examples compiled with MSVC so gotta revisit those too. Maybe there is even more https://github.com/search?q=repo%3AImmediate-Mode-UI%2FNuklear%20%2F%5C%2FO%5B1-9%5D%2F&type=code

EDIT: fixed in latest commit, hiding...

demo/d3d11/build.bat
```diff
@@ -3,7 +3,7 @@
 rem This will use VS2015 for compiler
 call "C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\vcvarsall.bat" x86
 
-fxc.exe /nologo /T vs_4_0_level_9_0 /E vs /O3 /Zpc /Ges /Fh nuklear_d3d11_vertex_shader.h /Vn nk_d3d11_vertex_shader /Qstrip_reflect /Qstrip_debug /Qstrip_priv nuklear_d3d11.hlsl
-fxc.exe /nologo /T ps_4_0_level_9_0 /E ps /O3 /Zpc /Ges /Fh nuklear_d3d11_pixel_shader.h /Vn nk_d3d11_pixel_shader /Qstrip_reflect /Qstrip_debug /Qstrip_priv nuklear_d3d11.hlsl
```
sleeptightAnsiC, line 7: I was skeptic about removing /O from FXC at first, because it changes checked shader blobs. However, those would change no matter what you do, because each version of FXC produces different output.

So I changed the flags but did not commit new blobs. I think, it's fine.

sleeptightAnsiC:
I intentionally split demo/*/build.bat changes into another commit.

RobLoach (approved):

PR #948, opened 2026-04-23
slider: Fix float type conversion

RobLoach:
Small fix for visual_cursor type conversion.

### DIFF
--- a/nuklear.h
+++ b/nuklear.h
@@ -25958,7 +25958,7 @@ nk_draw_slider(struct nk_command_buffer *out, nk_flags state,
     bar.h = style->bar_height;
 
     /* filled background bar style */
-    fill.w = visual_cursor->x+ 0.5*visual_cursor->w - bar.x;
+    fill.w = visual_cursor->x+ 0.5f*visual_cursor->w - bar.x;
     fill.x = bar.x;
     fill.y = bar.y;
     fill.h = bar.h;
--- a/nuklear.h
+++ b/nuklear.h
@@ -25958,7 +25958,7 @@ nk_draw_slider(struct nk_command_buffer *out, nk_flags state,
     bar.h = style->bar_height;
 
     /* filled background bar style */
-    fill.w = visual_cursor->x+ 0.5f*visual_cursor->w - bar.x;
+    fill.w = visual_cursor->x + 0.5f * visual_cursor->w - bar.x;
     fill.x = bar.x;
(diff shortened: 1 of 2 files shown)


src/nuklear_slider.c
```diff
@@ -86,7 +86,7 @@ nk_draw_slider(struct nk_command_buffer *out, nk_flags state,
     bar.h = style->bar_height;
 
     /* filled background bar style */
-    fill.w = visual_cursor->x+ 0.5*visual_cursor->w - bar.x;
+    fill.w = visual_cursor->x+ 0.5f*visual_cursor->w - bar.x;
```
rswinkle: Oh, my bad, sorry.  As long as we're making a fix here can you fix the spacing around the `+`?

`fill.w = visual_cursor->x + 0.5f*visual_cursor->w - bar.x;`

I don't really mind `a+b` or `a + b` and I do both depending on the situation but I can't believe I let that lopsided monstrosity through.

rswinkle (changes requested):

src/nuklear_slider.c
RobLoach: Personally prefer spaces on either side of every operator, but :shrug: ... What is represented elsewhere?
``` suggestion
    fill.w = visual_cursor->x + 0.5f * visual_cursor->w - bar.x;
```

nuklear.h
```diff
@@ -25958,7 +25958,7 @@ nk_draw_slider(struct nk_command_buffer *out, nk_flags state,
     bar.h = style->bar_height;
 
     /* filled background bar style */
-    fill.w = visual_cursor->x+ 0.5*visual_cursor->w - bar.x;
+    fill.w = visual_cursor->x+ 0.5f*visual_cursor->w - bar.x;
```
``` suggestion
    fill.w = visual_cursor->x + 0.5f * visual_cursor->w - bar.x;
```

rswinkle: > Personally prefer spaces on either side of every operator, but 🤷 ... What is represented elsewhere?

I think my natural tendency is to use no spaces to represent mental groupings to make it easier to read. I feel like longer expressions get harder to read if you use spaces everywhere and occasional tight couplings help you parse it better/more quickly, without the mess of unnecessary parenthesis.

But yes, I think spaces everywhere is more consistent with Nuklear's codebase in general, and for this function/file at the very least, so I think we're good.

PR #949, opened 2026-04-25
keys: Capture F1-F12, and ALT keys

RobLoach:
This allows checking if one of the function keys were pressed directly in Nuklear.


RobLoach:
Since this is just a bunch of enums, and the demo updates, I'm going to bring it in.

PR #952, opened 2026-05-02
images: Change stbi_load() to use 4 chans

### LINKED ISSUE #950: Images in X11 backend only works with RGBA
For example, `nk_xsurf_load_image_from_file("lena.png")` will display a garbled RGB image and segfault if you try to resize the window.

screenshot

```bash
$ file lena.png 
lena.png: PNG image data, 512 x 512, 8-bit/color RGB, non-interlaced
````

```c
/* --snip-- */
#define NK_IMPLEMENTATION
#define NK_XLIB_IMPLEMENTATION
#define NK_XLIB_INCLUDE_STB_IMAGE
#define NK_XLIB_IMPLEMENT_STB_IMAGE
#include "../../nuklear.h"
#include "nuklear_xlib.h"

/* --snip-- */
    struct nk_image img = nk_xsurf_load_image_from_file("lena.png");

    while (running)
    {
        /* --snip-- */
        if (nk_begin(ctx, "Image", nk_rect(0, 0, 300, 300), NK_WINDOW_SCALABLE))
        {
            float win_h = nk_window_get_content_region(ctx).h;
            nk_layout_row_dynamic(ctx, win_h, 1);
            nk_image(ctx, img);
        }
        nk_end(ctx);
        /* --snip-- */
```

RobLoach:
Switches the stb image loads to default to 4 channels.

Fixes #950

@TomJGooding Like this?

### DIFF
--- a/demo/common/file_browser.c
+++ b/demo/common/file_browser.c
@@ -105,7 +105,7 @@ icon_load(const char *filename)
 {
     int x,y,n;
     GLuint tex;
-    unsigned char *data = stbi_load(filename, &x, &y, &n, 0);
+    unsigned char *data = stbi_load(filename, &x, &y, &n, 4);
     if (!data) die("[SDL]: failed to load image: %s", filename);
 
     glGenTextures(1, &tex);
--- a/demo/x11/nuklear_xlib.h
+++ b/demo/x11/nuklear_xlib.h
@@ -556,7 +556,7 @@ nk_xsurf_load_image_from_file(char const *filename)
 {
     int x,y,n;
     unsigned char *data;
-    data = stbi_load(filename, &x, &y, &n, 0);
+    data = stbi_load(filename, &x, &y, &n, 4);
     return nk_stbi_image_to_xsurf(data, x, y, n);
 }
 #endif /* NK_XLIB_INCLUDE_STB_IMAGE */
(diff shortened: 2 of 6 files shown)


TomJGooding:
Sorry but I really don't feel qualified to say whether this is the proper fix as mentioned in https://github.com/Immediate-Mode-UI/Nuklear/issues/950#issuecomment-4364480410!

But wouldn't this only work if we presuppose the [DefaultDepth](https://tronche.com/gui/x/xlib/display/display-macros.html#DefaultDepth) is 24?  I don't fully understand what the `switch (depth)` is doing in `nk_stbi_image_to_xsurf`?

sleeptightAnsiC:
This is more of a workaround that hides the actual bug. Not the proper fix.

I haven't checked everything yet, but [these two branches in nk_stbi_image_to_xsurf()](https://github.com/Immediate-Mode-UI/Nuklear/blob/239f723ddb38a86000115ee0958fbf65603e433c/demo/x11/nuklear_xlib.h#L502-L531) are probably the root cause.

~~@TomJGooding mind sharing the file that caused it? `"lena.png"` ?~~ EDIT: got one from https://github.com/mikolalysenko/lena/blob/master/lena.png

EDIT EDIT: another PR at https://github.com/Immediate-Mode-UI/Nuklear/pull/954

PR #954, opened 2026-05-05
fix(x11): nk_stbi_image_to_xsurf corrupting memory when channels != bpl

### LINKED ISSUE #950: Images in X11 backend only works with RGBA

sleeptightAnsiC:
Fixes: https://github.com/Immediate-Mode-UI/Nuklear/issues/950
Obsoletes: https://github.com/Immediate-Mode-UI/Nuklear/pull/952

See commit message and linked issues for more info.
~~This might be totally wrong so turning it into the DRAFT.~~

cc @TomJGooding , @RobLoach

### DIFF
--- a/demo/x11/nuklear_xlib.h
+++ b/demo/x11/nuklear_xlib.h
@@ -33,6 +33,7 @@ NK_API void                 nk_xlib_copy(nk_handle, const char*, int len);
 #ifdef NK_XLIB_INCLUDE_STB_IMAGE
 NK_API struct nk_image nk_xsurf_load_image_from_file(char const *filename);
 NK_API struct nk_image nk_xsurf_load_image_from_memory(const void *membuf, nk_uint membufSize);
+NK_API void            nk_xsurf_image_free(struct nk_image *image);
 #endif
 
 /* Font */
@@ -476,28 +477,44 @@ nk_xsurf_draw_text(XSurface *surf, short x, short y, const char *text, int len,
 
 #ifdef NK_XLIB_INCLUDE_STB_IMAGE
 NK_INTERN struct nk_image
-nk_stbi_image_to_xsurf(unsigned char *data, int width, int height, int channels) {
+nk_stbi_image_to_xsurf(const void *filename_or_membuf, nk_uint zero_or_membufSize) {
     XSurface *surf = xlib.surf;
-    struct nk_image img;
-    int bpl = channels;
-    long i, isize = width*height*channels;
-    XImageWithAlpha *aimage = (XImageWithAlpha*)calloc( 1, sizeof(XImageWithAlpha) );
-    int depth = DefaultDepth(surf->dpy, surf->screen);
-    if (data == NULL) return nk_image_id(0);
-    if (aimage == NULL) return nk_image_id(0);
+    struct nk_image img = nk_image_id(0);
+    unsigned char *data;
+    int width, height, channels, actual_channels;
+    long i, isize;
+    XImageWithAlpha *aimage = NULL;
+    int depth;
+
+    if (filename_or_membuf == NULL) goto bail;
+
+    aimage = (XImageWithAlpha*)calloc( 1, sizeof(XImageWithAlpha) );
+    if (aimage == NULL) goto bail;
+    img = nk_image_ptr( (void*)aimage);
+
+    depth = DefaultDepth(surf->dpy, surf->screen);
 
     switch (depth){
         case 24:
-            bpl = 4;
+            channels = 4;
         break;
         case 16:
         case 15:
-            bpl = 2;
+            channels = 2;
         break;
         default:
-            bpl = 1;
+            channels = 1;
         break;
     }
+
+    if (zero_or_membufSize) {
+        data = stbi_load_from_memory((const stbi_uc *)filename_or_membuf, zero_or_membufSize, &width, &height, &actual_channels, channels);
+    } else {
+        data = stbi_load(filename_or_membuf, &width, &height, &actual_channels, channels);
+    }
+    if (data == NULL) goto bail;
+
+    isize = width*height*channels;
 
     /* rgba to bgra */
     if (channels >= 3){
@@ -509,7 +526,7 @@ nk_stbi_image_to_xsurf(unsigned char *data, int width, int height, int channels)
         }
     }
 
-    if (channels == 4){
+    if (channels == 4 && actual_channels == 4){
         const unsigned alpha_treshold = 127;
         aimage->clipMask = XCreatePixmap(surf->dpy, surf->drawable, width, height, 1);
 
@@ -535,29 +552,30 @@ nk_stbi_image_to_xsurf(unsigned char *data, int width, int height, int channels)
            ZPixmap, 0,
            (char*)data,
            width, height,
-           bpl*8, bpl * width);
-    img = nk_image_ptr( (void*)aimage);
+           channels*8, channels*width);
     img.h = height;
     img.w = width;
+
     return img;
+bail:
+    nk_xsurf_image_free(&img);
+    return nk_image_id(0);
 }
 
 NK_API struct nk_image
 nk_xsurf_load_image_from_memory(const void *membuf, nk_uint membufSize)
 {
-    int x,y,n;
-    unsigned char *data;
-    data = stbi_load_from_memory(membuf, membufSize, &x, &y, &n, 0);
-    return nk_stbi_image_to_xsurf(data, x, y, n);
+    if (membufSize == 0) {
+        return nk_image_id(0);
+    } else {
+        return nk_stbi_image_to_xsurf(membuf, membufSize);
+    }
 }
 
 NK_API struct nk_image
 nk_xsurf_load_image_from_file(char const *filename)
 {
-    int x,y,n;
-    unsigned char *data;
-    data = stbi_load(filename, &x, &y, &n, 0);
-    return nk_stbi_image_to_xsurf(data, x, y, n);
+    return nk_stbi_image_to_xsurf((const void *)filename, 0);
 }
 #endif /* NK_XLIB_INCLUDE_STB_IMAGE */
 
@@ -579,15 +597,15 @@ nk_xsurf_draw_image(XSurface *surf, short x, short y, unsigned short w, unsigned
     }
 }
 
-void
+NK_API void
 nk_xsurf_image_free(struct nk_image* image)
 {
     XSurface *surf = xlib.surf;
     XImageWithAlpha *aimage = image->handle.ptr;
     if (!aimage) return;
-    XDestroyImage(aimage->ximage);
-    XFreePixmap(surf->dpy, aimage->clipMask);
-    XFreeGC(surf->dpy, aimage->clipMaskGC);
+    if (aimage->ximage)     XDestroyImage(aimage->ximage);
+    if (aimage->clipMask)   XFreePixmap(surf->dpy, aimage->clipMask);
+    if (aimage->clipMaskGC) XFreeGC(surf->dpy, aimage->clipMaskGC);
     free(aimage);
 }
 


sleeptightAnsiC:
Hi @riri ,

I'm pretty sure you're the most acknowledged with all X11/Xlib/XCB-related stuff. Mind taking a look here?

I feel like nk_stbi_image_to_xsurf is simply wrong. The things it assumes and tries to do look strange in so many ways... but I kept them all unchanged.

One thing I did for testing was running it under Xephyr like so...
```sh
$ Xephyr :2 -screen 1000x1000x16 &
$ DISPLAY=:2 twm &
$ DISPLAY=:2 ./bin/demo
```
...which affects the result of `DefaultDepth()` and let's me hit all other branches in this function, but I'm not sure how correct this is.

demo/x11/nuklear_xlib.h
```diff
@@ -476,29 +477,45 @@ nk_xsurf_draw_text(XSurface *surf, short x, short y, const char *text, int len,
...
         default:
-            bpl = 1;
+            channels = 1;
         break;
     }
 
+    if (zero_or_membufSize) {
+        data = stbi_load_from_memory((const stbi_uc *)filename_or_membuf, zero_or_membufSize, &width, &height, &channels, channels);
+    } else {
+        data = stbi_load(filename_or_membuf, &width, &height, &channels, channels);
```
TomJGooding: Doesn't this inadvertently change the `channels` value? From stb_image.h:

> If desired_channels is non-zero, *channels_in_file has the number of components that _would_ have been output otherwise.

RobLoach: Yeah, we likely shouldn't depend on stbi_load() since it uses STB's stdio instead of whatever the nuklear platform provides.

To my knowledge, passing `4` makes stbi load the image, and then output to 4 channels. Doesn't x11 expect 4 channels here?

sleeptightAnsiC: > Doesn't this inadvertently change the channels value?

Weird... This should pretty much cause the same problem as before, but it did not happen, even when I tested it with ASan. I think you're right though. Gimme sec...

sleeptightAnsiC: > we likely shouldn't depend on stbi_load() since it uses STB's stdio instead of whatever the nuklear platform provides.

We don't really have anything that would fit 1-to-1 here. There is [nk_file_load](https://github.com/Immediate-Mode-UI/Nuklear/blob/5a54a9f677ead97a581b5c8ab83cc30fdf237885/src/nuklear_util.c#L984-L1018) but not only it's private (aka `NK_LIB`) it's simply not worth to replace only one symbol. This whole backend already depends on libc, and stb_image is entirely optional (`NK_XLIB_INCLUDE_STB_IMAGE`) so it's not a big deal here.

TomJGooding: > To my knowledge, passing 4 makes stbi load the image, and then output to 4 channels. Doesn't x11 expect 4 channels here?

I'm not sure why you think X11 expects 4 channels - do you mean 32 bpp? 

The `bitmap_pad` argument in [XCreateImage](https://tronche.com/gui/x/xlib/utilities/XCreateImage.html) is overly convoluted in the docs:

> Specifies the quantum of a scanline (8, 16, or 32). In other words, the start of one scanline is separated in client memory from the start of the next scanline by an integer multiple of this many bits.

But this is basically the pixel size in bits (bpp).

The problem is that currently this is based on [DefaultDepth](https://tronche.com/gui/x/xlib/display/display-macros.html#DefaultDepth) rather than the actual image data from  `stbi_load`.

Simply changing the desired_channels to 4 for  `stbi_load` only works if we presuppose the [DefaultDepth](https://tronche.com/gui/x/xlib/display/display-macros.html#DefaultDepth) is 24, as mentioned in https://github.com/Immediate-Mode-UI/Nuklear/pull/952#issuecomment-4364614727.

sleeptightAnsiC: @RobLoach 

I'm not familiar with X11/Xlib at all, but I think Tom is right on this one.

Just by reading the code of `nk_stbi_image_to_xsurf`, it was specifically designed to handle images of which `desired_channels` does not fit with the values returned from `DefaultDepth`. I can't explain why someone would implement all of this otherwise.

Hardcoding `desired_channels` to 4 is fundamentally wrong. It only worked for you in https://github.com/Immediate-Mode-UI/Nuklear/pull/952 because `DefaultDepth` just happens to return something that works with 4 on your machine. It would most likely fail during Xephyr test I did in https://github.com/Immediate-Mode-UI/Nuklear/pull/954#issuecomment-4376423662

sleeptightAnsiC: @TomJGooding 

> > Doesn't this inadvertently change the channels value?
> 
> Weird... This should pretty much cause the same problem as before, but it did not happen, even when I tested it with ASan. I think you're right though. Gimme sec...

Fixed in [the latest force push](https://github.com/Immediate-Mode-UI/Nuklear/compare/ef4953b249fbcd9dc937a41ca71cca75555c2b7f..da89463d1122731c5566be373da5ecbdd9f6664f), and even though I still can't explain why it didn't corrupt/crash on me the first time, it was clearly a bug according to stb_image docs. So much thanks for pointing this one!

Also, because I'm getting paranoid at this point, here is proof that I literally debugged this patch line-by-line and it works as expected even when `DefaultDepth` returns something smaller: https://gist.github.com/sleeptightAnsiC/de9eae6ac6637758013ad55facb5860c

Also, also, updated https://github.com/Immediate-Mode-UI/Nuklear/issues/950#issuecomment-4376133509 and https://github.com/Immediate-Mode-UI/Nuklear/pull/954#issuecomment-4376423662 with better repros :)

sleeptightAnsiC: I think everything in this thread was answered/solved.

~~If you agree, leave :+1: and I'll resolve it.~~ EDIT: I think it's fine, just resolving...

demo/x11/nuklear_xlib.h
```diff
@@ -31,6 +31,7 @@ NK_API void                 nk_xlib_copy(nk_handle, const char*, int len);
 
 /* Image */
 #ifdef NK_XLIB_INCLUDE_STB_IMAGE
+/* FIXME: This always leaks memory. There is no API for free'ing images */
 NK_API struct nk_image nk_xsurf_load_image_from_file(char const *filename);
 NK_API struct nk_image nk_xsurf_load_image_from_memory(const void *membuf, nk_uint membufSize);
```
sleeptightAnsiC: @TomJGooding would you like me to implement some destructor for this one?

It's not a big deal, if you're only loading few images during init, but if you're loading them dynamically, it does leak quite a lot of memory:

<details><summary>ASan report</summary>
<p>

```
==280343==ERROR: LeakSanitizer: detected memory leaks
                                                                                                                                                                                                                                                                         
Direct leak of 24 byte(s) in 1 object(s) allocated from:
    #0 0x7f5584f2bf49 in calloc (/usr/lib/libasan.so.8+0x12bf49) (BuildId: ee5fbab73143ab257a66a33afe0f038a4af7a74e)                                                                                                                                                     
    #1 0x55c951540c39 in nk_stbi_image_to_xsurf /home/korn/Nuklear/demo/x11/nuklear_xlib.h:491
    #2 0x55c951541a08 in nk_xsurf_load_image_from_file /home/korn/Nuklear/demo/x11/nuklear_xlib.h:583
    #3 0x55c951553637 in main /home/korn/Nuklear/demo/x11/main.c:174
    #4 0x7f5584a27740  (/usr/lib/libc.so.6+0x27740) (BuildId: 020d6f7c33b2413f4fe10814c4729dce1387f049)
    #5 0x7f5584a27878 in __libc_start_main (/usr/lib/libc.so.6+0x27878) (BuildId: 020d6f7c33b2413f4fe10814c4729dce1387f049)
    #6 0x55c951462884 in _start (/home/korn/Nuklear/demo/x11/bin/demo+0x1c884) (BuildId: f1a637b95f1a35b682a8a558df7f1a47deba6dd6)

Indirect leak of 1048576 byte(s) in 1 object(s) allocated from:
    #0 0x7f5584f2c161 in malloc (/usr/lib/libasan.so.8+0x12c161) (BuildId: ee5fbab73143ab257a66a33afe0f038a4af7a74e)                                                                                                                                                     
    #1 0x55c951514c62 in stbi__malloc ../../example/stb_image.h:902
    #2 0x55c951529c1b in stbi__create_png_image_raw ../../example/stb_image.h:3990
    #3 0x55c95152cbba in stbi__create_png_image ../../example/stb_image.h:4164
    #4 0x55c95153017b in stbi__parse_png_file ../../example/stb_image.h:4438
    #5 0x55c95153058e in stbi__do_png ../../example/stb_image.h:4481
    #6 0x55c951530b59 in stbi__png_load ../../example/stb_image.h:4504
    #7 0x55c951514d0f in stbi__load_main ../../example/stb_image.h:946
    #8 0x55c951514fa0 in stbi__load_flip ../../example/stb_image.h:982
    #9 0x55c95151572f in stbi_load_from_file ../../example/stb_image.h:1058
    #10 0x55c951515610 in stbi_load ../../example/stb_image.h:1048
    #11 0x55c951540dd4 in nk_stbi_image_to_xsurf /home/korn/Nuklear/demo/x11/nuklear_xlib.h:513
    #12 0x55c951541a08 in nk_xsurf_load_image_from_file /home/korn/Nuklear/demo/x11/nuklear_xlib.h:583
    #13 0x55c951553637 in main /home/korn/Nuklear/demo/x11/main.c:174
    #14 0x7f5584a27740  (/usr/lib/libc.so.6+0x27740) (BuildId: 020d6f7c33b2413f4fe10814c4729dce1387f049)
    #15 0x7f5584a27878 in __libc_start_main (/usr/lib/libc.so.6+0x27878) (BuildId: 020d6f7c33b2413f4fe10814c4729dce1387f049)
    #16 0x55c951462884 in _start (/home/korn/Nuklear/demo/x11/bin/demo+0x1c884) (BuildId: f1a637b95f1a35b682a8a558df7f1a47deba6dd6)

Indirect leak of 160 byte(s) in 1 object(s) allocated from:
    #0 0x7f5584f2c161 in malloc (/usr/lib/libasan.so.8+0x12c161) (BuildId: ee5fbab73143ab257a66a33afe0f038a4af7a74e)                                                                                                                                                     
    #1 0x7f55855db419 in XCreateGC (/usr/lib/libX11.so.6+0x1c419) (BuildId: 246911752f62c4b240f6ecca688811b7ebf67d36)
    #2 0x55c951541172 in nk_stbi_image_to_xsurf /home/korn/Nuklear/demo/x11/nuklear_xlib.h:536
    #3 0x55c951541a08 in nk_xsurf_load_image_from_file /home/korn/Nuklear/demo/x11/nuklear_xlib.h:583
    #4 0x55c951553637 in main /home/korn/Nuklear/demo/x11/main.c:174
    #5 0x7f5584a27740  (/usr/lib/libc.so.6+0x27740) (BuildId: 020d6f7c33b2413f4fe10814c4729dce1387f049)
    #6 0x7f5584a27878 in __libc_start_main (/usr/lib/libc.so.6+0x27878) (BuildId: 020d6f7c33b2413f4fe10814c4729dce1387f049)
    #7 0x55c951462884 in _start (/home/korn/Nuklear/demo/x11/bin/demo+0x1c884) (BuildId: f1a637b95f1a35b682a8a558df7f1a47deba6dd6)

Indirect leak of 136 byte(s) in 1 object(s) allocated from:
    #0 0x7f5584f2bf49 in calloc (/usr/lib/libasan.so.8+0x12bf49) (BuildId: ee5fbab73143ab257a66a33afe0f038a4af7a74e)                                                                                                                                                     
    #1 0x7f55855e40aa in XCreateImage (/usr/lib/libX11.so.6+0x250aa) (BuildId: 246911752f62c4b240f6ecca688811b7ebf67d36)
    #2 0x55c951541767 in nk_stbi_image_to_xsurf /home/korn/Nuklear/demo/x11/nuklear_xlib.h:552
    #3 0x55c951541a08 in nk_xsurf_load_image_from_file /home/korn/Nuklear/demo/x11/nuklear_xlib.h:583
    #4 0x55c951553637 in main /home/korn/Nuklear/demo/x11/main.c:174
    #5 0x7f5584a27740  (/usr/lib/libc.so.6+0x27740) (BuildId: 020d6f7c33b2413f4fe10814c4729dce1387f049)
    #6 0x7f5584a27878 in __libc_start_main (/usr/lib/libc.so.6+0x27878) (BuildId: 020d6f7c33b2413f4fe10814c4729dce1387f049)
    #7 0x55c951462884 in _start (/home/korn/Nuklear/demo/x11/bin/demo+0x1c884) (BuildId: f1a637b95f1a35b682a8a558df7f1a47deba6dd6)

SUMMARY: AddressSanitizer: 1048896 byte(s) leaked in 4 allocation(s).
```

</p>
</details>

demo/x11/nuklear_xlib.h
TomJGooding: Honestly I'm starting to regret creating this issue, as it seems to have opened a can of worms. I probably should have double-checked this earlier, but will the rest of Nuklear even work if the X11 depth isn't 24...?

sleeptightAnsiC: > Honestly I'm starting to regret creating this issue, as it seems to have opened a can of worms.

It's fine. No worries.

> I probably should have double-checked this earlier, but will the rest of Nuklear even work if the X11 depth isn't 24...?

Yes, it (kinda) does. 1st screenshot below is 24, the 2nd one is 16. The colors are all funky, but it's hard for me to imagine any other outcome really. This backend calls `DefaultDepth()` everywhere where it's about to `XCreatePixmap()` so it seems the different depth was always meant to be supported.

<p float="left">
    
    
</p>

TomJGooding: Yikes that's ugly, but thanks for confirming!

~~*I haven't fully reviewed the code or gist yet, but I don't think the `channels_in_file` should just be discarded?~~

I've created a new comment for this since the thread was resolved.

demo/x11/nuklear_xlib.h
```diff
@@ -476,29 +477,45 @@ nk_xsurf_draw_text(XSurface *surf, short x, short y, const char *text, int len,
...
         default:
-            bpl = 1;
+            channels = 1;
         break;
     }
 
+    if (zero_or_membufSize) {
+        data = stbi_load_from_memory((const stbi_uc *)filename_or_membuf, zero_or_membufSize, &width, &height, NULL, channels);
+    } else {
+        data = stbi_load(filename_or_membuf, &width, &height, NULL, channels);
```
TomJGooding: I haven't fully reviewed the code or gist yet, but I don't think the `channels_in_file` should just be discarded?

> E.g. if you set desired_channels to 4, you will always get RGBA output, but you can check *channels_in_file to see if it's trivially opaque because e.g. there were only 3 channels in the source image.

We could avoid all the rigmarole for transparency in X11 when there''s only 3 channels.

sleeptightAnsiC: ~~In this case stb_image fills each transparency byte with `255` (aka "opaque"). Later during  that "transparency rigmarole", the `XDrawPoint()` happens only when `alpha > 127`, so in this case it always happens, because `255 > 127` is always true. Thus we cannot avoid it (?) Am I missing something? I really don't want to touch this code, even if it does something unnecessary. I could break it by accident.~~

nvm, it makes sense to avoid creating the clipmask in this case. I just need to test it. I thought the code below does more than than just clipmask (like translating how transparency data looks like, or something...)

sleeptightAnsiC:
from https://github.com/Immediate-Mode-UI/Nuklear/pull/954#discussion_r3202697336 :

> > > Doesn't this inadvertently change the channels value?
> >
> > Weird... This should pretty much cause the same problem as before, but it did not happen [...]
> 
> Fixed in [the latest force push](https://github.com/Immediate-Mode-UI/Nuklear/compare/ef4953b249fbcd9dc937a41ca71cca75555c2b7f..da89463d1122731c5566be373da5ecbdd9f6664f), [...] I still can't explain why it didn't corrupt/crash on me the first time, [...] I'm getting paranoid at this point

Just for the record: our version of `./example/stb_image.h` is so old, it's still affected by https://github.com/nothings/stb/issues/244. That's why my previous patch did not crash even though it was wrong. I'm not paranoid...

demo/x11/nuklear_xlib.h
sleeptightAnsiC: Changed in [latest force push](https://github.com/Immediate-Mode-UI/Nuklear/compare/da89463d1122731c5566be373da5ecbdd9f6664f..a398938692efd3255f93526e6373c3305530341b) and it works just fine. However, in order to test it, I had to manually `set actual_channels=3` under GDB right after loading Lena, because of the issue mentioned it https://github.com/Immediate-Mode-UI/Nuklear/pull/954#issuecomment-4405699318

image

demo/x11/nuklear_xlib.h
TomJGooding: Doesn't `nk_xsurf_image_free` free the image data with [XDestroyImage](https://tronche.com/gui/x/xlib/utilities/XDestroyImage.html)?

TomJGooding:
> Just for the record: our version of `./example/stb_image.h` is so old, it's still affected by [nothings/stb#244](https://github.com/nothings/stb/issues/244). That's why my previous patch did not crash even though it was wrong. I'm not paranoid...

Ah, that makes sense now - good detective work!

demo/x11/nuklear_xlib.h
sleeptightAnsiC: Hmm... Yes, this one should do the job.

Weird. That symbol is not exported with `NK_API`. Maybe that's why I haven't noticed it. 

I'm only a bit worried about the data malloc'ed by stb_load. According to man page:
```
when the image is created using XCreateImage, XGetImage, or XSubImage,
the destroy procedure that the XDestroyImage function calls
frees both the image structure and the data pointed to by the image structure.
```
so it seems like Xlib takes ownership of said data, which makes NO sense... but I guess Xlib is 40 years old (ha!)

EDIT: fixed in latest force push: https://github.com/Immediate-Mode-UI/Nuklear/compare/a398938692efd3255f93526e6373c3305530341b..cbf81289ae874c23e5c68818170f4188eddf9ca7

sleeptightAnsiC:
Alright, I think that's enough. Turning draft into mergable PR.

It doesn't corrupt mem any more, doesn't crash, doesn't misuse stb_image, doesn't misbehave with different DefaultDepth, doesn't leak upon partial failure, etc... That's more than enough for optional feature in a demo.

Our thought process is well documented here. If someone with proper Xlib knowledge will ever come and see this in the future, they'll know what we did right and/or wrong, and what to do next.

RobLoach:
I defer my review to Tom here

sleeptightAnsiC:
@TomJGooding Is there anything left I should do here? Not really sure what Rob meant.

TomJGooding:
LGTM - thanks @sleeptightAnsiC for tackling this!

Issue #726, opened 2024-10-29
SDL_gpu API example


bovacu:
Hi, now that SDL_gpu new API is out, I was wondering if there will be a new example (in the examples folder) implementing this new API.

Thanks!

riri:
I doubt it, but PR are welcome ;-)

bovacu:
Hi, thanks! I will definitely take a look at this when I get some time and try to make that PR

orosbogdan:
Any news on this?

sleeptightAnsiC:
IHMO, this would be way too much for Nuklear demo/example, hear me out.

SDL_gpu expects that you pass your own shaders in binary form, which is different per platform/backend. You can read more about this problem here: https://moonside.games/posts/layers-all-the-way-down/

So in order to implement SDL_gpu demo/example, we would have to provide multiple binary blobs (or embed them into source code with `xxd`) or create dependency on something like [SDL_shadercross](https://github.com/libsdl-org/SDL_shadercross), which is extremely complex IMO.

Don't get me wrong, I'm not saying "no", I just feel like this would cause more harm than actually help anyone. A lot of people does not seem to understand the real implications of this new API, it's nothing like SDL_renderer, not a silver bullet like some may think.

For anyone who wants to implement this: you can take a look at how imgui solved it, here: https://github.com/ocornut/imgui/pull/8163 Per [these instructions](https://github.com/ocornut/imgui/blob/master/backends/sdlgpu3/build_instructions.txt), they used HLSL for human-readable shader code, then compiled it into SPIRV blobs, then recompiled the blobs yet again for appropriate platform, then packed the blobs into headers with `xxd`-like utility. So the end-user does not have to do this on its own, but there is no way to easily change the shaders, which in my opinion breaks the idea of example that is expected to be easy to modify.

ompadu:
> IHMO, this would be way too much for Nuklear demo/example, hear me out.
> 
> SDL_gpu expects that you pass your own shaders in binary form, which is different per platform/backend. You can read more about this problem here: https://moonside.games/posts/layers-all-the-way-down/
> 
> So in order to implement SDL_gpu demo/example, we would have to provide multiple binary blobs (or embed them into source code with `xxd`) or create dependency on something like [SDL_shadercross](https://github.com/libsdl-org/SDL_shadercross), which is extremely complex IMO.
> 
[...]

Issue #889, opened 2026-02-06
duplicated sizeof checks in src/nuklear_internal.h


sleeptightAnsiC:
Follow up from https://github.com/Immediate-Mode-UI/Nuklear/pull/884#discussion_r2750159640 

> https://github.com/Immediate-Mode-UI/Nuklear/blob/a329721db93eb336f8eaae20f9c822730dd2ed6c/src/nuklear_internal.h#L61-L75
> 
> I wonder why do we need these duplicated checks here? The ones from `src/nuklear.h` do the very same thing, and all src/*.c files are including both of those headers, so I'm not sure what is the benefit?
> 
> This thing can only really harm us, just like in case of #185 where the `sizeof(nk_bool)` checks were different between each file, and went unnoticed for so long.

In theory having some duplication in `src/nuklear_internal.h` may make sense, but in practice those checks just do nothing. The proper sizes are already ensured at the moment when any of those is hit. This just looks like a dead code to me at this point.

I'm planning to remove this unless someone has a good reason to keep it (?)

PR #955, opened 2026-05-06
feat: config inclusion

### LINKED ISSUE #886: add a way to reuse Nuklear configuration across multiple source files
Problems:
- Nuklear is configured with macro defines. If you're including it inside of multiple source files, you need to ensure that config stays the same across all of them.
- If you change any macro define, you should recompile whole Nuklear and all files using it. If you're using an incremental build system, you may have a problem to detect those changes and all files affected by it.
- Nuklear may depend on some specific header coming from other library (e.g. `<SDL3/SDL.h>` in case you want to reuse SDL's types) and ensuring that this header is available everywhere is complicated
- Currently, if you want those problems solved, you need all kinds of hacks to make it behave correctly, either in build recipe or somewhere in Nuklear.
- I've created a _kicka$$_ config for `demo/sdl3_renderer`, which lets you build this demo on multiple platforms, but due to mentioned limitations, it must live in the main file [here](https://github.com/Immediate-Mode-UI/Nuklear/blob/25f33c8780a8b41668447f83e7d154e735cf424f/demo/sdl3_renderer/main.c#L21-L114), which makes it quite awkward to copy and reuse in your own code ([example](https://github.com/Immediate-Mode-UI/Nuklear/pull/884#discussion_r2756939664))
- demos with similar backends could reuse the same configuration code, which is currently impossible
- clangd (LSP) and other static analysis tools can get quite confused when you use single-header-library with predefines coming indirectly from source file

I thought about all of these for quite some time, and the solution that I use in my [personal branch](https://github.com/sleeptightAnsiC/Nuklear/blob/b1057aa92836a27f6c6a344295b7fb36543abab5/src/nuklear.h#L12-L29) looks like this. Somewhere at the very top of `src/Nuklear.h` :

```c
/*
 * Useful when you have a lot of stuff that you want to overwrite
 * Create "nk_config.h" file somewhere and make sure that preprocessor can find it
 * For example, use following compiler flags: -I/path/to/config/dir -DNK_CONFIG
 */
#ifdef NK_CONFIG
  #include "nk_config.h"
#endif
```

^ This works very well for my build system (custom GNU Make recipe), and I'm pretty sure it should work with everything else too. I've seen this trick used throughout many C libraries. Every private file in Nuklear is including this thing so it solves the problem. We should probably ensure that `src/nuklear_internal.h` also includes it, and maybe add our own include guards just to make sure, but those are just details that I'll figure out.

If you clicked at the [link to the branch](https://github.com/sleeptightAnsiC/Nuklear/blob/b1057aa92836a27f6c6a344295b7fb36543abab5/src/nuklear.h#L12-L29) from above, you would also notice that I already tried a different solution:

```c
#ifdef NK_CONFIG_FILE
  #include NK_CONFIG_FILE
#endif
```
^ but this one is impossible to support portably. MSVC does not support this trick to this very day, and trying to support it would require additional macro. I also noticed that passing paths into macro defines at command line (e.g. `-Dmacro=/my/path`) is very error prone (expansion works differently between compilers and shell types) and so I abandoned this idea entirely. (but we can also add it later, I guess)

Any thoughts ?

sleeptightAnsiC:
Resolves: https://github.com/Immediate-Mode-UI/Nuklear/issues/886

See linked issue and commit message for more details and reasoning.

I found this to be a rather good moment to submit this feature as I was updating some of my private repos with it. Also I'm planning to share some code that reuses `demo/sdl3_renderer` with different `main.c` for the purpose of https://github.com/Immediate-Mode-UI/Nuklear/issues/726#issuecomment-3504593874 and https://github.com/Immediate-Mode-UI/Nuklear/issues/888#issuecomment-3855140005, and having its config split into another file will be helpful there.

### DIFF
--- a/.github/ci_compile_sources.sh
+++ b/.github/ci_compile_sources.sh
@@ -13,6 +13,26 @@
 
 set -e
 
+tmpdir="$(mktemp -d)"
+trap 'rm -rf "${tmpdir}"' EXIT INT QUIT TERM
+
+cat <<'EOF' >"${tmpdir}/nuklear_config.h"
+    /*
+     * Intentionally without header guards to test Nuklear handling this case.
+     * The foo/bar below proves you can put arbitrary symbols into the config.
+     */
+    static int foo = 0;
+    static int bar(void) { return 0; }
+    #define NK_INCLUDE_FIXED_TYPES
+    #define NK_INCLUDE_DEFAULT_ALLOCATOR
+    #define NK_INCLUDE_STANDARD_IO
+    #define NK_INCLUDE_STANDARD_VARARGS
+    #define NK_INCLUDE_STANDARD_BOOL
+    #define NK_INCLUDE_VERTEX_BUFFER_OUTPUT
+    #define NK_INCLUDE_DEFAULT_FONT
+    #define NK_INCLUDE_COMMAND_USERDATA
+EOF
+
 CC=cc
 SRCDIR=./src
 
@@ -21,17 +41,12 @@ set -- "$@" -std=c89
 set -- "$@" -Wall
 set -- "$@" -Wextra
 set -- "$@" -pedantic
+set -- "$@" -Wno-unused  # only to silence warnings about static foo/bar
 CFLAGS=$*
 
 set -- ""
-set -- "$@" -DNK_INCLUDE_FIXED_TYPES
-set -- "$@" -DNK_INCLUDE_DEFAULT_ALLOCATOR
-set -- "$@" -DNK_INCLUDE_STANDARD_IO
-set -- "$@" -DNK_INCLUDE_STANDARD_VARARGS
-set -- "$@" -DNK_INCLUDE_STANDARD_BOOL
-set -- "$@" -DNK_INCLUDE_VERTEX_BUFFER_OUTPUT
-set -- "$@" -DNK_INCLUDE_DEFAULT_FONT
-set -- "$@" -DNK_INCLUDE_COMMAND_USERDATA
+set -- "$@" -I"${tmpdir}"
+set -- "$@" -DNK_INCLUDE_CONFIG
 CPPFLAGS=$*
 
 retcode=0
--- a/demo/sdl3_renderer/main.c
+++ b/demo/sdl3_renderer/main.c
@@ -18,98 +18,7 @@
 #define SDL_MAIN_USE_CALLBACKS
 #include <SDL3/SDL_main.h>
 
-/* ===============================================================
- *
- *                          CONFIG
- *
- * ===============================================================*/
-
-/* optional: sdl3_renderer does not need any of these defines
- * (but some examples might need them, so be careful) */
-#define NK_INCLUDE_STANDARD_VARARGS
-#define NK_INCLUDE_STANDARD_IO
-
-/* note that sdl3_renderer comes with nk_sdl_style_set_debug_font()
- * so you may wish to use that instead of font baking */
-#define NK_INCLUDE_FONT_BAKING
-#define NK_INCLUDE_DEFAULT_FONT
-
-/* note that sdl3_renderer comes with nk_sdl_allocator()
- * and you probably want to use that allocator instead of the default ones */
-/*#define NK_INCLUDE_DEFAULT_ALLOCATOR*/
-
-/* mandatory: sdl3_renderer depends on those defines */
-#define NK_INCLUDE_COMMAND_USERDATA
-#define NK_INCLUDE_VERTEX_BUFFER_OUTPUT
-
-
-/* We can re-use the types provided by SDL which are extremely portable,
- * so there is no need for Nuklear to detect those on its own */
-/*#define NK_INCLUDE_FIXED_TYPES*/
-#ifndef NK_INCLUDE_FIXED_TYPES
-    #define NK_INT8              Sint8
-    #define NK_UINT8             Uint8
-    #define NK_INT16             Sint16
-    #define NK_UINT16            Uint16
-    #define NK_INT32             Sint32
-    #define NK_UINT32            Uint32
-    /* SDL guarantees 'uintptr_t' typedef */
-    #define NK_SIZE_TYPE         uintptr_t
-    #define NK_POINTER_TYPE      uintptr_t
-#endif
-
-/* We can reuse the `bool` symbol because SDL3 guarantees its existence */
-/*#define NK_INCLUDE_STANDARD_BOOL*/
-#ifndef NK_INCLUDE_STANDARD_BOOL
-    #define NK_BOOL               bool
-#endif
-
-/* We can re-use various portable libc functions provided by SDL */
-#define NK_ASSERT(condition)      SDL_assert(condition)
-#define NK_STATIC_ASSERT(exp)     SDL_COMPILE_TIME_ASSERT(, exp)
-#define NK_MEMSET(dst, c, len)    SDL_memset(dst, c, len)
-#define NK_MEMCPY(dst, src, len)  SDL_memcpy(dst, src, len)
-#define NK_VSNPRINTF(s, n, f, a)  SDL_vsnprintf(s, n, f, a)
-#define NK_STRTOD(str, endptr)    SDL_strtod(str, endptr)
-
-/* SDL3 does not provide "dtoa" (only integer versions)
- * but we can emulate it with SDL_snprintf */
-static char* nk_sdl_dtoa(char *str, double d);
-#define NK_DTOA(str, d) nk_sdl_dtoa(str, d)
-
-/* SDL can also provide us with math functions, but beware that Nuklear's own
- * implementation can be slightly faster at the cost of some precision */
-#define NK_INV_SQRT(f)            (1.0f / SDL_sqrtf(f))
-#define NK_SIN(f)                 SDL_sinf(f)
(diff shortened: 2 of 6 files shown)


src/nuklear.h
```diff
@@ -9,6 +9,14 @@
 #ifdef __cplusplus
 extern "C" {
 #endif
+
+#ifdef NK_USER_CONFIG
```
sleeptightAnsiC: todo: rename `NK_USER_CONFIG` to `NK_INCLUDE_USER_CONFIG` in order to keep it consistent with other macros

EDIT: fixed in latest force push: https://github.com/Immediate-Mode-UI/Nuklear/compare/3acb495637ba1ed5f4887a6f4619301e1b15dd5d..e146fe51653287597aa5fc83c916b9fe1e312a1b

src/nuklear.h
```diff
@@ -9,6 +9,14 @@
 #ifdef __cplusplus
 extern "C" {
 #endif
+
+#ifdef NK_INCLUDE_USER_CONFIG
+  #ifndef NK_INCLUDE_USER_CONFIG_H_
+    #include "nk_user_config.h"
```
sleeptightAnsiC: One nit that I'm still thinking about: the prefix nk_user_* could be a bit confusing, especially for the file.

I went with it because of suggestion from https://github.com/Immediate-Mode-UI/Nuklear/issues/886#issuecomment-3895832790, and it made sense to me back then, but... I don't know, it just seems "odd"

How about?

`NK_INCLUDE_USER_CONFIG` -> `NK_INCLUDE_BUILD_CONFIG`

`"nk_user_config.h"` -> `"nk_build_config.h"`

or maybe even `"nuklear_build_config.h"` for the file?

RobLoach: Before adopting cmake, SDL had a SDL_config.h....

Could be a nukear_config.h.

Or make it a macro with a sane default

sleeptightAnsiC: went with nukear_config.h

.github/ci_compile_sources.sh
```diff
@@ -13,6 +13,23 @@
 
 set -e
 
+tmpdir="$(mktemp -d)"
+trap 'rm -rf "${tmpdir}"' EXIT INT QUIT TERM
+
+cat <<'EOF' >"${tmpdir}/nuklear_config.h"
+    /* intentionally without header guards */
+    static int foo;
+    static int bar(void) { return 0; }
```
RobLoach, line 25: What are foo and bar? And why are they used in the ci?

demo/sdl3_renderer/main.c
```diff
@@ -421,12 +330,3 @@ SDL_AppQuit(void* appstate, SDL_AppResult result)
     }
 }
 
-static char*
-nk_sdl_dtoa(char *str, double d)
```
RobLoach, line 425: Wouldn't it make sense to have this be part of the render instead of the config?

src/nuklear_internal.h
```diff
@@ -1,6 +1,13 @@
 #ifndef NK_INTERNAL_H
 #define NK_INTERNAL_H
 
+#ifdef NK_INCLUDE_CONFIG
+  #ifndef NK_NUKLEAR_CONFIG_H_
+    #include "nuklear_config.h"
+    #define NK_NUKLEAR_CONFIG_H_
+  #endif
```
RobLoach, line 11: May have missed some previous discussion but this ends up appearing twice? I don't think there are people who include src/nuklear-internals.h... is that the use case?

RobLoach (changes requested):

.github/ci_compile_sources.sh
sleeptightAnsiC: It's a proof that putting arbitrary functions/variables into the config would work fine. 

Also ensures that config is being included only once, even without proper include guards.

demo/sdl3_renderer/main.c
sleeptightAnsiC: No, not really.

Every our demo works in the way that nuklear.h has to be included BEFORE the backend.

If I would put it into the backend, it would not propagate correctly to nuklear.h, or would create some kind of circular dependency. It's just not possible without creating a mess. We already talked about something similar at https://github.com/Immediate-Mode-UI/Nuklear/pull/884#discussion_r2757796570

sleeptightAnsiC: Small note: this thing should probably be `inline`, but I had to keep it `static` for compatibility with C89. It's not a problem when using amalgamation, but it will trigger `-Wunused-function` when building into separate TUs. Still, it's just a demo, so it's not a problem.

EDIT: I think I will slap `SDL_FORCE_INLINE` there, just gotta make sure it's always available.

EDIT EDIT: fixed in https://github.com/Immediate-Mode-UI/Nuklear/compare/0ad44bcd8f8075fe44b4de4400797891d7595513..62c16aba09828827f20b180722709774db4d2bb8

src/nuklear_internal.h
sleeptightAnsiC: No, it's just to ensure that config propagates correctly there. nuklear_internal.h doesn't include nuklear.h directly. It just "happens" that source files are including those in correct order.

I'll be fine with removing this one, however, maybe we should simply fix the whole nuklear_internal.h inclusion problem in some other PR? I'm already tracking https://github.com/Immediate-Mode-UI/Nuklear/issues/889 that I could deal with at the same time.

sleeptightAnsiC:
Turning into draft. TODO: tweak nk_sdl_dtoa per https://github.com/Immediate-Mode-UI/Nuklear/issues/961#issuecomment-4499385821, and add another commit with note changed per https://github.com/Immediate-Mode-UI/Nuklear/issues/961#issuecomment-4499385821 (since I'm already touching this code here)

src/nuklear_internal.h
RobLoach: Duplicate seems funky, yeah. Without guards, it'd end up being run twice, which is likely not intentional.

.github/ci_compile_sources.sh
RobLoach: May want to add some doc comments so the goals are clear. Without it, it just seems like debug cruft.

src/nuklear_internal.h
sleeptightAnsiC: I'm going to keep the duplication for now, and will enter a new issue for it when this PR gets merged. The situation with `nuklear_internal.h` is a bit annoying.

sleeptightAnsiC:
Just beware I snuck another commit into this PR that just rewords one comment. I was already touching this exact code and it was harmless. It's related to request from https://github.com/Immediate-Mode-UI/Nuklear/issues/961

nuklear.h
```diff
@@ -6075,6 +6083,13 @@ template<typename T> struct nk_alignof{struct Big {T x; char c;}; enum {
 #ifndef NK_INTERNAL_H
 #define NK_INTERNAL_H
 
+#ifdef NK_INCLUDE_CONFIG
+  #ifndef NK_NUKLEAR_CONFIG_H_
+    #include "nuklear_config.h"
```
RobLoach: Another thought....

```c
#ifndef NK_CONFIG_H
#define NK_CONFIG_H "nuklear_config.h"
#include NK_CONFIG_H
#endif
```

This would also having the SDL renderer demo use the config pattern.

```c
#define NK_CONFIG_H "sdl3_renderer_config.h"
#include "nuklear.h"
```

sleeptightAnsiC: Not entirely sure what you're suggesting here. Those snippets wouldn't really work. After defining NK_CONFIG_H in the demo, the preprocessor would skip the inclusion. It's impossible to reuse same predefine for everything.

sleeptightAnsiC: @RobLoach I suspect, what you had in mind was probably something like this?

```c
#ifndef NK_NUKLEAR_CONFIG_H_
    #ifndef NK_CONFIG_FILE
        #define NK_CONFIG_FILE "nuklear_config.h"
    #endif
    #ifdef NK_INCLUDE_CONFIG
        #include NK_CONFIG_FILE
    #endif
#endif
```

it must be separated into 3 macros:
- `NK_NUKLEAR_CONFIG_H_` for additional include guard (private, just for safety)
- `NK_CONFIG_FILE` for telling which file to include (optional, defaults to `"nuklear_config.h"`)
- `NK_INCLUDE_CONFIG` for enabling inclusion

and the SDL renderer demo would have to use it like so:

```c
#define NK_CONFIG_FILE "sdl3_renderer_config.h"
#define NK_INCLUDE_CONFIG
#include "nuklear.h"
```

but, since demos are only one file, I don't see the point of using macros there. This is enough:

```c
#include "sdl3_renderer_config.h"
#include "nuklear.h"
```

RobLoach: You just proved that this config thing you put together isn't needed :stuck_out_tongue_winking_eye:

sleeptightAnsiC: wut? The _"config inclusion"_ and _"splitting demo's config"_ are two different things.

EDIT: we spoke about it in Discord DMs. I guess I will add `NK_CONFIG_FILE` functionality

sleeptightAnsiC: added in https://github.com/Immediate-Mode-UI/Nuklear/compare/ef0237075df967206a8035abc2fc80fc8ef409f2..8362119113725fa4ad1b7b16a063e10c907fcb6b

PR #960, opened 2026-05-15
fix(demo): remove nuklear math functions

### LINKED ISSUE #959: chart appears overly enlarged in Overview
Demo > Overview > Chart, one of the charts appears overly enlarged. It draws as far upwards as possible.

This does NOT reproduce under `demo/sdl3_renderer` for some reason, but every other backend that I tested seems affected.

This must be regression. I'm fairly sure it worked few weeks ago. I haven't git bisect'ed yet.

Image

TomJGooding:
This reverts commit 07941526ccb9d4197f00ea92967ca68cfb1fc36e and removes the remaining nuklear sin/cos functions in the examples.

These nuklear math functions were tailored for internal use only, since "nuklear only needs sine and cosine to generate circles" as documented in `nuklear_math.c`.

Replacing the standard math library functions with nuklear's approximation implementations breaks the examples in multiple places.

Fixes https://github.com/Immediate-Mode-UI/Nuklear/issues/959

### DIFF
--- a/demo/common/overview.c
+++ b/demo/common/overview.c
@@ -695,7 +695,7 @@ overview(struct nk_context *ctx)
             ctx->style.chart.show_markers = show_markers;
             if (nk_chart_begin(ctx, NK_CHART_LINES, 32, -1.0f, 1.0f)) {
                 for (i = 0; i < 32; ++i) {
-                    nk_flags res = nk_chart_push(ctx, (float)NK_COS(id));
+                    nk_flags res = nk_chart_push(ctx, (float)cos(id));
                     if (res & NK_CHART_HOVERING)
                         index = (int)i;
                     if (res & NK_CHART_CLICKED)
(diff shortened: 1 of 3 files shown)


RobLoach:
The consequence of doing this is that linking `m` is now required for the demos.

TomJGooding:
The consequence of _not_ doing this is that currently the demos are broken. 

I'm new to nuklear, but based on the docs in `nuklear_math.c` and the discussion in #959, these functions were never intended for general purpose math?

What do you suggest instead?

sleeptightAnsiC:
https://discord.com/channels/922234432071548968/922234432071548973/1504968654134186107

```
sleeptightAnsiC — 00:08
Forwarded
> The consequence of doing this is that linking m is now required for the demos.
@RobLoach I'm genuinely confused. Is this causing you any troubles in some downstream project?
People usually want to avoid standard math library (m aka libm) only when avoiding whole standard C library. All our demos always used both.
Default nk_cos/nk_sin only works for very limited float range, and returns wrong values in all other cases. It doesn't work as drop-in replacement for m

sleeptightAnsiC — 00:16
Not sure if you followed the whole conversation under the issue, but there is more than one place where it breaks the math, and it's hard to tell if it isn't compromising anything else.

sleeptightAnsiC — 00:27
Tom might went too fast with submitting PR, but what would be the alternative of fixing it now?

RobLoach — 01:36
Unsure, I just highlighted the consequence. Didn't suggest an alternative. I'm fine with merging it.

sleeptightAnsiC — 01:37
Okey. I will copy this for Tom in order to not keep him confused.
In general, I'm down to fixing these someday. But it's neither easy nor safe to do right now.
```

Issue #953, opened 2026-05-02
stb_image inconsistencies


TomJGooding:
stb_image is used to load images in various `example/*` and `demo/*` examples.

There's currently two different stb_image versions, where  `example/stb_image.h`  is v2.08 and `demo/common/filebrowser/stb_image.h` is v2.28.

stb_image also isn't compatible with C89 because of the C++ style comments.

Would it make sense to create a single stb_image file with the latest version (2.30) that is compatible with C89?

RobLoach:
Yeah, optimally we'd update example/stb_image.h to the latest with C89 compatibility.

sleeptightAnsiC:
from https://github.com/Immediate-Mode-UI/Nuklear/issues/895#issuecomment-3891208097

> > demos using stb_image [...] wont compile with pedantic on c89 (because using features from more recent standard)
> 
> IIRC those were mostly caused by C++ comments `//` and upstream didn't want to fix it ([nothings/stb#1328](https://github.com/nothings/stb/issues/1328)). I remember writing some POSIX AWK script for replacing `//` and preserving their logical structure (but I don't remember where I put it, so I'll probably have to start over...)

I can probably write up some script for fetching, replacing comments, and applying our patches to all stb libraries. This would help with issues such as https://github.com/Immediate-Mode-UI/Nuklear/pull/905#issuecomment-3978787904 where the changes are being repeatedly lost.

Maybe we can add new folder for all those vendored libraries? `./ven` ? `./deps` ? This would make it more clear and scalable, since we're already considering GLAD addition (https://github.com/Immediate-Mode-UI/Nuklear/pull/777), and I've seen few other PRs where it would help (e.g. https://github.com/Immediate-Mode-UI/Nuklear/pull/664, https://github.com/Immediate-Mode-UI/Nuklear/pull/902)

What do you guys think?

RobLoach:
Yeah. vendor them, though may want a readme since it's essentially a fork for the C89 support.

TomJGooding:
I made a start on this in https://github.com/Immediate-Mode-UI/Nuklear/pull/956, but might have jumped the gun after reading this more carefully..Sorry @sleeptightAnsiC, I missed your comment about potentially having a generic script to replace C++ comments for various other files used in Nuklear. 

[...]

PR #965, opened 2026-06-07
New Image Handling v2

RobLoach:
This is originally from @FrostKiwi over at https://github.com/Immediate-Mode-UI/Nuklear/pull/466 . Hopefully we can finish it off...

- [ ] Add support across all renderers
- [ ] Review X11
- [ ] Make sure it's clear that stb_image works as a helper library


sleeptightAnsiC:
I really think this is a bad idea.

nk_image is completely opaque for the library (for good reasons) and user already has full control of it on all sides - decides what to put into the handle and how to render it. Nuklear just passes it through into command queue and doesn't care what's inside.

There are more nuances to rendering a texture than just stretching/wrapping, and even though we can emulate it with multiple subimages, it still requires some support from backend in order to work correctly. 

That's said, I have out-of-tree SDL3 backend that is able to draw with and WITHOUT nk_convert (without _draw list_ API) and I could take a look into some of the problems from https://github.com/Immediate-Mode-UI/Nuklear/pull/466, but under one condition: all those changes from the demos/examples need to go - it's all extremely outdated and hacky, and we're already tracking stb_image changes in another Issue (https://github.com/Immediate-Mode-UI/Nuklear/issues/953). We can figure it out later.

Not saying "no", just don't want to clutter the library with something that doesn't even seem to work.

demo/x11/nuklear_xlib.h
```diff
@@ -30,11 +30,8 @@ NK_API void                 nk_xlib_paste(nk_handle, struct nk_text_edit*);
 NK_API void                 nk_xlib_copy(nk_handle, const char*, int len);
 
 /* Image */
-#ifdef NK_XLIB_INCLUDE_STB_IMAGE
-NK_API struct nk_image nk_xsurf_load_image_from_file(char const *filename);
-NK_API struct nk_image nk_xsurf_load_image_from_memory(const void *membuf, nk_uint membufSize);
+NK_API struct nk_image nk_stbi_image_to_xsurf(const void *pixels, int width, int height, int channels);
```
sleeptightAnsiC, line 33: Why removing the API that we fixed not even a month ago? https://github.com/Immediate-Mode-UI/Nuklear/pull/954

src/nuklear.h
```diff
@@ -271,7 +293,7 @@ struct nk_rect {float x,y,w,h;};
 struct nk_recti {short x,y,w,h;};
 typedef char nk_glyph[NK_UTF_SIZE];
 typedef union {void *ptr; int id;} nk_handle;
-struct nk_image {nk_handle handle; nk_ushort w, h; nk_ushort region[4];};
+struct nk_image {nk_handle handle; nk_ushort w, h; enum nk_image_type type; nk_ushort region[4];};
```
sleeptightAnsiC, line 296: This needs to be added as the very last field, not in the middle.

RobLoach: Agreed. Will have to fix in src too, but marking this so I remember to do that.
``` suggestion
struct nk_image {nk_handle handle; nk_ushort w, h; nk_ushort region[4]; enum nk_image_type type;};
```

demo/x11/nuklear_xlib.h
RobLoach: Oops, must've been a bad fix of the git conflict.

RobLoach:
Going to have to figure out a good way to do this. The different image handling methods are nice, and I agree we should not tie ourselves to stb_image.

Would like to investigate that fork further, but we will need C89. Would love to check out that SDL3 approach once I have the time.

demo/common/image.c
```diff
@@ -0,0 +1,294 @@
...
+#include "../../demo/common/libs/stb_image_c89.h"
+
+
+/* abusing nk_image type to avoid dependency issues */
+#ifndef load_image
+struct nk_image load_image(const char* filepath)
+{
+	int w, h;
+	struct nk_image img = {0};
+	img.handle.ptr = stbi_load(filepath, &w, &h, NULL, STBI_rgb_alpha);
```
sleeptightAnsiC, line 12: This hack is a pretty fragile idea. It only works with hardcoded channels and expects that backend will magically solve the ownership of the buffer (which, as commented on lines 158-163 below, doesn't work the same everywhere).

I understand nk_image was reused here because demos don't use any headers, but it's really bad idea to suggest people they can do such things, and we all know someone will probably try to reuse this code sooner or later.

I'll think about some alternatives.

PR #969, opened 2026-06-28
sdl3: Add keypad enter

RobLoach:
This adds SDLK_KP_ENTER, which invokes the enter event.


sleeptightAnsiC (approved):
Looked briefly through demo/sdl_render and didn't find anything else that sdl3 one misses. This was probably lost during times where sdl3_render was using scancodes (https://github.com/Immediate-Mode-UI/Nuklear/pull/940 changed it recently)

PR #971, opened 2026-06-29
CI: bump ccpp.yml to ubuntu-26.04 and add sdl3_renderer job

### LINKED ISSUE #897: Ubuntu release used by our CI is too old for fetching SDL3

sleeptightAnsiC:
current ubuntu-latest (24.04.3 LTS) doesn't have package for libsdl3-dev
26.04 is still in "preview" but it let's us test demo/sdl3_renderer

Refs: https://github.com/Immediate-Mode-UI/Nuklear/pull/894
Fixes: https://github.com/Immediate-Mode-UI/Nuklear/issues/897
Refs: https://github.com/Immediate-Mode-UI/Nuklear/issues/895

### DIFF
--- a/.github/workflows/ccpp.yml
+++ b/.github/workflows/ccpp.yml
@@ -5,7 +5,7 @@ on: [push, pull_request]
 jobs:
   build:
 
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-26.04
 
     steps:
     - uses: actions/checkout@v4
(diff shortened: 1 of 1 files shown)


sleeptightAnsiC:
Few notes:
- according to [actions/runner-images](https://github.com/actions/runner-images), ubuntu-26.04 is still in "preview" state. I haven't noticed any problems but beware
- Sometimes, I'm using [nektos/act](https://github.com/nektos/act) for running CI's locally but I was unable to get 26.04 working under it. That's not a big deal for me but, if someone else is using a similar tool, it might be a problem
- create-tag.yml and docs.yml are still using older ubuntu. I thought it's better to keep the stable version there
- it will probably make sense to swap back to ubuntu-lates as soon as 26.04 propagates to it

Issue #958, opened 2026-05-13
Compiler warnings


StrikerTheHedgefox:
I get these compiler warnings when trying to compile anything with the nuklear.h header.

```
1>tfury_nuklear.c
1>D:\sourcecode\TerminalFury\TerminalFury\core\backend\nuklear\nuklear.h(5668,47): warning C5287: operands are different enum types 'nk_panel_set' and 'nk_panel_type'; use an explicit cast to silence this warning
1>(compiling source file 'backend/tfury_nuklear.c')
1>D:\sourcecode\TerminalFury\TerminalFury\core\backend\nuklear\nuklear.h(5669,42): warning C5287: operands are different enum types 'nk_panel_set' and 'nk_panel_type'; use an explicit cast to silence this warning
1>(compiling source file 'backend/tfury_nuklear.c')
1>D:\sourcecode\TerminalFury\TerminalFury\core\backend\nuklear\nuklear.h(5761,46): warning C5287: operands are different enum types 'nk_window_flags' and 'nk_panel_flags'; use an explicit cast to silence this warning
1>(compiling source file 'backend/tfury_nuklear.c')
1>D:\sourcecode\TerminalFury\TerminalFury\core\backend\nuklear\nuklear.h(6935,34): warning C4127: conditional expression is constant
1>(compiling source file 'backend/tfury_nuklear.c')
1>D:\sourcecode\TerminalFury\TerminalFury\core\backend\nuklear\nuklear.h(10067,29): warning C4127: conditional expression is constant
1>(compiling source file 'backend/tfury_nuklear.c')
1>D:\sourcecode\TerminalFury\TerminalFury\core\backend\nuklear\nuklear.h(29037,49): warning C5287: operands are different enum types 'nk_edit_types' and 'nk_edit_flags'; use an explicit cast to silence this warning
1>(compiling source file 'backend/tfury_nuklear.c')
1>D:\sourcecode\TerminalFury\TerminalFury\core\backend\nuklear\nuklear.h(30275,1): warning C4701: potentially uninitialized local variable 'text' used
1>D:\sourcecode\TerminalFury\TerminalFury\core\backend\nuklear\nuklear.h(30172,1): warning C4701: potentially uninitialized local variable 'sym_background' used
1>D:\sourcecode\TerminalFury\TerminalFury\core\backend\nuklear_console\nuklear_console_label.h(78,1): warning C4700: uninitialized local variable 'handle' used
[...]

PR #937, opened 2026-04-08
Demo QOL changes

rswinkle:
Solves #935

Sorry for the delay.

Also fixes some C++ issues in Overview that showed up when compiling the SFML demos.

I personally ran the all SDL (except the old broken sdl3), sfml, glfw backends, plus the x11, x11_opengl2, x11_opengl3, and allegro5 backends. Pretty sure I can vouch for those.

I have not tested the d3d or gdi backends that I did change.

There are others that I didn't touch and we have several that should probably just be removed if we aren't going to be able to keep them up to date. Or at least mark them officially.

During my testing I also noticed that in most backends I could not trigger the double click indicator in the new Input section of Overview. Something to look into for later.

@RobLoach @sleeptightAnsiC

Edit: typos, clarification etc.

PR #972, opened 2026-06-30
add CI workflow for MSVC and fix GDI demo

sleeptightAnsiC:
Notes:
- setting up Windows/MSVC manually would be probably a huge headache. I just reused [ilammy/msvc-dev-cmd@v1](https://github.com/ilammy/msvc-dev-cmd) action which seems to be the most popular one.
- every demo/*/build.bat hardcodes directory to the same VC2015 toolchain. That's why CI uses `arch: x86` and `toolset: 14.0`. Would be nice to clean it up someday but for now I just want to get it in
- the workflow is in separate file. I thought it makes sense since it needs to run under entirely different OS/toolchain and builds entirely different demos
- GDI demo appears to be broken since https://github.com/Immediate-Mode-UI/Nuklear/commit/063968520f7bb8129b7a164906be055da3634cd3. I've patched it in separate commit.
- `build d3d11` step is noticeably slower. I don't have any clear explanation for this. Maybe because it is first to run, maybe because of different shader compilation, I'm not sure...
- there are quite a lot of warnings. I may fix those later as part of https://github.com/Immediate-Mode-UI/Nuklear/issues/958
- I'm a total noob when it comes to Github CI, so maybe there is something obvious I didn't notice

DO NOT SQUASH

Refs: https://github.com/Immediate-Mode-UI/Nuklear/issues/895
Refs: https://github.com/Immediate-Mode-UI/Nuklear/pull/937

### DIFF
--- a/.github/workflows/msvc.yml
+++ b/.github/workflows/msvc.yml
@@ -1,0 +1,36 @@
+name: MSVC CI
+
+on: [push, pull_request]
+
+jobs:
+  build:
+
+    runs-on: windows-2025
+
+    steps:
+
+    - uses: actions/checkout@v4
+    - uses: ilammy/msvc-dev-cmd@v1
+      with:
+        arch: x86
+        toolset: 14.0
+
+    - name: build d3d11
+      working-directory: demo/d3d11
+      run: ./build.bat
+    - name: build d3d12
+      working-directory: demo/d3d12
+      run: ./build.bat
+    - name: build d3d9
+      working-directory: demo/d3d9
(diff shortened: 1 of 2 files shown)


sleeptightAnsiC:
@RobLoach I've noticed that all my recent PRs were already merged but not this one. Would you mind taking a look? or maybe you already noticed something wrong here? Would be nice to get MSVC-related demos finally tested. AFAIK, none of us uses Windows daily, and it was only me booting VMs from time to time.

PR #973, opened 2026-07-01
fix: nk_font_config oversampling 3x horizontaly by default

### LINKED ISSUE #855: Font baking issue.
Seems like a font baking via stbtt is quite broken. This is how looks the default font with default settings on SDL3 on GPU side. Custom TTF fonts looks the same - unnecessary wide.
Image

sleeptightAnsiC:
See linked issue for additional information.

According to the commit that introduced this, it was supposed to: _"oversample the default font to make it look better_" but it was oversampling every single font, potentially making it blurry. This fix preserves old behavior for code that may still expect it, but nk_font_config will now use oversample_h=1 by default. Some demos achieve sharp font by scaling their pixel_size and in this case there won't be any difference, but if someone used badly behaving font and didn't notice it only worked because of oversampling, that font may now appear differently (this is unlikely but still possible).

Fixes: https://github.com/Immediate-Mode-UI/Nuklear/issues/855
Refs: https://github.com/Immediate-Mode-UI/Nuklear/commit/1d7f0244f97e2d35a57bc90586863ab89bea9be3

DO NOT SQUASH

### DIFF
--- a/src/nuklear_font.c
+++ b/src/nuklear_font.c
@@ -865,7 +865,7 @@ nk_font_config(float pixel_height)
     cfg.ttf_size = 0;
     cfg.ttf_data_owned_by_atlas = 0;
     cfg.size = pixel_height;
-    cfg.oversample_h = 3;
+    cfg.oversample_h = 1;
     cfg.oversample_v = 1;
     cfg.pixel_snap = 0;
     cfg.coord_type = NK_COORD_UV;
@@ -1173,9 +1173,14 @@ nk_font_atlas_bake(struct nk_font_atlas *atlas, int *width, int *height,
 
 #ifdef NK_INCLUDE_DEFAULT_FONT
     /* no font added so just use default font */
-    if (!atlas->font_num)
-        atlas->default_font = nk_font_atlas_add_default(atlas, 13.0f, 0);
+    if (!atlas->font_num) {
+        struct nk_font_config config;
+        config = nk_font_config(0);
+        config.oversample_h = 3;
+        atlas->default_font = nk_font_atlas_add_default(atlas, 13.0f, &config);
+    }
 #endif
+
     NK_ASSERT(atlas->font_num);
     if (!atlas->font_num) return 0;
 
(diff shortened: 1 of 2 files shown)


src/nuklear_font.c
```diff
@@ -1173,9 +1173,21 @@ nk_font_atlas_bake(struct nk_font_atlas *atlas, int *width, int *height,
...
+     * with code that creates empty atlas and immediately bakes it.
+     * Several demos do this, but it doesn't make sense for API to allow it.
+     * It was never documented anywhere and it's more of a hack than feature.
+     * App/backend should call nk_font_atlas_add_default() on it's own
+     * with whatever config it wants, and treat it like any other font.
+     * Worth to consider this for removal during next major release... */
+    if (!atlas->font_num) {
+        struct nk_font_config config;
+        config = nk_font_config(0);
+        config.oversample_h = 3;
```
sleeptightAnsiC, line 1186: I felt tempted to remove it even from here, but after testing a bit, it appears that some rawfb demos indeed look worse without it. Again, that probably just _**happens**_ to work and there is most likely a better way to achieve it, but I don't want to cause any unnecessary breakage.

src/nuklear_font.c
```diff
@@ -1173,9 +1173,21 @@ nk_font_atlas_bake(struct nk_font_atlas *atlas, int *width, int *height,
...
     /* no font added so just use default font */
-    if (!atlas->font_num)
-        atlas->default_font = nk_font_atlas_add_default(atlas, 13.0f, 0);
+    /* FIXME(sleeptightAnsiC): This "fallback" exists for compatibility
+     * with code that creates empty atlas and immediately bakes it.
+     * Several demos do this, but it doesn't make sense for API to allow it.
+     * It was never documented anywhere and it's more of a hack than feature.
+     * App/backend should call nk_font_atlas_add_default() on it's own
+     * with whatever config it wants, and treat it like any other font.
+     * Worth to consider this for removal during next major release... */
```
sleeptightAnsiC, line 1182: the comment is pretty much unrelated to PR, but since I already invested some time into understanding this code, I thought I would simply put it here. ~~I can split it into another commit, if it's a big deal.~~ EDIT: I just moved this to another commit.

RobLoach (approved):

PR #978, opened 2026-07-20
Make README example compilable

RobLoach:
Replaces the README example fragment with a complete program that compiles standalone (cc -std=c89 example.c -o example) and adds the compile command

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,41 +54,98 @@ This is very important; not doing it either leads to compiler errors, or even wo
 
 ## Example
 
+The following is a complete, minimal example. It builds one frame of user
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -54,41 +54,98 @@ This is very important; not doing it either leads to compiler errors, or even wo
 
 ## Example
 
+The following is a complete, minimal example. It builds one frame of user
+interface and iterates over the resulting draw commands, without needing a
+rendering backend or window system:
+
```
RobLoach:
``` suggestion
```

RobLoach:
On second thought, having a bare example is a bad idea. We should likely point to one of the demos instead.

Issue #815, opened 2025-06-16
Why c89? Why not c99 or c11?


awschult002:
Honestly curious about the decision to stick with c89. I understand that c99 had some flaws that prevented certain projects from adopting (Eg Linux kernel), but c11 has definitely fixed a bunch of stuff and provides nicer facilities for variable scoping to help remove bugs (eg Linux kernel allows c11).  Plus, more and more people are getting involved with C in a post c99/c11 world. So further development assistance would feel more natural if we could push for a newer version. 

What is the reason that nuclear mandates c89?what is the ethos behind sticking with this?

ghost:
It's all backwards compatibility. The project began targeted for this standard, and so it must continue as this standard lest it break all older programs that depend on it.

RobLoach:
There is nothing from compiling the project with C11. If you find any compatibility clashes, feel free to make an issue or a pull request. Thanks.

sleeptightAnsiC:
C89 is compatible with bunch of old and non-compliant compilers. It's also better for compatibility with C++ standards. Even some more recent compilers are quite bad with supporting C99/C11 (like MSVC)

There are few features I would rather not argue about - like VLAs and _Generic macros - and knowing all of those are disabled due to C89 strictness is beneficial for me personally. 

The only benefit from going past C89 would be use of _mixed declarations and code_. Nuklear uses `code... {decl... { code...}}` hack in order to emulate this syntax in few places, so cleaning this mess would be nice, but I don't see anything else as benefitial. Nuklear doesn't care about newer libc functions at all, neither about threads or atomics, or variadic macros. 

There is just no real reason for breaking C89 compatibility at this point. It could only harm people depending on it.

ColleagueRiley:
That is entirely wrong.

ColleagueRiley:
The maintainability and portability benefits greatly outweigh the supposed benefits of c89.

I have a library written in C99, RGFW, and I have to maintain support for multiple C++ compilers. MSVC, zig-cc, clang (C, C++, Objective-C, Objective-C++), mingw and gcc. Yes, certain features are not included in C++, but there are still plenty of features C99 and C++ share. I have encountered ZERO C99 compatibility issues with modern MSVC.

You are being complete ignorant about the down sides of C89 and c types. 

[...]

PR #981, opened 2026-08-08
Update to Nuklear 5.x

RobLoach:
Initial work based on the discussions on https://github.com/Immediate-Mode-UI/Nuklear/issues/815 .


RobLoach:
The diff isn't that large at this point. We can continue with further changes once `master` is on c99, and we have the `4.x` branch split.

PR #986, opened 2026-09-02
Fix tooltip width bug

rswinkle:
I started this just fixing the tooltip issue (unnecessary extra padding showing up on the right due to NK_TEXT_LEFT).

However, there were a couple of other sizing issues (a similar extra padding issue for tree_element selectable and combo height issues) that I fixed in the second commit. I included changes to sdl3_renderer/main.c for testing it.

I have several pull requests I need to make. I'd be fine if they went into 5.x instead of main but I'd rather we went the other way. Make a 4.x branch and let master be the 5.x. It just makes things easier imo since most people including myself tend to branch of master when contributing and 5.x will be the faster moving branch sooner or later.

### DIFF
--- a/demo/sdl3_renderer/main.c
+++ b/demo/sdl3_renderer/main.c
@@ -323,6 +323,113 @@ SDL_AppEvent(void *appstate, SDL_Event* event)
     return SDL_APP_CONTINUE;
 }
 
+/* Visual checks for the tree-element hit box and nk_combo height cap.
+ * Styles are exaggerated so a regression is obvious. */
+static void
+layout_sizing_bugs(struct nk_context *ctx)
+{
+    static const char *items[] = {
+        "Mercury", "Venus", "Earth", "Mars", "Jupiter", "Saturn", "Uranus", "Neptune"
+    };
+    static int combo_small = 0;
+    static int combo_large = 0;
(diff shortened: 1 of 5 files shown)


RobLoach:
Was playing around with this a bit. Seems good. Do you think we still need this Sizing bugs window?

Screenshot from 2026-09-01 23-45-23

RobLoach:
Regarding 5.x... It's not that big a change so far, when do you think is the best time to bring it in?

rswinkle:
> Was playing around with this a bit. Seems good. Do you think we still need this Sizing bugs window?

No just wanted you to have something to see before and after. Should've posted it as a gist but I was lazy. 

> Regarding 5.x... It's not that big a change so far, when do you think is the best time to bring it in?

The first time someone digs their heels in about a breaking change, don't bother fighting, just make 5.x master and keep moving, let 4.x be the stagnant branch. Since we already have a long list of stalled pull requests that require or benefit from breaking changes we could do it now, but I have at least one more pull request to make that is breaking (still crazy to worry about "breaking" for a statically compiled single header library to me) and so will likely end up in 5.x. I'll let others judge.

RobLoach (approved):
We could leave the sizing window in, or not. Either way it's a solid change. We can merge in 5.x once your PRs are good.

PR #988, opened 2026-09-02
Link widget

rswinkle:
@RobLoach 

What it says on the tin. Wire them up like buttons with something like SDL_OpenURL() or similar. Sure we could just use buttons or labels and jump through hoops with custom styling to try to get it right but I think we really needed proper first class links with common options.

https://github.com/user-attachments/assets/89a8e493-2376-454b-9843-a0fba7d6ab99


RobLoach (approved):
That's really cool! Nice work.

PR #989, opened 2026-09-02
DRAFT: Popups being clipped bug fix

rswinkle:
This was an idea I had months ago to deal with tooltips going off screen. Basically we add a member to the context to track the window size (requires changes to backends but everything is backward compatible so the backends that I didn't update get the old behavior). With that being updated by the backend, most things follow a relatively obvious sane policy (drop downs flip upward if they don't fit below, menus align with the right edge if they get clipped on the right, scrollbars get introduced only if there's not enough space total, user popups (nk_popup_begin) are fixed etc.) but tooltips are actually the hardest to decide. Right now they behave non-symmetrically.

First place it according to origin and offset. If it overflows on the right/bottom align the cursor with the right/bottom of the tooltip unless that would cause overflow on the left/top. If it's still overflowing the right/bottom, slide it left/up till it lines up with the edge. If x/y < 0 then slide right/down to x/y = 0. So if there's insufficient space for the tooltip, it will always clip the right/bottom rather than the left/top.

This asymmetry is easiest to see using the "Hover for magic" orbiting tooltip:

https://github.com/user-attachments/assets/72137afa-8210-4c7b-a408-9f3c90c89ceb

Honestly it's probably fine as it is, especially since the alternative is losing the content entirely to clipping, but there are arguments to be made for many other policies, symmetric, trying more to take into account the original origin specification, etc. I think this is probably the simplest policy that makes sense though.

Here's a longer video demonstrating the state of everything. I think I covered every type of popup and multiple edges for many, if not insufficient space for all of them:

https://github.com/user-attachments/assets/c8c7646f-3db4-43ec-b0cc-9baaa9a557eb

### DIFF
--- a/clib.json
+++ b/clib.json
@@ -1,6 +1,6 @@
 {
   "name": "nuklear",
-  "version": "4.13.3",
+  "version": "4.13.4",
   "repo": "Immediate-Mode-UI/Nuklear",
   "description": "A small ANSI C gui toolkit",
   "keywords": ["gl", "ui", "toolkit"],
--- a/demo/common/overview.c
+++ b/demo/common/overview.c
@@ -102,10 +102,11 @@ overview(struct nk_context *ctx)
                     const float values[]={26.0f,13.0f,30.0f,15.0f,25.0f,10.0f,20.0f,40.0f,12.0f,8.0f,22.0f,28.0f};
                     menu_state = MENU_CHART;
                     nk_layout_row_dynamic(ctx, 150, 1);
-                    nk_chart_begin(ctx, NK_CHART_COLUMN, NK_LEN(values), 0, 50);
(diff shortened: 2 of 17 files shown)


RobLoach:
This is nice :+1: Popups are always problematic, so being able to have clipping bounds is a nice addition.

rswinkle:
> This is nice 👍 Popups are always problematic, so being able to have clipping bounds is a nice addition.

Thanks. I think it's fine as is, I'm already using it in the latest release of [Shader Workshop](shaderworkshop.com). An unfortunately clipped combobox making it into the previous release was the final push I needed to finally do this.. I figure I'll wait for the next release and if I haven't run into any other problems by then, I'll mark it as ready for review.

